PR #42, opened 2024-02-24
feat: System config update event parsing

clabby:
## Overview

Adds `SystemConfig` event parsing for the L1 traversal stage.


crates/derive/src/types/system_config.rs
```diff
@@ -54,15 +54,135 @@ impl SystemConfig {
...
+        }
+        let update_type = u64::from_be_bytes(
+            log.topics()[2].as_slice()[24..]
+                .try_into()
+                .map_err(|_| anyhow!("Failed to convert update type to u64"))?,
+        );
+        let log_data = log.data.data.as_ref();
+
+        match update_type {
+            0 => {
```
tynes, line 94: It would be nice to have enum types defined to make these human readable

tynes (approved):

PR #50, opened 2024-02-26
feat(derive): Data Sources

### LINKED ISSUE #49: feat(`derive`): Calldata + Blobs sources
## Overview

In the L1 retrieval stage, a `DataAvailabilitySource` is required. This source contains a `DataIter`, which is an associated type within the trait that can iterate through raw data.

There are two types of data that the source can retrieve:
* Calldata
* Blobs

In the `op-node`, these can be found & [`calldata_source.go`](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/calldata_source.go) & [`blob_data_source.go`](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/blob_data_source.go).

We need to implement these data sources, and improve the `DataAvailabilitySource` trait's abstraction around the source of data. 

See also: [`DataSourceFactory`](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/data_source.go#L62-L80) for inspiration.

The actual endpoint that this data is received from should also be abstracted away, so that we can implement sources backed by the `PreimageOracle` or an RPC. There should be no concrete implementations for data fetching in this ticket.

refcell:
**Description**

Implements the calldata, blobs, and plasma data sources as specified in #49.

Fixes #49


crates/derive/src/sources/factory.rs
```diff
@@ -0,0 +1,63 @@
...
+            chain_provider: provider,
+            ecotone_timestamp: cfg.ecotone_time,
+            plasma_enabled: cfg.is_plasma_enabled(),
+        }
+    }
+}
+
+#[async_trait]
+impl<F: ChainProvider + Send + Sync + Clone + Debug> DataAvailabilityProvider
+    for DataSourceFactory<F>
```
clabby, line 39:
```suggestion
impl<F> DataAvailabilityProvider for DataSourceFactory<F>
where
   F: ChainProvider + Send + Sync + Clone + Debug
```

clabby (commented):
clean so far 🥳

crates/derive/src/stages/l1_retrieval.rs
```diff
@@ -60,7 +60,13 @@ where
...
         }
 
-        let data = self.data.as_mut().expect("Cannot be None").next();
+        let data = self
+            .data
+            .as_mut()
+            .expect("Cannot be None")
+            .next()
+            .await
+            .unwrap_or(Err(StageError::Empty));
```
clabby, line 69: This `unwrap_or` looks in an odd spot - if there's an error, we want to catch the EOF in the below match. What is the significance of `StageError::Empty`?

crates/derive/src/sources/factory.rs
```diff
@@ -0,0 +1,68 @@
...
+        }
+    }
+}
+
+#[async_trait]
+impl<F> DataAvailabilityProvider for DataSourceFactory<F>
+where
+    F: ChainProvider + Send + Sync + Clone + Debug,
+{
+    type DataIter = DataSource<F>;
```
clabby, line 45: Recommend removing this associated type, and just returning a concrete `DataSource<F>` in `open_data`

crates/derive/src/sources/plasma.rs
```diff
@@ -0,0 +1,27 @@
...
+        Self::default()
+    }
+}
+
+#[async_trait]
+impl AsyncIterator for PlasmaSource {
+    type Item = StageResult<Bytes>;
+
+    async fn next(&mut self) -> Option<Self::Item> {
+        unimplemented!()
```
clabby, line 25: Can we leave a note here that we won't implement this until further notice?

crates/derive/src/sources/calldata.rs
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    /// Loads the calldata into the source if it is not open.
+    async fn load_calldata(&mut self) -> anyhow::Result<()> {
+        if self.open {
+            return Ok(());
+        }
+
+        let info = self
```
clabby, line 57: Destructure the tuple?
```suggestion
        let (block_info, txs) = self
```

crates/derive/src/sources/calldata.rs
```diff
@@ -0,0 +1,93 @@
...
+#[async_trait]
+impl<CP: ChainProvider + Send> AsyncIterator for CalldataSource<CP> {
+    type Item = StageResult<Bytes>;
+
+    async fn next(&mut self) -> Option<Self::Item> {
+        if self.load_calldata().await.is_err() {
+            return Some(Err(StageError::BlockFetch(self.block_ref.hash)));
+        }
+
+        Some(self.calldata.pop().ok_or(StageError::Eof))
```
clabby, line 91: This is incorrect. The `pop()` function removes the last element from the iterator, we want to `pop_front`. Maybe `VecDequeue` is helpful here?

ref: https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/calldata_source.go#L67-L73

crates/derive/src/sources/calldata.rs
```diff
@@ -0,0 +1,93 @@
...
+            return Ok(());
+        }
+
+        let info = self
+            .chain_provider
+            .block_info_and_transactions_by_hash(self.block_ref.hash)
+            .await?;
+
+        self.calldata = info
+            .1
```
clabby, line 63: See: https://github.com/ethereum-optimism/kona/pull/50/files#r1503613856

crates/derive/src/sources/factory.rs
```diff
@@ -0,0 +1,68 @@
...
+                batcher_address,
+                *block_ref,
+                self.signer,
+            )));
+        }
+        if self.plasma_enabled {
+            return Ok(DataSource::Plasma(PlasmaSource::new()));
+        }
+        return Err(anyhow!("No data source available"));
+    }
```
clabby, line 67:
```suggestion
    async fn open_data(
        &self,
        block_ref: &BlockInfo,
        batcher_address: Address,
    ) -> Result<Self::DataIter> {
        if let Some(ecotone) = self.ecotone_timestamp {
            let source = (block_ref.timestamp >= ecotone)
                .then(|| DataSource::Blob(BlobSource::new()))
                .unwrap_or_else(|| {
                    DataSource::Calldata(CalldataSource::new(
                        self.chain_provider.clone(),
                        batcher_address,
                        *block_ref,
                        self.signer,
                    ))
                });
            Ok(source)
        } else if self.plasma_enabled {
            Ok(DataSource::Plasma(PlasmaSource::new()))
        } else {
            Err(anyhow!("No data source available"))
        }
    }
```

crates/derive/src/sources/calldata.rs
refcell: Since `VecDeque` is not available in the alloc crate (?), I've modified the `self.calldata` assignment to iterate in reverse. This allows us to just `.pop()` :)

crates/derive/src/sources/factory.rs
refcell: Very clean Paul Allen

crates/derive/src/stages/l1_retrieval.rs
refcell: Ah this is an unwrap on the `Option<>`. I've modified it to unwrap to `StageError::Eof`... don't think `StageError::Empty` makes sense .

crates/derive/src/sources/calldata.rs
clabby: I think it should be available @ [`alloc::collections::VecDeque`](https://doc.rust-lang.org/stable/alloc/collections/index.html#reexport.VecDeque) - we also use it in the `ChannelBank` + `FrameQueue` atm :D

clabby (approved):

PR #56, opened 2024-02-27
feat(derive): Span Batch Types

### LINKED ISSUE #55: feat(derive): `SpanBatch` type
## Overview

Implement a `SpanBatch` type, with the ability to encode and decode the type from raw data.

**References**
* [`op-node` SpanBatch type](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/span_batch.go)
* [SpanBatch specification](https://specs.optimism.io/protocol/span-batches.html)
* [@tynes' initial pass on the type](https://gist.github.com/clabby/bed87307caddadb2a2c5480ad534ea8b), inspired by `magi`'s current implementation.

refcell:
**Description**

Implements naive span batch types as specified in #55.

**Metadata**

Fixes #55.


crates/derive/src/types/batch/span_batch/bits.rs
```diff
@@ -0,0 +1,187 @@
...
+            let v = b[..buffer_len].to_vec();
+            b.advance(buffer_len);
+            v
+        };
+        let sb_bits = SpanBatchBits(bits.to_vec());
+
+        // TODO(clabby): Why doesn't this check work?
+        // if sb_bits.bit_len() > bit_length {
+        //     return Err(SpanBatchError::BitfieldTooLong);
+        // }
```
refcell, line 56: I think this is an incorrect check - the bitlist is left-padded with zeros using the `buffer_len` variable.

The `bit_length` variable is not padded to a multiple of 8 bits so the `sb_bits.bit_len()` can be larger that than the `bit_length` variable and be valid.

I believe the correct check should be:
```suggestion
        if sb_bits.bit_len() > (buffer_len * 8) {
        	return Err(SpanBatchError::BitfieldTooLong);
        }
```

crates/derive/src/types/batch/span_batch/bits.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// The encoded bitlist cannot be longer than [MAX_SPAN_BATCH_SIZE].
+    pub fn encode(
+        w: &mut Vec<u8>,
+        bit_length: usize,
+        bits: &SpanBatchBits,
+    ) -> Result<(), SpanBatchError> {
+        // TODO(clabby): Why doesn't this check work?
+        // if bits.bit_len() > bit_length {
+        //     return Err(SpanBatchError::BitfieldTooLong);
+        // }
```
refcell, line 72: ibid?

crates/derive/src/types/batch/span_batch/bits.rs
```diff
@@ -0,0 +1,187 @@
...
+    /// Decodes a standard span-batch bitlist from a reader.
+    /// The bitlist is encoded as big-endian integer, left-padded with zeroes to a multiple of 8 bits.
+    /// The encoded bitlist cannot be longer than [MAX_SPAN_BATCH_SIZE].
+    pub fn decode(b: &mut &[u8], bit_length: usize) -> Result<Self, SpanBatchError> {
+        let buffer_len = bit_length / 8 + if bit_length % 8 != 0 { 1 } else { 0 };
+        if buffer_len > MAX_SPAN_BATCH_SIZE {
+            return Err(SpanBatchError::TooBigSpanBatchSize);
+        }
+
+        // TODO(refcell): This can definitely be optimized.
```
refcell, line 40: nit: remove
```suggestion
```

PR #62, opened 2024-03-07
feat(derive): Test Utilities

refcell:
**Description**

Introduces test utilities to the `derive` crate `traits` module to allow for clean `Stage` unit testing.

### DIFF
--- a/crates/derive/src/traits/data_sources.rs
+++ b/crates/derive/src/traits/data_sources.rs
@@ -21,16 +21,16 @@ pub trait ChainProvider {
 /// Describes the functionality of a data source that can provide data availability information.
 #[async_trait]
 pub trait DataAvailabilityProvider {
-    /// A data iterator for the data source to return.
-    type DataIter<T: Into<Bytes>>: DataIter<T> + Send + Debug;
+    /// An iterator over returned bytes data.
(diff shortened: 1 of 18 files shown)


clabby:
👀

crates/derive/src/traits/data_sources.rs
```diff
@@ -21,16 +21,16 @@ pub trait ChainProvider {
...
     /// Returns the data availability for the block with the given hash, or an error if the block does not exist in the
     /// data source.
-    async fn open_data<T: Into<Bytes>>(
+    async fn open_data(
         &self,
         block_ref: &BlockInfo,
         batcher_address: Address,
-    ) -> Result<Self::DataIter<T>>;
+    ) -> Result<Self::DataIter>;
 }
```
refcell, line 34: This is a minor change I'd like your review on @clabby

clabby (approved):
Looks great. Let's get this in so we can keep iterating on tests 🚀

clabby (approved):

crates/derive/src/traits/data_sources.rs
clabby: This looks fine.

PR #65, opened 2024-03-26
feat(derive): Channel Reader Implementation

### LINKED ISSUE #51: feat(derivation): Channel Reader
## Overview

The `ChannelReader` stage is responsible for decompressing the channel read in the `ChannelBank` stage and parsing the batches from the decompressed byte stream.

**References**
* [Batch Format](https://specs.optimism.io/protocol/derivation.html#batch-format) specification.

refcell:
**Description**

Implements the basic enumerable `Batch` type for the `ChannelReader`.
Implements the `ChannelReader` stage.

**Metadata**

Fixes #51

### DIFF
--- a/crates/derive/src/stages/channel_reader.rs
+++ b/crates/derive/src/stages/channel_reader.rs
@@ -60,11 +60,7 @@ where
     /// Creates the batch reader from available channel data.
     async fn set_batch_reader(&mut self) -> StageResult<()> {
         if self.next_batch.is_none() {
-            let channel = match self.prev.next_data().await {
-                Ok(Some(channel)) => channel,
-                Ok(None) => return Err(anyhow!("no channel").into()),
-                Err(err) => return Err(err),
-            };
+            let channel = self.prev.next_data().await?.ok_or(anyhow!("no channel"))?;
             self.next_batch = Some(BatchReader::from(&channel[..]));
         }
         Ok(())
--- a/crates/derive/src/stages/channel_reader.rs
+++ b/crates/derive/src/stages/channel_reader.rs
@@ -96,7 +96,7 @@ impl BatchReader {
         if let Some(data) = self.data.take() {
             self.decompressed = decompress_to_vec(&data).ok()?;
         }
-        let batch = Batch::try_from(self.decompressed.as_ref()).ok()?;
+        let batch = Batch::decode(&mut self.decompressed.as_ref()).ok()?;
         Some(batch)
     }
 }
--- a/crates/derive/src/stages/channel_reader.rs
+++ b/crates/derive/src/stages/channel_reader.rs
@@ -6,7 +6,6 @@ use crate::{
     types::{Batch, BlockInfo, StageError, StageResult},
 };
 use alloc::vec::Vec;
-use alloy_rlp::Decodable;
 use anyhow::anyhow;
 use core::fmt::Debug;
 use miniz_oxide::inflate::decompress_to_vec;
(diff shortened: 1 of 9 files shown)


crates/derive/src/stages/channel_reader.rs
```diff
@@ -1 +1,112 @@
...
+                self.next_channel();
+                Err(StageError::Custom(e))
+            }
+        }
+    }
+
+    /// Creates the batch reader from available channel data.
+    async fn set_batch_reader(&mut self) -> StageResult<()> {
+        if self.next_batch.is_none() {
+            let channel = self.prev.next_data().await?.ok_or(anyhow!("no channel"))?;
```
clabby, line 64: The reference implementation makes a clear distinction between `io.EOF` and other errors - will need to follow that here to maintain correct error bubbling https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/channel_in_reader.go#L70-L78

crates/derive/src/types/batch.rs
```diff
@@ -0,0 +1,54 @@
...
+/// A Batch.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Batch {
+    /// A single batch
+    Single(SingleBatch),
+    /// Span Batches
+    Span(SpanBatch),
+}
+
+impl Decodable for Batch {
```
clabby, line 22: Think the `Decodable` and `Encodable` traits are probably wrong for this type, given that `SpanBatch`es only contain components which are RLP encoded. Maybe add custom functions for this logic in a standard `impl Batch` block?

ref: https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/channel_in_reader.go#L92-L118

crates/derive/src/types/batch_type.rs
```diff
@@ -0,0 +1,66 @@
...
+pub(crate) const SPAN_BATCH_TYPE: u8 = 0x02;
+
+/// The Batch Type.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum BatchType {
+    /// Single Batch.
+    Single = SINGLE_BATCH_TYPE as isize,
+    /// Span Batch.
+    Span = SPAN_BATCH_TYPE as isize,
+}
```
clabby, line 18:
```suggestion
/// The Batch Type.
#[derive(Debug, Clone, PartialEq, Eq)]
#[repr(u8)]
pub enum BatchType {
    /// Single Batch.
    Single = SINGLE_BATCH_TYPE,
    /// Span Batch.
    Span = SPAN_BATCH_TYPE,
}
```

clabby (changes requested):
Looking good! Would like to get this in soon, can finish up the span batch portions and some further testing in other PRs.

crates/derive/src/stages/channel_reader.rs
refcell: So I believe this will bubble up the correct error because the `?` operator will return the EOF error. The `ok_or` applies to the Option which wouldn't be an EOF error in the reference implementation.

clabby: Ah I see - you're right.

crates/derive/src/types/batch.rs
```diff
@@ -0,0 +1,42 @@
...
+/// A Batch.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum Batch {
+    /// A single batch
+    Single(SingleBatch),
+    /// Span Batches
+    Span(SpanBatch),
+}
+
+impl TryFrom<&[u8]> for Batch {
```
clabby, line 23: Can we make this a `decode` function in an `impl Batch` block?

When we have a reader (`&mut &[u8]`), we want to update the ptr as the data is read from it. With this we have to pass a reference to the actual data rather than the ptr, so our reader wouldn't get updated.

```rs
let buf = [0, 1, 2, 3, 4, ...];
let reader = &mut &buf;
let batch = Batch.try_from(reader).unwrap();
// reader was not advanced by the batch data read here
```

clabby (approved):

PR #87, opened 2024-04-04
fix(derive): OriginProvider Trait Usage

refcell:
**Description**

Implements the `OriginProvider` across stages, removing the method definitions.


clabby (approved):

clabby:
Follow up with bounding the previous stages to the new `OriginProvider` trait

PR #90, opened 2024-04-05
feat(derive): Add `ecrecover` trait + features

clabby:
## Overview

Adds a new trait, `SignedRecoverable`, which represents the functionality for recovering a public key from a signed type that has access to the message hash.

With the `k256` feature enabled, a default implementation of this trait is implemented on `TxEnvelope` from `alloy-consensus`. This feature can be disabled to allow for a custom secp256k1 recovery method. This will be used in the program to facilitate an accelerator for signature recovery during verification through the host program.

If the `k256` feature is not enabled, and there is no implementation of the `SignedRecoverable` extension trait on `TxEnvelope`, the `kona-derive` crate will fail to compile.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,6 +69,7 @@ dependencies = [
  "derive_more",
  "hex-literal",
  "itoa",
+ "k256",
  "ruint",
  "serde",
  "tiny-keccak",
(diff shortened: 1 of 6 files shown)


refcell (approved):
🫳🎤 You dropped this 👑

Issue #54, opened 2024-02-27
feat(derive): EngineQueue


clabby:
## Overview

The Engine Queue stage of the derivation pipeline is responsible for taking the payload attributes generated in the payload attributes derivation stage and send them to the execution engine.

For the sake of this implementation, the engine queue should be abstracted over the destination of the execution payloads (engine queue is probably not the best name). The best way to do this is probably to have a typed function that takes in the payloads and processes them.

**References**
- [Engine Queue](https://specs.optimism.io/protocol/derivation.html#engine-queue) specification.

refcell:
We shouldn't need this since attribute processing is application-specific. For example, the initial kona exex validation logic will just need to compare payload attributes. `kona-derive` is lower-level and shouldn't need to define how downstream applications consume generate payload attributes.

PR #92, opened 2024-04-05
feat(derive): Payload Attribute Building

refcell:
**Description**

Starts implementing a `StatefulAttributesBuilder` as a default implementation of the `AttributesBuilder` trait.

**Metadata**

Makes significant, tested progress on #54. There are a few more things to fix and implement to fully build payload attributes.


crates/derive/src/stages/attributes_queue/builder.rs
```diff
@@ -0,0 +1,170 @@
...
+    /// This can be done by either setting the `no_tx_pool` to false as sequencer, or by appending
+    /// batch transactions as the verifier.
+    fn prepare_payload_attributes(
+        &mut self,
+        l2_parent: L2BlockInfo,
+        epoch: BlockID,
+    ) -> Result<PayloadAttributes, BuilderError>;
+}
+
+/// The [L1ReceiptsFetcher] fetches L1 Header Info and [Receipt]s for
```
clabby, line 30: We should add a new function to `ChainProvider` for `info_by_hash` - we already have `receipts_by_hash` in there 😄

crates/derive/src/stages/attributes_queue/deposits.rs
```diff
@@ -0,0 +1,35 @@
...
+    deposit_contract: Address,
+) -> anyhow::Result<Vec<RawTransaction>> {
+    let receipts = receipts.into_iter().filter(|r| r.status).collect::<Vec<_>>();
+    // Flatten the list of receipts into a list of logs.
+    let addr = |l: &Log| l.address == deposit_contract;
+    let topics = |l: &Log| l.data.topics().first().map_or(false, |i| *i == DEPOSIT_EVENT_ABI_HASH);
+    let filter_logs =
+        |r: Receipt| r.logs.into_iter().filter(|l| addr(l) && topics(l)).collect::<Vec<Log>>();
+    let logs = receipts.into_iter().flat_map(filter_logs).collect::<Vec<Log>>();
+    // TODO(refcell): are logs **and** receipts guaranteed to be _in order_?
```
clabby, line 28: Almost positive they are, witth the RLP encoding of the receipts giving an in-order list 👍

crates/derive/src/stages/attributes_queue/builder.rs
```diff
@@ -0,0 +1,167 @@
...
+        if self.rollup_cfg.is_ecotone_active(next_l2_time) {
+            // parent_beacon_root = Some(l1_info.parent_beacon_root().unwrap_or_default());
+            // if the parent beacon root is not available, default to zero hash
+        }
+
+        Ok(PayloadAttributes {
+            timestamp: next_l2_time,
+            // TODO: The mix digest is pulled from the l1 info, which is a **full** block.
+            //       Currently, the l1_info is only the minimal `BlockInfo` defined in our types.
+            prev_randao: B256::default(),
```
clabby, line 156: Ah yeah we'll need this. I think we only need to have the header here?

crates/derive/src/stages/attributes_queue/builder.rs
```diff
@@ -0,0 +1,167 @@
...
+        // l1_info, next_l2_time)?;
+
+        let mut txs = vec![];
+        // txs.push(l1_info_tx);
+        txs.extend(deposit_transactions);
+        txs.extend(upgrade_transactions);
+
+        let withdrawals = None;
+        if self.rollup_cfg.is_canyon_active(next_l2_time) {
+            // withdrawals = Some(Withdrawals::default());
```
clabby, line 143: This looks right; We include the withdrawals field after Canyon (Shanghai activation), but it's always empty.

clabby (approved):
Looking very good; Nice work.

Approving to get this into trunk and continue iteration. Looks like `L1BlockInfo` parsing is blocking completion, where it's also blocking #98. Maybe we split the parsing of the L1 block info from calldata into another PR and put #98 on pause for the moment?

wrt needing more information in the payload builder, maybe we change `info_by_hash` in the `ChainProvider` to `header_by_hash`? The `BlockInfo` (for L1) is just a subset of the header's information, looks like we need the full header here. This contains the mix hash, parent beacon block root, etc., and we have a hint route for it as well: https://specs.optimism.io/experimental/fault-proof/index.html#l1-block-header-blockhash

crates/derive/src/stages/attributes_queue/builder.rs
clabby: See [review comment](https://github.com/ethereum-optimism/kona/pull/92#pullrequestreview-1999686310)

PR #93, opened 2024-04-06
feat(derive): Online `ChainProvider`

clabby:
## Overview

> [!NOTE]
> Blocked by need for `TxDeposit` variant in `TxEnvelope`.

Implements an online implementation of the `ChainProvider` trait, backed by an alloy RPC client. This provider is gated by the `alloy-providers` feature flag.


refcell (changes requested):
Let's keep the lints if possible by adding a new non-default online feature flag?

refcell (approved):

PR #94, opened 2024-04-06
feat(derive): Move to `tracing` for telemetry

clabby:
## Overview

> [!NOTE]
> RFC on if we want to keep our own telemetry utils.

Moves to the [`tracing`](https://crates.io/tracing) crate for telemetry, rather than the hand-rolled solution. The `tracing` crate provides simple macros and a robust collection system for traces, which can be leveraged in testing and in live environments. Because consumers of this crate, like `magi` and `reth`, use `tracing`, these telemetry logs will also be captured by their trace subscirbers.

In addition, trace collection can be avoided by not having a global subscriber, leaving no impact on performance in the case of the telemetry logs not being used. This will be advantageous for performance in the `client` program, where we want to avoid superfluous heap allocation.


refcell (approved):
Very nice, thank you for cleaning this up

PR #97, opened 2024-04-09
chore(derive): Derive full `SpanBatch` in channel reader

clabby:
## Overview

Swaps over to deriving the full `SpanBatch`, now that the implementation of `RawSpanBatch` is mostly stable.

### DIFF
--- a/crates/derive/src/stages/batch_queue.rs
+++ b/crates/derive/src/stages/batch_queue.rs
@@ -8,7 +8,7 @@ use crate::{
         SingleBatch, StageError, StageResult, SystemConfig,
     },
 };
-use alloc::{boxed::Box, vec::Vec};
+use alloc::{boxed::Box, sync::Arc, vec::Vec};
 use anyhow::anyhow;
 use async_trait::async_trait;
 use core::fmt::Debug;
@@ -45,7 +45,7 @@ where
     BF: L2ChainProvider + Debug,
 {
     /// The rollup config.
-    cfg: RollupConfig,
+    cfg: Arc<RollupConfig>,
     /// The previous stage of the derivation pipeline.
     prev: P,
     /// The l1 block ref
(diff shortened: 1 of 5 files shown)


crates/derive/src/types/batch/span_batch/raw.rs
```diff
@@ -32,27 +29,6 @@ impl RawSpanBatch {
...
-    }
-
-    /// Derives [SingleBatch]s from the span batch.
-    pub fn get_singular_batches(
-        &self,
-        _l1_blocks: &[BlockInfo],
-        _parent: L2BlockInfo,
-    ) -> Vec<SingleBatch> {
-        unimplemented!()
-    }
```
clabby, line 54: These are to be implemented on the `SpanBatch` type, not in `RawSpanBatch`

crates/derive/src/stages/batch_queue.rs
```diff
@@ -343,7 +343,11 @@ where
         match batch {
             Batch::Single(sb) => Ok(sb),
             Batch::Span(sb) => {
-                let batches = sb.get_singular_batches(&self.l1_blocks, parent);
+                let batches = sb.get_singular_batches(&self.l1_blocks, parent).map_err(|e| {
+                    StageError::Custom(anyhow!(
+                        "Could not get singular batches from span batch: {e}"
+                    ))
+                })?;
```
refcell, line 350: Can we add a test that the batch queue bubbles up this error if it fails to get singular batches? Not sure how easy it would be to construct the span batch such that the singular batch decoding fails here.

refcell (approved):

crates/derive/src/stages/batch_queue.rs
clabby: Going to follow up on this, but should be able to do this with some improved test utilities. Would like something similar to [`RandomRawSpanBatch`](https://github.com/ethereum-optimism/optimism/blob/8dd17a7b114a7c25505cd2e15ce4e3d0f7e3f7c1/op-node/rollup/derive/batch_test.go#L21) from the `op-node` test suite.

All of the `alloy` types we use have `Arbitrary` trait implementations with the `arbitrary` feature, which we can activate in the `test` profile. Would need to derive `Arbitrary` for our own types and add an `arbitrary` feature flag to `kona-derive`, but will be well worth :D

PR #100, opened 2024-04-14
feat(derive): Add `L1BlockInfoTx`

clabby:
## Overview

Adds the `L1BlockInfoTx` type, which supports encoding and decoding L1 info transactions on the OP Stack for both `Bedrock` and `Ecotone` versions.

A secondary constructor, `try_new_with_deposit_tx`, has been added that allows for generating the `TxDeposit` required in the attributes builder.


refcell (approved):
Beautiful

PR #105, opened 2024-04-15
fix(derive): Doc Touchups and Telemetry

refcell:
**Description**

Adds a bunch of missing telemetry, removing most inline `TODO` comments.

Also starts to add readme docs for `kona-derive`.

### DIFF
--- a/crates/derive/README.md
+++ b/crates/derive/README.md
@@ -2,5 +2,21 @@
 
 > **Notice**: This crate is a WIP.
 
-A `no_std` compatible implementation of the OP Stack's
-[derivation pipeline](https://specs.optimism.io/protocol/derivation.html#l2-chain-derivation-specification).
+A `no_std` compatible implementation of the OP Stack's [derivation pipeline][derive].
+
+[derive]: (https://specs.optimism.io/protocol/derivation.html#l2-chain-derivation-specification).
+
+## Features
+
+The most up-to-date feature list will be available on the [docs.rs `Feature Flags` tab][ff] of the `kona-derive` crate.
+
+Some features include the following.
+- `serde`: Serialization and Deserialization support for `kona-derive` types.
+- `k256`: [secp256k1][k] public key recovery support.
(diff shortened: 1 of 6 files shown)


crates/derive/README.md
```diff
@@ -2,5 +2,21 @@
...
+## Features
+
+The most up-to-date feature list will be available on the [docs.rs `Feature Flags` tab][ff] of the `kona-derive` crate.
+
+Some features include the following.
+- `serde`: Serialization and Deserialization support for `kona-derive` types.
+- `k256`: [secp256k1][k] public key recovery support.
+- `online`: Exposes an [alloy-provider][ap] powered data source using "online" HTTP requests.
+
+By default, `kona-derive` enables features `serde` and `k256`.
```
clabby, line 18: Maybe a bit early before it's all been sorted out, but can we highlight the difference between "online" and "offline" modes (i.e. disabling default features, custom trait impls), with a short piece on how the data source abstractions work?

crates/derive/README.md
```diff
@@ -2,5 +2,21 @@
...
+
+[derive]: (https://specs.optimism.io/protocol/derivation.html#l2-chain-derivation-specification).
+
+## Features
+
+The most up-to-date feature list will be available on the [docs.rs `Feature Flags` tab][ff] of the `kona-derive` crate.
+
+Some features include the following.
+- `serde`: Serialization and Deserialization support for `kona-derive` types.
+- `k256`: [secp256k1][k] public key recovery support.
```
clabby, line 15: Also would note _why_ we allow for disabling the native key recovery, to allow for an accelerator within the host program in "offline" mode

clabby (approved):
Looking very good. Couple nits on expanding the README, but can always do later :)

crates/derive/README.md
refcell: I pushed that off for now until we are really confident we have the right trait abstractions and solid implementations.

I think the data source abstractions deserves an entire section of it's own. Likely each top-level module in the derive crate deserves it's own section as well.

PR #127, opened 2024-04-19
feat(derive): Pipeline Builder

### LINKED ISSUE #110: feat(derive): Pipeline Builder
**Description**

Implement a `DerivationPipeline` following the [golang reference implementation](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/pipeline.go).

The pipeline will need to chain each Stage together from the [`stages`](https://github.com/ethereum-optimism/kona/tree/main/crates/derive/src/stages) module, successively passing each stage into the previous one. An example of this is shown below. This would be [`crates/derive/src/lib.rs`](https://github.com/ethereum-optimism/kona/blob/main/crates/derive/src/lib.rs). Note, this may break since various constructors and stage implementation change.

```rust
use alloc::sync::Arc;
use core::fmt::Debug;
use stages::{AttributesProvider, AttributesBuilder};
use traits::{ChainProvider, DataAvailabilityProvider, L2ChainProvider, OriginProvider};
use types::RollupConfig;

mod params;
pub use params::{
    ChannelID, CHANNEL_ID_LENGTH, CONFIG_UPDATE_EVENT_VERSION_0, CONFIG_UPDATE_TOPIC,
    DERIVATION_VERSION_0, FRAME_OVERHEAD, MAX_CHANNEL_BANK_SIZE, MAX_FRAME_LEN,
    MAX_RLP_BYTES_PER_CHANNEL, MAX_SPAN_BATCH_BYTES,
};
pub mod sources;
pub mod stages;
pub mod traits;
pub mod types;

/// The derivation pipeline is responsible for deriving L2 inputs from L1 data.
#[derive(Debug, Clone, Copy)]
pub struct DerivationPipeline<P, AB>
where
    P: AttributesProvider + OriginProvider + Debug,
    AB: AttributesBuilder + Debug,
{
	/// The attributes queue to retrieve the next attributes.
	pub attributes: AttributesQueue<P, AB>,
}

impl DerivationPipeline {
    /// Creates a new instance of the [DerivationPipeline].
    pub fn new<P, DAP, F, B>(
        rollup_config: Arc<RollupConfig>,
        chain_provider: P,
		dap_source: DAP,
		fetcher: F,
		builder: B,
    ) -> Self
    where
        P: ChainProvider + Clone + Debug + Send,
		DAP: DataAvailabilityProvider + Clone + Debug + Send,
		F: L2ChainProvider + Clone + Debug + Send,
		B: AttributesBuilder + Clone + Debug + Send,
    {
		let l1_traversal = L1Traversal::new(chain_provider, rollup_config.clone());
		let l1_retrieval = L1Retrieval::new(l1_traversal, dap_source);
		let frame_queue = FrameQueue::new(l1_retrieval);
		let channel_bank = ChannelBank::new(rollup_config.clone(), frame_queue);
		let channel_reader = ChannelReader::new(channel_bank, rollup_config.clone());
		let batch_queue = BatchQueue::new(rollup_config.clone(), channel_reader, fetcher);
		let attributes_queue = AttributesQueue::new(rollup_config.clone(), batch_queue, builder);
		Self { attributes: attributes_queue }
    }
}

// TODO: implement async iterator for derivation pipeline to call `next_attributes` on the internal `AttributesQueue` stage/field.
```

refcell:
**Description**

Implements the pipeline builder.

**Metadata**

Fixes #110


crates/derive/src/types/payload.rs
```diff
@@ -121,30 +121,30 @@ impl L2ExecutionPayloadEnvelope {
...
+                        "L2 block is missing L1 info deposit transaction, block hash: {}",
+                        execution_payload.block_hash
+                    );
+                }
 
-            let OpTxEnvelope::Deposit(tx) = tx else {
-                anyhow::bail!("First payload transaction has unexpected type: {:?}", tx.tx_type());
-            };
+                let tx = TxDeposit::decode(&mut execution_payload.transactions[0][1..].as_ref())
+                    .map_err(|e| anyhow::anyhow!(e))?;
```
refcell, line 139: I really don't like that I had to do this to strip the `0x7e` deposit prefix @clabby, I think there's an easier way to do this, but the issue is we check if the bytes are deposit transactions based on their first byte and then I think we should be decoding the deposit tx here. Maybe this should just be refactored into a `TxDeposit::decode_with_prefix()` method that gracefully strips the prefix? Curious as to your thoughts here.

crates/derive/src/types/payload.rs
```diff
@@ -121,30 +121,30 @@ impl L2ExecutionPayloadEnvelope {
...
 
-            let l1_info = L1BlockInfoTx::decode_calldata(tx.input.as_ref())?;
-            (l1_info.id(), l1_info.sequence_number())
-        };
+                // let OpTxEnvelope::Deposit(tx) = tx else {
+                //     anyhow::bail!("First payload transaction has unexpected type: {:?}",
+                // tx.tx_type()); };
```
refcell, line 143: You can see here that we were previously attempting to decode the tx bytes as an `OpTxEnvelope` which was failing and will fail if the first byte is `0x7e`.

crates/derive/src/types/payload.rs
clabby: You can also use `OpTxEnvelope::decode_2718` (I think `OpTxEnvelope::decode` does the same), which will decode a `OpTxEnvelope::TxDeposit(TxDeposit)`. Just need some extra matching.

We should do this - need to ensure that the first tx is always a deposit.

crates/derive/src/lib.rs
```diff
@@ -6,10 +6,14 @@
 
 extern crate alloc;
 
-use alloc::sync::Arc;
-use core::fmt::Debug;
-use traits::ChainProvider;
-use types::RollupConfig;
+/// Prelude exports common types and traits.
+pub mod prelude {
```
clabby, line 10: Can we not do a prelude? can just directly export from `pub mod`s @ base?

crates/derive/src/online/mod.rs
```diff
@@ -3,11 +3,42 @@
...
+use core::fmt::Debug;
+
+/// Creates a new [OnlineStageStack].
+#[cfg(feature = "online")]
+pub fn new_online_stack(
+    rollup_config: Arc<RollupConfig>,
+    chain_provider: impl ChainProvider + Clone + Debug + Send,
+    dap_source: impl DataAvailabilityProvider + OriginProvider + Clone + Debug + Send,
+    fetcher: impl L2ChainProvider + Clone + Debug + Send,
+    builder: impl AttributesBuilder + Clone + Debug + Send,
```
clabby, line 31: Can we provide the online implementations of these for the caller? i.e. an online stack uses the `AlloyChainProvider`, etc.

crates/derive/src/stages/attributes_queue.rs
```diff
@@ -152,10 +195,15 @@ where
...
+    P: AttributesProvider + PreviousStage + Send + Debug,
     AB: AttributesBuilder + Send + Debug,
 {
-    async fn reset(&mut self, _: BlockInfo, _: &SystemConfig) -> StageResult<()> {
+    async fn reset(
+        &mut self,
+        block_info: BlockInfo,
+        system_config: &SystemConfig,
+    ) -> StageResult<()> {
+        self.prev.reset(block_info, system_config).await?;
```
clabby, line 206: qq - is it ever significant the order in which stages are reset? We're resetting the prev stage recursively before updating the current stage in a few places.

crates/derive/src/stages/batch_queue.rs
```diff
@@ -476,6 +662,7 @@ mod tests {
         let mut bq = BatchQueue::new(cfg, mock, fetcher);
         let parent = L2BlockInfo::default();
         let result = bq.next_batch(parent).await;
-        assert!(result.is_err());
+        assert!(result.is_ok());
+        // assert_eq!(result, Err(StageError::NotEnoughData));
```
clabby, line 666: What's going on here?

clabby (approved):
Looking good. The ol' `Box<&dyn Trait>` recursion isn't too bad.

crates/derive/src/online/mod.rs
refcell: Great callout - an oversight on my part since I adapted generic code to be online 😄

refcell: Added concrete types for all except the dap source since we don't have an alloy-backed dap impl. Will stack a pr ontop to add that.

crates/derive/src/stages/attributes_queue.rs
refcell: So that's always the order in which the reference implementation performs the reset.

Essentially, the way the reference implementation performs a reset is in the following order:
a). the engine queue.
b). resets iteratively up the pipeline and resets each stage from l1 traversal to the attributes queue.

See the order of the `stages` list in the reference implementation here: https://github.com/ethereum-optimism/optimism/blob/6f99544c57ef837afaa96f979a6154a4e8a3604a/op-node/rollup/derive/pipeline.go#L93-L96

crates/derive/src/stages/batch_queue.rs
refcell: Oops, there were a few comments from a bad merge I believe. I've removed them and cleaned this up.

crates/derive/src/types/payload.rs
refcell: So unfortunately `TxDeposit::decode` does not support this, it reads the header first in it's decoding and that's why the unit tests strip the `0x7e` type prefix. e.g. https://github.com/clabby/op-alloy/blob/refcell/consensus-port/crates/op-consensus/src/transaction/optimism.rs#L177

refcell: For now, I've added a manual deposit tx type check. Will write up a ticket to fix this.

refcell: Added a ticket here https://github.com/ethereum-optimism/kona/issues/147

PR #128, opened 2024-04-21
feat: `kona-mpt` crate

clabby:
## Overview

> [!NOTE]
> Blocked by https://github.com/alloy-rs/alloy/pull/596, which will allow for continuously testing @ tip rather than a fixed block that contains no blob transactions. _edit:_ will just update later

Adds a new crate, `kona-mpt`, where we can put utilities for interacting with the Merkle Patricia Trie in the client program.

### Features

To start off, this crate includes the `OrderedListWalker`, which allows for traversing over a merkle patricia trie root that represents an ordered list (i.e., the transactions or receipt root in ETH block headers.) It does so by taking in a preimage fetching function, which is able to return the raw RLP of blinded nodes within the MPT.

### DIFF
--- a/crates/mpt/src/node.rs
+++ b/crates/mpt/src/node.rs
@@ -11,6 +11,18 @@ const BRANCH_LIST_LENGTH: usize = 17;
 /// The length of a leaf or extension node's RLP encoded list
 const LEAF_OR_EXTENSION_LIST_LENGTH: usize = 2;
 
+/// Prefix for even-nibbled extension node paths.
+const PREFIX_EXTENSION_EVEN: u8 = 0;
+
+/// Prefix for odd-nibbled extension node paths.
+const PREFIX_EXTENSION_ODD: u8 = 1;
+
+/// Prefix for even-nibbled leaf node paths.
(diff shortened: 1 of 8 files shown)


crates/mpt/src/node.rs
```diff
@@ -0,0 +1,191 @@
...
+        /// The value of the leaf node
+        value: Bytes,
+    },
+    /// An extension node is a 2-item pointer node with the encoding `rlp([encoded_path, key])`
+    Extension {
+        /// The path prefix of the extension
+        prefix: Bytes,
+        /// The pointer to the child node
+        node: Bytes,
+    },
```
refcell, line 48: Can these types be restricted further than a generic Bytes type? ie something with a fixed size

refcell (approved):
Very solid just one comment/question

crates/mpt/src/node.rs
clabby: Both represent opaque RLP encoded values in both variants, as the `TrieNode` doesn't recursively deserialize. They're encoded as strings as well, so having them as `Bytes` makes them easy to store

PR #129, opened 2024-04-22
fix(derive): Pipeline

refcell:
**Description**

This is an ugly refactor of the derivation pipeline builder using a simple, concrete online stack of the stages.


refcell:
Struggled with iteration over previous stages due to conflicting nested associated trait types. So I've modified each stage's reset to first call the previous stage's reset and bubble up any errors there.

This follows the reference implementation but optimistically resets all stages instead of iterating over each and resetting in order. See: https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/pipeline.go#L93-L96

If we need to implement iterative reset, we can have the `ResettableStage` trait's `reset` method return an enum that has an already reset variant that'd allow the recursive reset to dispatch on the results at each stage level.

refcell:
Additionally, the current pipeline gives ownership of each stage to the successive stage. This means the top-level pipeline object or "stage stack" wrapper does not have a handle to the concrete `L1Traversal` stage. But the pipeline needs to call the `advance_l1_block` method when a `StageError::Eof` is encountered. The current naive solution just adds an atomic reference `advance` flag to the L1 traversal stage that the stack can set when an eof error is received when fetching the next attributes.

refcell:
Merged pr handled the above questions 😄

Issue #132, opened 2024-04-24
feat(host): Host Tracking Issue


refcell:
**Description**

This is a tracking issue for the `host` component of the Fault Proof Program.

**Ordered Tickets**

- [ ] https://github.com/ethereum-optimism/kona/issues/131

refcell:
Assigning myself to shepherd this

PR #136, opened 2024-04-24
WIP host tracking

N0xMare:
meant to address issue #132 Host Tracking

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,7 @@ exclude = ["**/target", "benches/", "tests"]
 anyhow = { version = "1.0.79", default-features = false }
 tracing = { version = "0.1.40", default-features = false }
 cfg-if = "1.0.0"
-alloy-primitives = { version = "0.7.0", default-features = false, features = ["serde"] }
+alloy-primitives = { version = "0.7.0", default-features = false }
 
 [profile.dev]
 opt-level = 1
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,5 +1,5 @@
 [workspace]
-members = ["crates/*"]
+members = ["crates/*", "bin/host"]
 exclude = ["examples/minimal", "examples/simple-revm", "fpvm-tests/cannon-rs-tests"]
 resolver = "2"
 
@@ -15,7 +15,7 @@ exclude = ["**/target", "benches/", "tests"]
 anyhow = { version = "1.0.79", default-features = false }
 tracing = { version = "0.1.40", default-features = false }
 cfg-if = "1.0.0"
-alloy-primitives = { version = "0.7.0", default-features = false }
+alloy-primitives = { version = "0.7.0", default-features = false, features = ["serde"] }
 
 [profile.dev]
 opt-level = 1
--- a/bin/host/Cargo.toml
+++ b/bin/host/Cargo.toml
@@ -19,4 +19,6 @@ alloy-primitives.workspace = true
 tokio = { version = "1.37.0", features = ["full"] }
 clap = { version = "4.5.4", features = ["derive", "env"] }
 serde = { version = "1.0.198", features = ["derive"] }
+tracing-subscriber = "0.3.18"
+tracing = { version = "0.1.40", default-features = false }
 
--- a/bin/host/Cargo.toml
+++ b/bin/host/Cargo.toml
@@ -12,7 +12,7 @@ exclude.workspace = true
 # workspace
 anyhow.workspace = true
 tracing.workspace = true
-alloy-primitives.workspace = true
+alloy-primitives = { workspace = true, features = ["serde"] }
 
 tokio = { version = "1.37.0", features = ["full"] }
 clap = { version = "4.5.4", features = ["derive", "env"] }
--- a/bin/host/Cargo.toml
+++ b/bin/host/Cargo.toml
@@ -1,5 +1,5 @@
 [package]
-name = "host"
+name = "kona-host"
 version = "0.1.0"
 edition.workspace = true
 license.workspace = true
@@ -7,8 +7,6 @@ authors.workspace = true
 repository.workspace = true
 homepage.workspace = true
 exclude.workspace = true
-
-# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
 # workspace
@@ -20,5 +18,4 @@ tokio = { version = "1.37.0", features = ["full"] }
 clap = { version = "4.5.4", features = ["derive", "env"] }
 serde = { version = "1.0.198", features = ["derive"] }
 tracing-subscriber = "0.3.18"
-tracing = { version = "0.1.40", default-features = false }
 
--- a/bin/host/Cargo.toml
+++ b/bin/host/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "host"
+version = "0.1.0"
+edition.workspace = true
+license.workspace = true
+authors.workspace = true
+repository.workspace = true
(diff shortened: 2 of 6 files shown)


bin/host/Cargo.toml
```diff
@@ -0,0 +1,24 @@
+[package]
+name = "host"
```
clabby, line 2:
```suggestion
name = "kona-host"
```

bin/host/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+version = "0.1.0"
+edition.workspace = true
+license.workspace = true
+authors.workspace = true
+repository.workspace = true
+homepage.workspace = true
+exclude.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
clabby, line 11:
```suggestion
```

bin/host/src/main.rs
```diff
@@ -0,0 +1,82 @@
+use alloy_primitives::B256;
+use clap::{Parser, ValueEnum};
+use serde::Serialize;
+use std::path::PathBuf;
+use std::str::FromStr;
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    let _cli = Cli::parse();
+    tracing_subscriber::Registry::default().init();
```
clabby, line 10: It would be nice to have a tracing subscriber with configurable verbosity for `tracing::Level`s.

Suggested:

1. Add a `v` field to the `Cli` struct, using `ArgAction` to increment (i.e. `-v` up to `-vvvvv`)

```rs
struct Cli {
    /// Verbosity level (0-4)
    #[arg(long, short, help = "Verbosity level (0-4)", action = ArgAction::Count)]
    v: u8,
    // ...
}
```
2. Add a function that can initialize the tracing subscriber with a configured verbosity level:
```rs
/// Initializes the tracing subscriber
///
/// # Arguments
/// * `verbosity_level` - The verbosity level (0-4)
///
/// # Returns
/// * `Result<()>` - Ok if successful, Err otherwise.
fn init_tracing_subscriber(verbosity_level: u8) -> Result<()> {
    let subscriber = tracing_subscriber::fmt()
        .with_max_level(match verbosity_level {
            0 => Level::ERROR,
            1 => Level::WARN,
            2 => Level::INFO,
            3 => Level::DEBUG,
            _ => Level::TRACE,
        })
        .finish();
    tracing::subscriber::set_global_default(subscriber).map_err(|e| anyhow!(e))
}
```
3. Initialize the subscriber in main using the above function, i.e. `init_tracing_subscriber(cli.v)?`

bin/host/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+    Ok(())
+}
+
+fn parse_b256(s: &str) -> Result<B256, String> {
+    B256::from_str(s).map_err(|_| format!("Invalid B256 value: {}", s))
+}
+
+/// Available networks.
+#[derive(Debug, Clone, ValueEnum, Serialize)]
+pub enum Network {
```
clabby, line 21: Please move this into a new module, same with `RpcKind`

bin/host/src/main.rs
```diff
@@ -0,0 +1,82 @@
...
+
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    let _cli = Cli::parse();
+    tracing_subscriber::Registry::default().init();
+    tracing::info!("host telemetry initialized");
+    Ok(())
+}
+
+fn parse_b256(s: &str) -> Result<B256, String> {
```
clabby, line 15: This should not be necessary idt? Does `B256` require a value parser?

clabby (commented):
Looking good so far.

To get this in, we'll need to make some edits to CI that ensures that the host is not built or linted with the `cannon`/`asterisc` build pipelines, as it will depend on the standard library.

bin/host/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+# workspace
+anyhow.workspace = true
+tracing.workspace = true
+alloy-primitives.workspace = true
+
+tokio = { version = "1.37.0", features = ["full"] }
+clap = { version = "4.5.4", features = ["derive", "env"] }
+serde = { version = "1.0.198", features = ["derive"] }
+tracing-subscriber = "0.3.18"
+tracing = { version = "0.1.40", default-features = false }
```
clabby, line 23: This is a duplicate

bin/host/Cargo.toml
clabby: This crate will also need to be put inside of the workspace.

bin/host/Cargo.toml
```diff
@@ -0,0 +1,24 @@
...
+homepage.workspace = true
+exclude.workspace = true
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+# workspace
+anyhow.workspace = true
+tracing.workspace = true
+alloy-primitives.workspace = true
```
clabby, line 17: Activate the `serde` feature so that `B256` and other primitives from alloy implement `serde::Serialize`

clabby: Yeah this removes the need for your custom value parser

bin/host/src/main.rs
clabby: ah ya, necessary. Please place custom value parsers in a new module

bin/host/Cargo.toml
clabby: Yeah for sure - given that this is a bin crate, think we can just always enable serde. Folks shouldn't be consuming this code downstream.

bin/host/src/parser.rs
```diff
@@ -0,0 +1,6 @@
+use alloy_primitives::B256;
+use std::str::FromStr;
+
+pub fn parse_b256(s: &str) -> Result<B256, String> {
```
clabby, line 4: Docs pls

bin/host/src/types.rs
```diff
@@ -0,0 +1,14 @@
...
+use serde::Serialize;
+
+/// Available networks.
+#[derive(Debug, Clone, ValueEnum, Serialize)]
+pub enum Network {
+    /// Optimism Mainnet
+    Optimism,
+}
+
+#[derive(Debug, Clone, ValueEnum, Serialize)]
```
clabby, line 11: docs pls

bin/host/src/main.rs
```diff
@@ -0,0 +1,100 @@
+mod parser;
+mod types;
+use crate::{
+    parser::parse_b256,
+    types::{Network, RpcKind}
+};
+
```
clabby, line 7:
```suggestion
```

bin/host/src/main.rs
```diff
@@ -0,0 +1,100 @@
+mod parser;
+mod types;
+use crate::{
```
clabby, line 3:
```suggestion

use crate::{
```

Cargo.toml
```diff
@@ -15,7 +15,7 @@ exclude = ["**/target", "benches/", "tests"]
 anyhow = { version = "1.0.79", default-features = false }
 tracing = { version = "0.1.40", default-features = false }
 cfg-if = "1.0.0"
-alloy-primitives = { version = "0.7.0", default-features = false }
+alloy-primitives = { version = "0.7.0", default-features = false, features = ["serde"] }
```
clabby, line 18: This shouldn't be enabled by default in the workspace dep - should be enabled in `kona-host`'s `Cargo.toml`

refcell (approved):

PR #144, opened 2024-04-26
feat(ci): Add release plz check

clabby:
Adds a check on PRs for the release plz job. Does a dry run of the release so that we can see if it will fail prior to landing in `main`


clabby:
This isn't gonna work

PR #146, opened 2024-04-27
[WIP] Large Primitives and Provider Refactor

refcell:
**Description**

This is a large refactor of types in `kona-derive` into the `kona-primitives` crate and non-derivation specific (or "external facing") traits into a `kona-providers` crate.

Splitting types out from the derivation pipeline forces type implementations not to pull in traits and derivation objects into primitives. For example, we shouldn't be able to pass a `fetcher` into the batch validation. It should be a non-colored method checking batch validation.


refcell:
Keeping this open for now since there are some embedded changes in here that need to be split out

PR #156, opened 2024-04-28
feat(derive): Abstract Alt DA out of `kona-derive`

clabby:
## Overview

Abstracts alternative data availability adapters out of `kona-derive` by doing away with the `DataSourceFactory` in favor of adapters implementing the `DataAvailabilityProvider` trait themselves. Right now, the `DataSourceFactory` composes the various DA options like so:

```rs
struct DataSourceFactory<C, B, PIF, I>
where
    C: ChainProvider + Send + Clone,
    B: BlobProvider + Clone,
    PIF: PlasmaInputFetcher<C> + Clone,
    I: Iterator<Item = Bytes> + Send + Clone,
{ /* ... */ }
```

Removing the dependency on the `DataSourceFactory` as a composition type wrapping around the `DataAvailabilityProvider` trait lends towards more flexibility for other future alternative DA adapter implementors, given that `kona-derive`'s business logic only depends on the `DataAvailabilityProvider` trait. This also allows us to completely decouple alt data availability modifications from `kona-derive`, and allow for them to exist as extensions (think: `revm` + the handler abstractions, allowing for modifications to the base EVM logic). Other consumers of the crate can make new extensions without having to upstream changes to `kona-derive` for their modifications.

This also makes composition a bit cleaner for the consumer of the library - for example, in the `online` feature of `kona-derive`'s constructor, we can use the concrete implementations of the various online traits, as well as abstract over the DA (useful, for example, in a rollup client that may support several DAs through configuration at a higher level):

```rs
/// Creates a new online stack.
#[cfg(feature = "online")]
pub fn new_online_stack<DAP>(
    rollup_config: Arc<RollupConfig>,
    chain_provider: AlloyChainProvider<ReqwestProvider>,
    dap_source: DAP,
    fetcher: AlloyL2ChainProvider<ReqwestProvider>,
    builder: StatefulAttributesBuilder<
        AlloyChainProvider<ReqwestProvider>,
        AlloyL2ChainProvider<ReqwestProvider>,
    >,
) -> impl NextAttributes + ResettableStage + Debug + Send
where
    DAP: DataAvailabilityProvider + Debug + Send,
{ /* ... */ }
```

Note that above, we no longer need to supply types for DA options we are not using.

### Other File Jumblin'

1. `kona-plasma` now contains _all_ logic related to the alternative Plasma DA extension for `kona-derive`.
2. `kona-providers` has been moved back into `kona-derive`; These providers are specific to the pipeline and pretty minimal interfaces of the full RPCs, shouldn't be useful except to consumers of `kona-derive`.


crates/plasma/src/source.rs
```diff
@@ -89,14 +89,14 @@ where
...
-                    return Some(Err(StageError::Plasma(e)));
+                    return Some(Err(StageError::Custom(anyhow!(e))));
                 }
             };
 
             // If the data is empty,
             if data.is_empty() {
                 tracing::warn!("empty data from plasma source");
-                return Some(Err(StageError::Plasma(PlasmaError::NotEnoughData)));
+                return Some(Err(StageError::Custom(anyhow!(PlasmaError::NotEnoughData))));
```
refcell, line 99: Ugh I really dislike `StageError` leaking it's way out of the kona-derive crate. We should fix this in a follow up issue.

crates/plasma/Cargo.toml
```diff
@@ -15,7 +15,7 @@ tracing.workspace = true
 
 # Local
 kona-primitives = { path = "../primitives" }
-kona-providers = { path = "../providers" }
+kona-derive = { path = "../derive" }
```
refcell, line 18: I'd like to avoid importing `kona-derive` in `kona-plasma`. Feels leaky.

crates/plasma/src/source.rs
clabby: Gotcha! Yeah, totally open to make it not leak out if it's possible. 

This is also basically a substitute component of the derivation pipeline, so maybe it makes sense?

crates/plasma/Cargo.toml
clabby: But `kona-plasma` is an extension of `kona-derive`?

crates/derive/src/types/errors.rs
```diff
@@ -5,7 +5,6 @@ use crate::types::{BlockID, Frame};
 use alloc::vec::Vec;
 use alloy_primitives::{Bytes, B256};
 use core::fmt::Display;
-use kona_plasma::types::PlasmaError;
```
refcell, line 8: Gotta admit, removing this is very nice.

refcell (approved):
This is fantastic work, thank you @clabby

crates/plasma/Cargo.toml
refcell: Discussed offline - this is the dependency graph we will go with for now. Since the dependency is not a component but a trait, I think it's reasonable not to refactor into a third node in the dependency graph, concretely a new crate, and add edges from both `kona-derive` and `kona-plasma` to that third node (eg `kona-providers`).

PR #159, opened 2024-04-28
fix(derive): Ethereum Data Source

refcell:
**Description**

Fixes the ethereum data source to default to a calldata-based provider if the ecotone timestamp is not set, following the [reference implementation](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/data_source.go#L75-L82).

### DIFF
--- a/crates/derive/src/sources/ethereum.rs
+++ b/crates/derive/src/sources/ethereum.rs
@@ -8,7 +8,7 @@ use crate::{
 };
 use alloc::{boxed::Box, fmt::Debug};
 use alloy_primitives::{Address, Bytes};
-use anyhow::{anyhow, Result};
+use anyhow::Result;
 use async_trait::async_trait;
 
 /// A factory for creating an Ethereum data source provider.
@@ -19,9 +19,9 @@ where
     B: BlobProvider + Clone,
 {
     /// The chain provider to use for the factory.
-    pub chain_provider: Option<C>,
+    pub chain_provider: C,
     /// The blob provider
-    pub blob_provider: Option<B>,
+    pub blob_provider: B,
     /// The ecotone timestamp.
     pub ecotone_timestamp: Option<u64>,
     /// The L1 Signer.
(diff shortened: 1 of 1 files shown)


refcell:
* **#159** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/kona/159?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/derive/src/sources/ethereum.rs
```diff
@@ -19,9 +19,9 @@ where
     B: BlobProvider + Clone,
 {
     /// The chain provider to use for the factory.
-    pub chain_provider: C,
+    pub chain_provider: Option<C>,
     /// The blob provider
-    pub blob_provider: B,
+    pub blob_provider: Option<B>,
```
clabby, line 24: Hmm - all Ethereum providers should need both a chain provider and a blob source. Don't think we should make these optional?

refcell: What if someone wants to configure a chain to have only one or the other?

clabby: Good q - I think that this diverges from Ethereum DA-configured OP Stack chains, and should probably require a new data source implementation (which thankfully isn't hard now! Should def document this). The OP Stack specification describes a clear transition between calldata-available batches and blob-available batches with the transition from the Delta -> Ecotone hardfork, so I think the default source we provide should meet that spec.

clabby: Reflecting that this is an OP Stack Ethereum DA in the type name might make this more clear. When I said 
>all Ethereum providers

I think I more meant "all OP Stack Ethereum DA providers"

refcell: That's fine with me - you're right they can just define a new non-ethereum data source. Though the "Ethereum" data source name is then a misnomer - it should be `OPStackDataSource` ;P

clabby (approved):

PR #163, opened 2024-04-28
feat(primitives): move attributes into primitives

N0xMare:
Overview: migrate attributes from kona-derive types into kona-primitives


crates/primitives/src/lib.rs
```diff
@@ -35,5 +35,8 @@ pub use payload::{
 pub mod rollup_config;
 pub use rollup_config::RollupConfig;
 
+mod attributes;
 pub mod system_config;
+pub use attributes::{L2AttributesWithParent, L2PayloadAttributes};
+
 pub use system_config::{SystemAccounts, SystemConfig, SystemConfigUpdateType};
```
refcell, line 42: Can we just order these to follow convention

```suggestion
pub mod attributes;
pub use attributes::{L2AttributesWithParent, L2PayloadAttributes};

pub mod system_config;
pub use system_config::{SystemAccounts, SystemConfig, SystemConfigUpdateType};
```

refcell (approved):

PR #175, opened 2024-05-05
fix(derive): Inline Blob Validation

refcell:
**Description**

Blob processing is specific to the functionality of the `BlobProvider`. This pr refactors the functional blob validation method into a type method and inline use.


clabby (approved):
nice!

PR #182, opened 2024-05-26
fix(primitives): use decode_2718() to gracefully handle the tx type

### LINKED ISSUE #147: fix(derive): L2ExecutionPayload Deposit Transaction Decoding
**Description**

When converting the `L2ExecutionPayload` into an `L2BlockInfo`, if the execution payload's block number is not genesis, it should decode the l1 block info deposit transaction from the execution payloads transactions. The l1 block info deposit transaction is always the first transaction and it's calldata contains the l1 origin and sequence number used for initializing an `L2BlockInfo`.

Currently, the decoding manually checks the first byte against the deposit type and then decodes all trailing data.

It should instead push all decoding logic onto the `OpTxEnvelope` type and just perform an eip2718 typed decode using the `OpTxType::Deposit` type.

This will likely require modifications to the `op_alloy_consensus` dependency to modify the `OpTxEnvelope`'s `typed_decode` method or the `TxDeposit`'s `decode` method to strip the deposit type prefix.

merklefruit:
**Description**

I believe `decode_2718()` will gracefully remove the transaction type byte, so I imported it from alloy_eips.

**Tests**

I ran all existing tests, including all the decoding unit tests in `stages`.

**Metadata**

- Fixes #147

### DIFF
--- a/crates/primitives/Cargo.toml
+++ b/crates/primitives/Cargo.toml
@@ -16,6 +16,7 @@ alloy-primitives = { workspace = true, features = ["rlp"] }
 
 # Alloy Types
 alloy-sol-types = { version = "0.7.1", default-features = false }
+alloy-eips = { git = "https://github.com/alloy-rs/alloy", rev = "cb95183", default-features = false }
 op-alloy-consensus = { git = "https://github.com/clabby/op-alloy", branch = "refcell/consensus-port", default-features = false }
 
 # `serde` feature dependencies
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1664,6 +1664,7 @@ name = "kona-primitives"
 version = "0.0.1"
 dependencies = [
  "alloy-consensus 0.1.0 (git+https://github.com/alloy-rs/alloy?rev=e3f2f07)",
+ "alloy-eips 0.1.0 (git+https://github.com/alloy-rs/alloy?rev=cb95183)",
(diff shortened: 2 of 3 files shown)


clabby:
Rebased on #186 - Closing & re-opening to bump CI :)

crates/primitives/Cargo.toml
```diff
@@ -16,6 +16,7 @@ alloy-primitives = { workspace = true, features = ["rlp"] }
 
 # Alloy Types
 alloy-sol-types = { version = "0.7.1", default-features = false }
+alloy-eips = { git = "https://github.com/alloy-rs/alloy", rev = "cb95183", default-features = false }
```
clabby, line 19: We've been pinned to `e3f2f07` in other crates that are using `alloy-rs/alloy` - mind updating?

```suggestion
alloy-eips = { git = "https://github.com/alloy-rs/alloy", rev = "e3f2f07", default-features = false }
```

merklefruit: Unfortunately build fails with that rev :( 

I used the rev `cb95183` because it's the one pinned inside op-alloy-consensus.
Want me to open a PR on your fork of op-alloy instead?

clabby: Ah I'll update that later - for now this is fine!

PR #195, opened 2024-05-31
feat(mpt): Trie node insertion

clabby:
## Overview

Adds the logic required to insert a given value into a `TrieNode` at a passed path.

### DIFF
--- a/crates/mpt/src/node.rs
+++ b/crates/mpt/src/node.rs
@@ -62,14 +62,14 @@ pub enum TrieNode {
     /// A leaf node is a 2-item node with the encoding `rlp([encoded_path, value])`
     Leaf {
         /// The key of the leaf node
-        key: Bytes,
+        prefix: Nibbles,
         /// The value of the leaf node
         value: Bytes,
     },
     /// An extension node is a 2-item pointer node with the encoding `rlp([encoded_path, key])`
     Extension {
         /// The path prefix of the extension
-        prefix: Bytes,
+        prefix: Nibbles,
         /// The pointer to the child node
         node: Box<TrieNode>,
     },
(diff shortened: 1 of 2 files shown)


clabby:
* **#196** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/kona/196?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#195** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/kona/195?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/mpt/src/node.rs
```diff
@@ -169,33 +141,26 @@ impl TrieNode {
...
             }
-            TrieNode::Leaf { key, value } => {
-                let key_nibbles = Nibbles::unpack(key.clone());
-                let shared_nibbles = key_nibbles[1..].as_ref();
-
+            TrieNode::Leaf { prefix, value } => {
                 // If the key length is one, it only contains the prefix and no shared nibbles.
                 // Return the key and value.
-                if key.len() == 1 || nibble_offset + shared_nibbles.len() >= path.len() {
+                if prefix.len() == 1 || nibble_offset + prefix.len() >= path.len() {
```
refcell, line 147: This seems to change the accounting here where `shared_nibbles` was 1 less than the now `prefix.len()`. Is this correct?

clabby: Yes great spot - you are correct that this is a bug. This is fixed in #196 @ https://github.com/ethereum-optimism/kona/pull/196/files#diff-f3393c37018f3745aed7a3c1f44de945d0804f997516c18372d692d48c39a13aL145-R145

refcell: Ah thank you!

crates/mpt/src/node.rs
```diff
@@ -361,6 +495,20 @@ fn rlp_list_element_length(buf: &mut &[u8]) -> alloy_rlp::Result<usize> {
...
+/// - `first` - first nibble of the path if it is odd. Must be <= 0x0F, or will create invalid
+///   nibbles.
+/// - `rest` - rest of the nibbles packed
+///
+/// ## Returns
+/// - `Nibbles` - unpacked nibbles
+fn unpack_path_to_nibbles(first: Option<u8>, rest: &[u8]) -> Nibbles {
+    let rest = Nibbles::unpack(rest);
+    Nibbles::from_vec_unchecked(first.into_iter().chain(rest.iter().copied()).collect::<Vec<u8>>())
+}
```
refcell, line 510: Wonder if this should just be a macro. Preference?

refcell: Would be neat to do: `nibbles!("646f")`

refcell (approved):

crates/mpt/src/node.rs
clabby: Ah yeah, maybe not for this function since it's specific to unpacking path nibbles, but that'd be nice. Maybe we should make a PR into `nybbles` for that :D

PR #217, opened 2024-06-06
feat: Pipeline Builder

refcell:
**Description**

Implements a pipeline builder with a doc comment example using online provider implementations.

Also updates the online blob provider to remove the unused RPC client.


clabby (approved):
Very nice

PR #229, opened 2024-06-10
fix(kona-derive): Use Upstream ReqwestProvider

refcell:
**Description**

Now that alloy dependencies are updated to use revision [`cb95183`](https://github.com/alloy-rs/alloy/blob/cb95183d477024b57b27336035d10403e8ba55b8/crates/provider/src/lib.rs#L25), online `kona-derive` provider implementations can reuse the exported `ReqwestProvider` type instead of defining a custom `ReqwestClient` type.

In a later revision we won't need to specify the return type or wrap the method str in a Borrowed Cow. For now though, we comply with the compiler's demands.

### DIFF
--- a/crates/derive/src/online/alloy_providers.rs
+++ b/crates/derive/src/online/alloy_providers.rs
@@ -11,16 +11,13 @@ use crate::{
 use alloc::{boxed::Box, sync::Arc, vec::Vec};
 use alloy_consensus::{Header, Receipt, ReceiptWithBloom, TxEnvelope, TxType};
 use alloy_primitives::{Bytes, B256, U64};
+use alloy_provider::{Provider, ReqwestProvider};
 use alloy_rlp::{Buf, Decodable};
-use alloy_rpc_client::RpcClient;
+use alloy_transport::TransportResult;
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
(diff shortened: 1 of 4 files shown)


refcell:
* **#229** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/kona/229?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/derive/src/online/alloy_providers.rs
```diff
@@ -72,8 +69,9 @@ impl ChainProvider for AlloyChainProvider {
             return Ok(header.clone());
         }
 
-        let raw_header: Bytes =
-            self.inner.request("debug_getRawHeader", [hash]).await.map_err(|e| anyhow!(e))?;
+        let method = alloc::borrow::Cow::Borrowed("debug_getRawHeader");
```
clabby, line 72: Can we just `.into()` on `"<method>"` rather than doing the long instantiation on all of these? the clone on write object has a direct `From<&'static str>` impl [here](https://doc.rust-lang.org/stable/std/borrow/enum.Cow.html#impl-From%3C%26str%3E-for-Cow%3C'a,+str%3E)

clabby (approved):

PR #230, opened 2024-06-10
feat(kona-derive): Updated interface

refcell:
**Description**

Updates the `Pipeline` interface to include a `set_origin` method and updates the `reset` method to perform the pipeline reset.


refcell:
* **#230** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/kona/230?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/derive/src/online/validation.rs
```diff
@@ -0,0 +1,92 @@
...
+        let block = block.map_err(|e| anyhow::anyhow!(e))?;
+        Ok(block)
+    }
+
+    /// Gets the payload for the specified [BlockNumberOrTag].
+    pub(crate) async fn get_payload(&self, tag: BlockNumberOrTag) -> Result<L2PayloadAttributes> {
+        // TODO: we can't use the provider's get_block_by_number here because
+        // upstream alloy will return Mainnet Ethereum transactions and not work for Optimism
+        // Deposit Transactions.
+        let block = self.get_block(tag).await?;
```
clabby, line 55: This will not work for L2 blocks, since the `N = Ethereum` bound does not support OP transactions. Longer term I think we want to implement the `Network` trait for OP, though for now, can use `debug_*` endpoints to get the raw tx RLP and decode into `OpTxEnvelope`.

crates/derive/src/online/validation.rs
```diff
@@ -0,0 +1,92 @@
...
+        let transactions = match block.transactions {
+            BlockTransactions::Full(txns) => txns,
+            _ => {
+                return Err(anyhow::anyhow!("Block {tag} missing full transactions"));
+            }
+        };
+        let transactions = transactions
+            .iter()
+            .map(|tx| {
+                serde_json::to_vec(&tx)
```
clabby, line 65: This is the wrong encoding; The `Vec<RawTransaction>` is a vector of EIP-2718 encoded txs. If we end up going w/ `debug_getRawTransaction` in the short term, should just be able to pass the result straight in. If we go ahead and implement `Network` for OP, we'll need to use `encode_2718` on the `OpTxEnvelope`s within the block.

crates/derive/src/online/validation.rs
```diff
@@ -0,0 +1,94 @@
...
+        let (header, transactions) = self.get_block(tag).await?;
+        Ok(L2PayloadAttributes {
+            timestamp: header.timestamp,
+            prev_randao: header.mix_hash.unwrap_or_default(),
+            fee_recipient: header.miner,
+            // Withdrawals on optimism are always empty
+            withdrawals: Default::default(),
+            parent_beacon_block_root: Some(header.parent_hash),
+            transactions,
+            no_tx_pool: false,
```
clabby, line 80:
```suggestion
            no_tx_pool: true,
```

clabby (approved):
Very nice mr wick

PR #243, opened 2024-06-11
feat(kona-derive): Towards Derivation

### LINKED ISSUE #34: dep: `no_std` derivation pipeline
## Overview

In order to develop the `client` program, we need a `no_std` compatible derivation pipeline. Ideally, the derivation pipeline implementation consists of pure functions that take in L1 inputs and produce L2 transactions, without any opinions on where the data comes from. At its core, this derivation pipeline implementation should be fully synchronous. The fault proof program will provide data directly to this derivation pipeline, and receipts and other L1 data from the `host` program through the `PreimageOracle` as it processes the L1 block range.

### LINKED ISSUE #111: chore(derive): Strict Log and Receipt Ordering
**Description**

When deriving deposit transactions from receipts, the Attributes Queue uses block logs to decode Deposit objects from events.

`kona-derive` uses the [`alloy_primitives::Log`](https://docs.rs/alloy-primitives/latest/alloy_primitives/struct.Log.html) object with it's associated type being [`alloy_primitives::LogData`](https://docs.rs/alloy-primitives/latest/alloy_primitives/struct.LogData.html) which contain insufficient Log data to reconstruct the deposit. With the alloy primitive log type, there is the address, raw log data, and log topics.

In order to fully derive the [`TxDeposit`](https://github.com/clabby/op-alloy/blob/refcell/consensus-port/crates/op-consensus/src/transaction/optimism.rs#L8-L34), we generate the source hash using the block hash and log index. (See the [`UserDepositSource`](https://github.com/ethereum-optimism/kona/blob/main/crates/derive/src/types/deposits.rs#L47-L54) type).

The inconsistency here arrises since the [go-ethereum `Log` type](https://github.com/ethereum/go-ethereum/blob/master/core/types/log.go#L29) is populated with the `Index` field, denoting the index of the log within the block. But the [`alloy_primitives`](https://docs.rs/alloy-primitives/latest/alloy_primitives/index.html) types do not have this index.

This ticket is to explore whether logs and receipts are guaranteed to be in order. Once resolved, the inline `TODO` comment in the [attributes queue stage](https://github.com/ethereum-optimism/kona/blob/main/crates/derive/src/stages/attributes_queue/deposits.rs#L28-L29) can be removed.

### LINKED ISSUE #245: chore(kona-sync): Migrate Binary to Example
**Description**

This is a small ticket to migrate the `kona-sync` binary to an example.

It should be moved from `bin/sync` to a new `examples` directory and be able to be run with: `cargo run --example kona-sync ...`.

### LINKED ISSUE #246: fix(kona-derive): Fix Batch Tests
**Description**

In #244, Single Batch decoding was fixed to include the RLP Header. We've moved away from manually implementing the RLP `Encodable` and `Decodable` traits to using `alloy_rlp`'s `RlpEncodable` and `RlpDecodable` derive impls.

In so doing, the serialized batch testdata became invalid since it was serialized without the rlp header. The raw batch data needs to be udpated by encoding a valid batch. The tests should then _just work_.

Output: No tests are ignored when running `just test`.

refcell:
**Description**

- Small fixes to fetch the genesis block info for the L1 tip and L2 cursor.
- Dogfoods [superchain-registry primitive type bindings][scr] for types.
- Implements the "trusted-sync" example that runs the derivation pipeline for sync testing.

[scr]: https://github.com/ethereum-optimism/superchain-registry/tree/main/bindings/rust-primitives

**Metadata**

Fixes #245 
Fixes #246 
Fixes #34 
Fixes #111


refcell:
* **#243** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/kona/243?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/derive/src/pipeline/builder.rs
```diff
@@ -32,9 +32,8 @@ where
     dap_source: Option<D>,
     chain_provider: Option<P>,
     builder: Option<B>,
+    origin: Option<BlockInfo>,
```
clabby, line 35: hmm - I don't think we should ever be setting the L1 origin of the pipeline manually, yeah? We should initialize it with an L2 safe head, and find the L1 origin through the L2 chain provider?

crates/derive/src/stages/batch_queue.rs
```diff
@@ -436,24 +436,24 @@ mod tests {
...
 
     fn new_batch_reader() -> BatchReader {
-        let raw_data = include_bytes!("../../testdata/raw_batch.hex");
-        let mut typed_data = vec![BatchType::Span as u8];
-        typed_data.extend_from_slice(raw_data.as_slice());
-        let compressed = compress_to_vec_zlib(typed_data.as_slice(), 5);
-        BatchReader::from(compressed)
+        let file_contents =
+            alloc::string::String::from_utf8_lossy(include_bytes!("../../testdata/batch.hex"));
+        let file_contents = &(&*file_contents)[..file_contents.len() - 1];
```
clabby, line 453: Can we re-encode the .hex file as binary, rather than a UTF-8 hex string?

crates/derive/src/stages/batch_queue.rs
```diff
@@ -643,7 +644,7 @@ mod tests {
         assert!(logs[1].contains("Deriving next batch for epoch: 16988980031808077784"));
         let warns = trace_store.get_by_level(Level::WARN);
         assert_eq!(warns.len(), 1);
-        assert!(warns[0].contains("batch is for future epoch too far ahead, while it has the next timestamp, so it must be invalid"));
+        assert!(warns[0].contains("span batch has no new blocks after safe head"));
```
clabby, line 647: These are fundamentally different errors. What changed here?

crates/derive/src/stages/channel_reader.rs
```diff
@@ -172,15 +174,15 @@ impl<T: Into<Vec<u8>>> From<T> for BatchReader {
...
     use alloc::vec;
-    use miniz_oxide::deflate::compress_to_vec_zlib;
 
     fn new_compressed_batch_data() -> Bytes {
-        let raw_data = include_bytes!("../../testdata/raw_batch.hex");
-        let mut typed_data = vec![BatchType::Span as u8];
-        typed_data.extend_from_slice(raw_data.as_slice());
-        compress_to_vec_zlib(typed_data.as_slice(), 5).into()
+        let file_contents =
+            alloc::string::String::from_utf8_lossy(include_bytes!("../../testdata/batch.hex"));
```
clabby, line 182: ibid on binary encoding

examples/trusted-sync/src/validation.rs
```diff
@@ -26,18 +16,20 @@ pub trait Validator {
...
+    /// The canyon activation timestamp.
+    canyon_activation: u64,
 }
 
 impl OnlineValidator {
     /// Creates a new `OnlineValidator`.
-    pub fn new(provider: ReqwestProvider) -> Self {
-        Self { provider }
+    pub fn new(provider: ReqwestProvider, canyon: u64) -> Self {
+        Self { provider, canyon_activation: canyon }
```
clabby, line 26: Can we pass the `RollupConfig` instead of just the canyon activation timestamp?

clabby (commented):
This is looking great. Couple final nits, then let's get this in to keep iterating 😄

crates/derive/src/pipeline/builder.rs
refcell: This is just a matter of prefetching the data. I'd prefer to keep the pipeline construction synchronous and not embed this async fetching of the associated l1 origin.

clabby (approved):

PR #248, opened 2024-06-13
fix: output root version to 32 bytes

zobront:
**Description**

The `compute_output_root()` function set the `OUTPUT_ROOT_VERSION` to just 1 byte, when it should be 32 bytes per [the spec](https://specs.optimism.io/protocol/proposals.html#l2-output-commitment-construction). 

This PR corrects the calculation by left padding the `OUTPUT_ROOT_VERSION` with 31 bytes of 0s.

**Tests**

None of the existing block tests align with known output roots on chain, so will require aggregating new test data. I can use `debug` RPC calls to get the header and transactions, but not sure about setting up the testdata to act as the fetcher. Do you have a workflow for putting that together?

**Additional context**

N/A

**Metadata**

N/A


zobront:
@clabby Put reminder to add a test later, ready to merge.

PR #282, opened 2024-06-19
fix(common): Pipe IO support

clabby:
## Overview

Alters the native IO to support Golang's `os.Pipe`, intended to allow for the `host` program in server mode to be able to communicate with the file descriptors passed from the VM.

### DIFF
--- a/bin/programs/client/src/kona.rs
+++ b/bin/programs/client/src/kona.rs
@@ -21,7 +21,7 @@ extern crate alloc;
 /// The size of the LRU cache in the oracle.
 const ORACLE_LRU_SIZE: usize = 1024;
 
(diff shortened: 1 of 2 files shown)


bin/programs/client/src/kona.rs
```diff
@@ -21,7 +21,7 @@ extern crate alloc;
 /// The size of the LRU cache in the oracle.
 const ORACLE_LRU_SIZE: usize = 1024;
 
-#[client_entry(0x77359400)]
+#[client_entry(100_000_000)]
```
clabby, line 24: Note - we're gonna have to play with this, but 100MB is far more than we need on avg.

refcell (approved):

PR #295, opened 2024-06-21
[CLOSED] feat(client): providers generic over oracles

zobront:
**Description**

This PR includes a range of changes that will make it easiest for our `zkvm-client` (and likely future integrating teams) to work with `kona-client` directly rather than forking and modifying it.

Changes:
- The `CommsClient` trait encompasses `PreimageOracleClient + Clone + HintWriterClient`. Therefore anything that implements this trait (which will be the Oracles) can both fetch and hint data.
- Update `CachingOracle` to include the hint writer, and therefore properly implement `CommsClient`.
- Providers perform all hinting and fetching via `self.oracle` (which implements `CommsClient`) and therefore never need to access the hint writer directly. This includes `TrieDBHinter`, which is now implemented on L2Provider rather than requiring the `TrieDBHintWriter` type.
- A zkVM io type is added that is unimplemented. This doesn't do anything because we don't use the io system at all in the zkVM, but is required for compilation. 
- The PreimageKey type in the `preimage` crate now has an `rkyv` flag for serialization instead of `serde`. This is used for more efficient serialization that is better for us when passing data into the zkVM.

**Tests**

N/A

**Additional context**

N/A

**Metadata**

N/A


crates/common/src/io.rs
```diff
@@ -12,6 +12,9 @@ cfg_if! {
     } else if #[cfg(target_arch = "riscv64")] {
         #[doc = "Concrete implementation of the [BasicKernelInterface] trait for the `riscv64` target architecture."]
         pub type ClientIO = crate::asterisc::io::AsteriscIO;
+    } else if #[cfg(target_os = "zkvm")] {
+        #[doc = "Concrete implementation of the [BasicKernelInterface] trait for the `zkvm` target architecture."]
+        pub type ClientIO = crate::zkvm::io::ZkvmIO;
```
zobront, line 17: I would like to get rid of the need for any of these changes, but putting them in here for now so it compiles and works.

crates/executor/src/lib.rs
```diff
@@ -57,6 +57,7 @@ where
     pub fn new(
         config: &'a RollupConfig,
         parent_header: Sealed<Header>,
+        // TODO: @Clabby - would you rather refactor this to assume Provider implements F + H and just pass one argument?
```
zobront, line 60: Question for you @clabby: We can keep this as fetcher F and hinter H and just use the provider for both, or we can refactor the executor to expect the provider to handle everything. Which would you prefer?

Same question with the TrieDB, which also expects separate fetcher and hinter.

crates/common/src/io.rs
zobront: Totally right. Clabby and I brainstormed a few alternatives to remove the need for it, since we don't actually use the io system so it's not doing anything.

He's thinking through it more, but my perspective is that best option would be to create a no op BKI implementation and organize the feature flags and defaults in such a way that it typically defaults to NativeIO, but that we can import with default features off to fall back to the no op.

zobront:
I cannot get CI to run, so going to close this PR and reopen a new one to see if it fixes...

PR #299, opened 2024-06-22
feat(build): Dockerize trusted-sync

refcell:
**Description**

Dockerizes the `trusted-sync` example to run as a part of infrastructure.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -1,0 +1,24 @@
+name: Release and version
+
+on:
+  workflow_dispatch:
+
+jobs:
+  publish:
+    name: Publish Trusted Sync
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: docker/setup-buildx-action@v3
+      - name: Login to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.PERSONAL_DOCKERHUB_ACCESS_TOKEN_USERNAME }}
+          password: ${{ secrets.PERSONAL_DOCKERHUB_ACCESS_TOKEN_SECRET }}
+      - name: Build and Push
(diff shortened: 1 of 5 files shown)


refcell:
* **#299** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/kona/299?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

examples/trusted-sync/Dockerfile
```diff
@@ -0,0 +1,19 @@
...
+RUN rustup target add x86_64-unknown-linux-musl
+RUN apt-get update && apt-get install -y musl-tools
+
+COPY ./crates/ /kona/crates
+COPY ./bin/ /kona/bin
+COPY ./examples/ /kona/examples
+COPY ./Cargo.toml /kona
+COPY ./Cargo.lock /kona
+COPY ./rustfmt.toml /kona
+COPY ./README.md /kona
```
clabby, line 12: (dumb question) does glob matching not work, as to copy the full contents of the folder in one copy directive?

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,24 @@
+name: Release and version
```
clabby, line 1: Any reason we're not pushing to the `ghcr.io` registry?

Example of a release workflow for GHCR, used for our `build` images: https://github.com/ethereum-optimism/kona/blob/main/.github/workflows/cross_docker.yaml

examples/trusted-sync/Dockerfile
refcell: We can add a dockerignore - this was just so we don't copy unneeded build artifacts. Ended up taking way less time since things like `target` were not copied. I don't think it's bad practice to be deliberate about what you're placing into an image like this.

clabby (approved):

PR #308, opened 2024-06-24
feat: trusted sync metrics

refcell:
feat: trusted sync metrics

feat: online derive provider metrics

fix: chain provider metrics

feat(derive): blob provider metrics


crates/derive/Cargo.toml
```diff
@@ -39,6 +39,8 @@ alloc-no-stdlib = "2.0.4"
 serde = { version = "1.0.203", default-features = false, features = ["derive"], optional = true }
 
 # `online` feature dependencies
+lazy_static = "1.5.0"
```
clabby, line 42: Can we make this optional, gated behind `online`?

clabby (approved):
Looking great! One nit on the `lazy_static` dependency, think that can be behind the `online` feature flag.

PR #336, opened 2024-06-27
feat(client): providers generic over oracles

zobront:
**Description**

This PR includes a range of changes that will make it easier for our `zkvm-client` (and likely future integrating teams) to work with `kona-client` directly rather than forking and modifying it.

Changes:
- The `CommsClient` trait encompasses `PreimageOracleClient + Clone + HintWriterClient`. Therefore anything that implements this trait (which will be the Oracles) can both fetch and hint data.
- Update `CachingOracle` to include the hint writer, and therefore properly implement `CommsClient`.
- Providers perform all hinting and fetching via `self.oracle` (which implements `CommsClient`) and therefore never need to access the hint writer directly. This includes `TrieDBHinter`, which is now implemented on L2Provider rather than requiring the `TrieDBHintWriter` type.
- A zkVM io type is added that is unimplemented. This doesn't do anything because we don't use the io system at all in the zkVM, but is required for compilation. 
- The PreimageKey type in the `preimage` crate now has an `rkyv` flag for serialization instead of `serde`. This is used for more efficient serialization that is better for us when passing data into the zkVM.

**Tests**

N/A

**Additional context**

N/A

**Metadata**

N/A


bin/client/src/kona.rs
```diff
@@ -60,8 +60,8 @@ fn main() -> Result<()> {
         let mut executor = StatelessL2BlockExecutor::new(
             &boot.rollup_config,
             driver.take_l2_safe_head_header(),
+            l2_provider.clone(),
```
zobront, line 63: @clabby My only question here is if we're confident enough in this "provider owns oracle" pattern to avoid the clone by changing this function to just take in a `provider` instead of both a `fetcher` and `hinter`?

It passes them to another function further downstream, so your call whether we want to make that change across the board or just use this clone and leave the other functions as is.

bin/client/src/kona.rs
```diff
@@ -57,12 +57,14 @@ fn main() -> Result<()> {
         .await?;
         let L2AttributesWithParent { attributes, .. } = driver.produce_disputed_payload().await?;
 
-        let precompile_overrides =
-            FPVMPrecompileOverride::<OracleL2ChainProvider, TrieDBHintWriter>::default();
+        let precompile_overrides = FPVMPrecompileOverride::<
+            OracleL2ChainProvider<CachingOracle>,
+            OracleL2ChainProvider<CachingOracle>,
+        >::default();
```
zobront, line 63: @clabby I understand the logic for keeping a separate Fetcher and Hinter in the library crates, as we want to keep flexibility for future users to implement these separately.

But if we are going to have the L2 Provider fulfill both of these roles in `kona-client`, do you think it makes sense to refactor `FPVMPrecompileOverride` to just be generic over a single type that implements both traits?

crates/preimage/src/key.rs
```diff
@@ -56,8 +57,9 @@ impl TryFrom<u8> for PreimageKeyType {
 /// |---------|-------------|
 /// | [0, 1)  | Type byte   |
 /// | [1, 32) | Data        |
-#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
```
refcell, line 59: Can we keep `serde` here as well?

refcell: Maybe we could add a new `rkyv` feature flag? We use `serde` throughout the workspace so it would be nice to keep consistency

crates/preimage/Cargo.toml
```diff
@@ -18,13 +18,12 @@ async-trait.workspace = true
...
-serde = { version = "1.0.203", features = ["derive"], optional = true }
+rkyv = { version = "0.7.44", optional = true }
 
 [dev-dependencies]
 tokio = { version = "1.38.0", features = ["full"] }
 tempfile = "3.10.1"
 
 [features]
 default = []
-serde = ["dep:serde"]
```
refcell, line 30: Let's keep both `serde` and `rkyv`. Can have a mutual exclusion compile error like so in `lib.rs`.

```
#[cfg(all(feature = "serde", feature = "rkyv"))]
compile_error!("serde and rkyv are mutually exclusive feature flags and cannot be enabled together");
```

bin/client/src/kona.rs
refcell: Since this simplifies the ux and isn't clear who will need the split logic, I support doing this refactor in a follow-on pr. The FPVM Precompile Override can always be made even more abstract by encapsulating it's internal logic in a trait, so not too worried about increasing the abstractions here.

refcell (changes requested):
Generally looks good - just have a preference to keep `serde` as a feature flag with mutual exclusion.

refcell (approved):

PR #361, opened 2024-07-09
feat(host): refactor kona-host into library

zobront:
**Description**

It is useful to be able to access the functionality of `kona-host` from an outside crate, so that projects leveraging kona can invoke it within their flow. For example, in the ZK context, we use `kona-host` out of the box in our witness generation step.

This PR moves the core `kona-host` functionality into a library, and refactors `main.rs` to use this library.

**Tests**

N/A - existing tests should ensure that this is safe.

**Additional context**

N/A

**Metadata**

N/A


refcell (approved):
This looks good, nice work.

PR #368, opened 2024-07-11
feat(client, preimage, common): no-io flag

zobront:
**Description**

The `kona-common` crate determines which ClientIO to use based on the compilation environment. This crate is inherited by `kona-client` and `kona-preimage`, so in the case that any of these crates are imported, we try to compile ClientIO.

In the case that we aren't compiling to MIPS or RISCV64, we fallback to NativeIO, which fails in the ZKVM.

In the ZKVM, we don't use IO at all, so to avoid this issue we simply add a `no-io` feature flag to the 3 crates that avoids compiling anything that touches IO when called in this context. 

Then, in our downstream program, we set this flag depending on whether or not we are in the ZKVM context.

**Tests**

N/A

**Additional context**

N/A

**Metadata**

N/A

### DIFF
--- a/bin/client/Cargo.toml
+++ b/bin/client/Cargo.toml
@@ -41,6 +41,7 @@ serde_json = "1.0.117"
 
 [features]
 tracing-subscriber = ["dep:tracing-subscriber"]
+no-io = ["kona-common/no-io", "kona-preimage/no-io"]
 
 [[bin]]
 name = "kona"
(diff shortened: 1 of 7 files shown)


zobront:
@clabby This is failing because the linting test tests for `all-features`, but compilation fails when `no-io` flag is passed when it's needed. How do you think we should address?

clabby:
> @clabby This is failing because the linting test tests for `all-features`, but compilation fails when `no-io` flag is passed when it's needed. How do you think we should address?

Cargo features are meant to be additive; It's a bit hard to make mutually exclusive features. We'll need to split this up a bit further. Glad to work on this today.

refcell:
Will review once this is flipped to enabled-by-default `io` feature that can be toggled off.

refcell:
Closing this out in favor of #394

Issue #369, opened 2024-07-11
feat(host): fetcher builder to customize hint types & implementations


zobront:
**Is your feature request related to a problem? Please describe.**
Different users may have slightly different needs in terms of what data is saved by the fetcher. Currently, the only choice is to fork Kona and modify it. It would be helpful for users to be able to override hint implementations or even add new hint types when importing `zkvm-host` as a library.

**Describe the solution you'd like**
An interface similar to the accelerated precompiles where new hint types can be registered or overridden when building the fetcher.

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

clabby:
Taking this on this week or next - it'll be useful to have for the interop proof.

PR #370, opened 2024-07-11
feat(host): kzg proof for blobs

zobront:
**Description**

Currently, when the fetcher gathers blob data, it saves the preimages for each element in the blob, but does not save the KZG proof (because it is not needed in FPVM context). However, in the ZKVM context, we need this information.

This PR creates a short term workaround to save the kzg proof as a psuedo element 4096 in the Blob type (since the blob only uses elements 0 to 4095 inclusive).

For a longer term fix, I've opened this ticket for a feature to override fetcher implementations: https://github.com/ethereum-optimism/kona/issues/369

**Tests**

N/A

**Additional context**

N/A

**Metadata**

N/A

### DIFF
--- a/bin/host/src/fetcher/mod.rs
+++ b/bin/host/src/fetcher/mod.rs
@@ -219,6 +219,19 @@ where
                         sidecar.blob[(i as usize) << 5..(i as usize + 1) << 5].to_vec(),
                     );
                 }
+
+                // Write the KZG Proof as the 4096th element.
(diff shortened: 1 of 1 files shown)


clabby (approved):
LGTM. Before merging, can we please put up an issue to further abstract the host? This extra piece of data eventually should be able to be added by you guys downstream.

zobront:
Issue is already there. It’s linked in this PR description (#369).

PR #372, opened 2024-07-11
fix(examples): Reduce Origin Advance to Warn

refcell:
**Description**

Reduces the origin advance error to warning.

If the next block isn't available causing an intended block fetch failure, this will error which happens often.

We should also work in some backoffs to the providers, especially for frequently repeated method calls like origin advancing.

### DIFF
--- a/examples/trusted-sync/src/main.rs
+++ b/examples/trusted-sync/src/main.rs
@@ -122,7 +122,7 @@ async fn sync(cli: cli::Cli) -> Result<()> {
             }
             StepResult::OriginAdvanceErr(e) => {
                 metrics::PIPELINE_STEPS.with_label_values(&["origin_advance_failure"]).inc();
-                warn!(target: "loop", "Error advancing origin: {:?}", e);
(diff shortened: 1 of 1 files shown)


refcell:
* **#372** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/kona/372?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

examples/trusted-sync/src/main.rs
```diff
@@ -122,7 +122,7 @@ async fn sync(cli: cli::Cli) -> Result<()> {
             }
             StepResult::OriginAdvanceErr(e) => {
                 metrics::PIPELINE_STEPS.with_label_values(&["origin_advance_failure"]).inc();
-                error!(target: "loop", "Error advancing origin: {:?}", e);
+                warn!(target: "loop", "Error advancing origin: {:?}", e);
```
clabby, line 125: "Could not advance origin" w/ a reason maybe?

clabby (approved):

PR #390, opened 2024-07-23
fix: pin two dependencies due to upstream semver issues

zobront:
**Description**

Two upstream dependencies had patch releases that changes interfaces. This causes `kona` to break if `cargo update` is called. Pinning these two dependencies on the previous versions fixes the issue.

**Tests**

N/A

**Additional context**

N/A

**Metadata**

N/A


refcell (approved):
Generally not a fan of pinning issues, we want to know if an upstream dependency changes and breaks logic when running a `cargo update`. But for now we can approve this and I will put up a ticket for us to fix this in the future.

zobront:
Closing this because CI checks aren't running.

PR #391, opened 2024-07-23
fix: pin two dependencies due to upstream semver issues

zobront:
**Description**

Two upstream dependencies had patch releases that changes interfaces. This causes `kona` to break if `cargo update` is called. Pinning these two dependencies on the previous versions fixes the issue.

**Tests**

N/A

**Additional context**

- `op-alloy-consensus` upgraded `alloy-eips` to `0.2.0` which broke interface.
- `superchain-primitives` changed `batcher_address` field to `batcher_addr`, which broke interface.

**Metadata**

N/A


refcell:
I'll fix `superchain-primitives` in another pr - actually working on that now :)

refcell (approved):

PR #394, opened 2024-07-23
feat: add zkvm target for io

zobront:
**Description**

As an alternative to #368, this PR involves adding a simple alternative IO target to `kona-common`, which is currently a no-op but might later be used for more advanced cycle tracking, etc.

**Tests**

N/A

**Additional context**

N/A

**Metadata**

N/A


refcell (approved):
LGTM

PR #395, opened 2024-07-24
fix(mpt): hint `TrieNode` in `insert`

ratankaliani:
**Description**

When executing multiple blocks with the Kona derive pipeline, we ran into an issue where a `TrieNode` was requested by the fetcher before it was hinted in the state root computation within `update_accounts`.


clabby:
I'd rather not merge this until we understand what's going on.

It's interesting that this fixes the issue, but it's also a bit odd that it does. When inserting a new node, we do always have the full path to its parent opened within the trie. This is why we don't need to send hints for individual trie nodes within `insert` during regular operation, since we've front-loaded the account/storage proof hint.

clabby:
Closing in favor of #396

PR #396, opened 2024-07-24
fix(executor): Don't hold onto intermediate execution cache across block boundaries

clabby:
## Overview

Drops the intermediate in-memory `revm` cache between blocks when executing multiple blocks in a row with the `StatelessL2BlockExecutor`.

This fixes a bug where, when executing multiple blocks, the client program does not reach out to the host for the path to the account in the state trie in the _new_ state root. While we do not need support for contiguous block execution in upstream, this allows downstream consumers like Succinct to perform this action with `kona-executor`.


refcell (approved):
Very nice

PR #400, opened 2024-07-26
fix(kona-derive): Remove SignedRecoverable Shim

refcell:
**Description**

Previously, building the `kona-derive` crate would fail with `--no-default-features` since the tx envelope's public key recovery method is only available when the `k256` feature is enabled on alloy consensus.

This removes the `SignedRecoverable` trait shim and requires that `k256` is enabled.

We will need to refactor `kona-derive` sources to be abstract over the signer recovery method so we're not locked into `k256`.


bin/client/Cargo.toml
```diff
@@ -29,7 +29,7 @@ kona-common-proc = { path = "../../crates/common-proc", version = "0.0.2" }
 kona-preimage = { path = "../../crates/preimage", version = "0.0.2" }
 kona-primitives = { path = "../../crates/primitives", version = "0.0.1" }
 kona-mpt = { path = "../../crates/mpt", version = "0.0.2" }
-kona-derive = { path = "../../crates/derive", default-features = false, version = "0.0.2" }
+kona-derive = { path = "../../crates/derive", default-features = false, features = ["k256"], version = "0.0.2" }
```
clabby, line 32: I'd lobby for enabling this by default in the dependency itself rather than as a `kona-derive` feature. For now, it should just always be on - no `k256` feature in `kona-derive.

crates/derive/src/sources/calldata.rs
```diff
@@ -23,6 +23,7 @@ where
     /// Block Ref
     block_ref: BlockInfo,
     /// The L1 Signer.
+    #[allow(dead_code)] // Allow dead code for non-k256 builds.
```
clabby, line 26:
```suggestion
```

clabby (approved):

crates/derive/src/sources/blobs.rs
```diff
@@ -27,6 +27,7 @@ where
     /// Block Ref
     block_ref: BlockInfo,
     /// The L1 Signer.
+    #[allow(dead_code)] // Allow dead code for non-k256 builds.
```
clabby, line 30:
```suggestion
```

PR #410, opened 2024-07-31
feat(derive+trusted-sync): online blob provider with fallback

### LINKED ISSUE #316: fix(trusted-sync): Fallback blob source
## Overview

Currently, the `trusted-sync` binary communicates directly with the beacon client to fetch blobs. When the chain reaches Ecotone activation, it will halt immediately due to standard L1 consensus layers' blob retention policy of 18 days.

To derive the chain entirely from genesis -> tip, we will need a fallback blob archiver.

merklefruit:
**Description**

Added a fallback online blob provider in the form of `OnlineBlobProviderWithFallback`.
It will first try to fetch blobs from the primary provider, and proceed with the fallback one if the primary fails.

I added the `BlobSidecarProvider` trait because blob archive APIs aren't required to implement all functions contained in `BeaconClient` (as is the case for https://github.com/base-org/blob-archiver). Furthermore, `genesis_time` and `slot_interval` are taken from the primary provider which is already an online beacon client.

Would something like a `BlobArchiver` struct implementation be useful? 
I'm not sure if that should be left to the caller or to the library.

**Tests**

Added unit tests to make sure that blob fetching is completed even without any sidecar loaded in the primary.

**Additional context**

I need some pointers on error handling and metrics, as I'm not sure of the standard used across the codebase.

**Metadata**

- Fixes #316

### DIFF
--- a/crates/derive/src/online/blob_provider.rs
+++ b/crates/derive/src/online/blob_provider.rs
@@ -243,9 +243,16 @@ impl<B: BeaconClient, F: BlobSidecarProvider, S: SlotDerivation>
             return Ok(Vec::new());
         }
 
+        // Extract the genesis timestamp and slot interval from the primary provider.
+        let slot = S::slot(
+            self.primary.genesis_time.expect("Genesis Config Loaded"),
+            self.primary.slot_interval.expect("Config Spec Loaded"),
+            block_ref.timestamp,
+        )
+        .map_err(BlobProviderError::Slot)?;
+
         // Fetch blob sidecars for the given block reference and blob hashes.
-        let sidecars =
-            self.fallback.beacon_blob_side_cars(block_ref.timestamp, blob_hashes).await?;
+        let sidecars = self.fallback.beacon_blob_side_cars(slot, blob_hashes).await?;
 
         // Filter blob sidecars that match the indicies in the specified list.
         let blob_hash_indicies = blob_hashes.iter().map(|b| b.index).collect::<Vec<_>>();
--- a/crates/derive/src/online/blob_provider.rs
+++ b/crates/derive/src/online/blob_provider.rs
@@ -353,8 +353,8 @@ where
 /// The fallback provider is optional and can be set using the [Self::with_fallback] method.
 ///
 /// Two convenience methods are available for initializing the providers from beacon client URLs:
-/// - [Self::with_primary_beacon_client_url] for the primary beacon client.
-/// - [Self::with_fallback_beacon_client_url] for the fallback beacon client.
+/// - [Self::with_primary] for the primary beacon client.
+/// - [Self::with_fallback] for the fallback beacon client.
 #[derive(Debug, Clone)]
 pub struct OnlineBlobProviderBuilder<B: BeaconClient, F: BlobSidecarProvider, S: SlotDerivation> {
     beacon_client: Option<B>,
--- a/crates/derive/src/online/blob_provider.rs
+++ b/crates/derive/src/online/blob_provider.rs
@@ -589,7 +589,53 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_get_blobs_fallback() {
+    async fn test_get_blob_fallback() {
+        let json_bytes = include_bytes!("testdata/eth_v1_beacon_sidecars_goerli.json");
+        let sidecars: APIGetBlobSidecarsResponse = serde_json::from_slice(json_bytes).unwrap();
+
+        // Provide no sidecars to the primary provider to trigger a fallback fetch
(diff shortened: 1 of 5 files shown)


crates/derive/src/online/blob_provider.rs
```diff
@@ -176,6 +176,168 @@ where
...
+    async fn get_blobs(
+        &mut self,
+        block_ref: &BlockInfo,
+        blob_hashes: &[IndexedBlobHash],
+    ) -> Result<Vec<Blob>, BlobProviderError> {
+        crate::inc!(PROVIDER_CALLS, &["blob_provider", "get_blobs"]);
+        crate::timer!(START, PROVIDER_RESPONSE_TIME, &["blob_provider", "get_blobs"], timer);
+        match self.primary.get_blobs(block_ref, blob_hashes).await {
+            Ok(blobs) => Ok(blobs),
+            Err(_primary_err) => {
```
refcell, line 292: We should log the primary error here. If we're seeing much traffic routed to the fallback provider, we'd want to investigate the logs to see why the primary provider is throwing errors.

crates/derive/src/online/pipeline.rs
```diff
@@ -18,7 +18,7 @@ pub type OnlinePipeline =
 /// An `online` Ethereum data source.
 pub type OnlineDataProvider = EthereumDataSource<
     AlloyChainProvider,
-    OnlineBlobProvider<OnlineBeaconClient, SimpleSlotDerivation>,
+    OnlineBlobProviderWithFallback<OnlineBeaconClient, OnlineBeaconClient, SimpleSlotDerivation>,
```
refcell, line 21: I don't really mind doing this, but can we make the fallback optional so that the current functionality is backwards compatible?

examples/trusted-sync/src/main.rs
```diff
@@ -39,6 +39,7 @@ async fn sync(cli: cli::Cli) -> Result<()> {
     let l1_rpc_url = cli.l1_rpc_url()?;
     let l2_rpc_url = cli.l2_rpc_url()?;
     let beacon_url = cli.beacon_url()?;
+    let blob_archiver_url = cli.blob_archiver_url()?;
```
refcell, line 42: Even thought the argument is optional, this will cause the application to crash since the error is returned here.

We should introduce a builder pattern here IMO.
```
OnlineBlobProviderBuilder::new().with_beacon_client(beacon_client).with_fallback(Option<block_archiver_url>).build()?
```

This could return either the `OnlineBlobProvider` if no fallback is provided or `OnlineBlobProviderWithFallback` otherwise.

refcell (changes requested):
Request for backwards compatibility and a builder pattern

examples/trusted-sync/src/main.rs
merklefruit: Good point! Adding!

merklefruit:
@refcell Addressed the review. Just one point: `OnlineProviderBuilder::build()` now always returns `OnlineBlobProviderWithFallback` instead of two possible types depending if the fallback is defined. I can add that functionality with an extra typestate variable, however it would break the types for the convenience method `new_online_pipeline` used in the example.

In short, these are the two options:

1. (current implementation) always work with an `OnlineBlobProviderWithFallback`, which will act just like a regular `OnlineBlobProvider` in the case no fallback is added at build time. If a blob is not found, it will error with a clear message stating that blobs couldn't be retrieved and no fallback is configured.
2. break the types of `new_online_pipeline` and work with a generic `BlobProvider` trait in the example


I personally like (1) more, wdyt? Is there a better option? Open for feedback :)

examples/trusted-sync/src/main.rs
```diff
@@ -71,9 +72,14 @@ async fn sync(cli: cli::Cli) -> Result<()> {
...
-    let blob_provider =
-        OnlineBlobProvider::<_, SimpleSlotDerivation>::new(beacon_client, None, None);
+    let blob_provider = if let Some(blob_archiver_url) = blob_archiver_url {
+        OnlineBlobProviderBuilder::new()
+            .with_primary_beacon_client_url(beacon_url)
+            .with_fallback_beacon_client_url(blob_archiver_url)
+            .build()
+    } else {
+        OnlineBlobProviderBuilder::new().with_primary_beacon_client_url(beacon_url).build()
+    };
```
refcell, line 82: Can we just have `with_fallback_beacon_client_url` accept an Option here?

Then we can just do:
```rust
let blob_provider = OnlineBlobProviderBuilder::new().with_primary(beacon_url).with_fallback(cli.blob_archiver_url()).build()

```

refcell (approved):
Very nice!!

PR #411, opened 2024-08-01
fix(examples): Invalid Payload Retries With Backoff

refcell:
**Description**

Since validation can produce an invalid payload attributes to compare to the derived payload attributes (eg when calling `debug_getRawTransaction`, empty tx bytes are returned when they are in fact _not_ empty), we want to retry payload validation a few times before actually recording the **derived** payload attributes as invalid.

There have been multiple instances where this has thrown an alert, without the derived payload attributes being invalid, but instead the validation payload attributes were constructed incorrectly.

Also adds a `trusted_sync_validation_retries` metric that tracks the total number of retries.

### DIFF
--- a/examples/trusted-sync/src/cli.rs
+++ b/examples/trusted-sync/src/cli.rs
@@ -54,6 +54,13 @@ pub struct Cli {
         help = "Parameterized threshold amount of blocks to drift from before fast-forwarding or walking back"
     )]
     pub drift_threshold: u64,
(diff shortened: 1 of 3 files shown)


refcell:
* **#411** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/kona/411?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

clabby (approved):

PR #415, opened 2024-08-03
chore(examples): Add logs to trusted-sync

refcell:
**Description**

Adds logs to trusted-sync.

### DIFF
--- a/examples/trusted-sync/src/main.rs
+++ b/examples/trusted-sync/src/main.rs
@@ -278,7 +278,8 @@ async fn sync(cli: cli::Cli) -> Result<()> {
         let derived = attributes.parent.block_info.number as i64 + 1;
         metrics::SAFE_L2_HEAD.set(derived);
         metrics::DERIVED_ATTRIBUTES_COUNT.inc();
-        println!(
+        info!(
+            target: LOG_TARGET,
             "Validated Payload Attributes {} [L2 Block Num: {}] [L2 Timestamp: {}] [L1 Origin Block Num: {:?}]",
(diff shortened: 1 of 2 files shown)


refcell:
* **#415** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/kona/415?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

examples/trusted-sync/src/main.rs
```diff
@@ -277,12 +279,11 @@ async fn sync(cli: cli::Cli) -> Result<()> {
         metrics::SAFE_L2_HEAD.set(derived);
         metrics::DERIVED_ATTRIBUTES_COUNT.inc();
         println!(
```
clabby, line 281: `info!`?

refcell: We can, but since the point of `trusted-sync` is to show payloads are validated, I think it makes sense to always print this even without verbosity enabled. Or do you think I should just have the INFO as the default verbosity level?

clabby: `INFO`, `WARN`, `ERROR` should always be emitted imo

clabby: `DEBUG`/`TRACE` can be optional

refcell: Ok I'm fine with this, let's stick to tracing for io :)

clabby (approved):

PR #420, opened 2024-08-06
fix(derive): Granite Hardfork Support

refcell:
**Description**

> [!NOTE]
>
> Blocked by [superchain-registry#440](https://github.com/ethereum-optimism/superchain-registry/pull/440)

Updates the kona derivation pipeline to use the updated `superchain-primitives` and `superchain-registry` with the granite hardfork.

### DIFF
--- a/crates/derive/src/stages/channel_bank.rs
+++ b/crates/derive/src/stages/channel_bank.rs
@@ -88,7 +88,9 @@ where
         });
 
         // Check if the channel is not timed out. If it has, ignore the frame.
-        if current_channel.open_block_number() + self.cfg.channel_timeout < origin.number {
+        if current_channel.open_block_number() + self.cfg.channel_timeout(origin.timestamp) <
+            origin.number
+        {
             warn!(target: "channel-bank", "Channel {:?} timed out", frame.id);
             return Ok(());
         }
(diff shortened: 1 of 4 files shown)


refcell:
* **#420** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/kona/420?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

clabby (approved):
ty :)

crates/derive/src/stages/channel_bank.rs
```diff
@@ -88,7 +88,9 @@ where
         });
 
         // Check if the channel is not timed out. If it has, ignore the frame.
-        if current_channel.open_block_number() + self.cfg.channel_timeout < origin.number {
+        if current_channel.open_block_number() + self.cfg.channel_timeout(origin.timestamp) <
```
clabby, line 91: Oh wait actually - I don't think we should be using the origin timestamp here, yeah? We want an L2 timestamp for the L2 fork activation.

clabby (changes requested):

crates/derive/src/stages/channel_bank.rs
refcell: Oh shoot you're right, good catch

clabby: nvm https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/channel_bank.go#L101

clabby (approved):

PR #424, opened 2024-08-12
fix(client): Follow channel timeout consensus rule

### LINKED ISSUE #417: fix(client): Respect channel timeout
## Overview

Currently, we don't respect the channel timeout in `kona-client`. This change will be a painful one to make for performance, but we need to respect consensus rules rather than batcher policy.

In real-world application, the OP Stack batcher almost always opens and closes channels in a single transaction. However, this is strictly _not guaranteed_ by the protocol.

To accommodate for this, we must start derivation at least `RollupConfig::channel_timeout` blocks before the L1 origin of the trusted safe head.

https://github.com/ethereum-optimism/kona/blob/e069eb4178af1a2a7ea1f772f3f35c240d178534/bin/client/src/l1/driver.rs#L203-L203

clabby:
## Overview

Follows the consensus rule of walking back the channel timeout in sync-start. This should be followed up with an optimistic first try without walking back, and then if that fails, retrying derivation with the channel timeout walkback.

fixes #417

### DIFF
--- a/bin/client/src/l1/driver.rs
+++ b/bin/client/src/l1/driver.rs
@@ -199,8 +199,12 @@ where
         let safe_header = l2_chain_provider.header_by_hash(safe_hash)?;
         let safe_head_info = l2_chain_provider.l2_block_info_by_number(safe_header.number).await?;
 
-        let l1_origin =
-            chain_provider.block_info_by_number(safe_head_info.l1_origin.number).await?;
+        let l1_origin = chain_provider
+            .block_info_by_number(
+                safe_head_info.l1_origin.number +
+                    boot_info.rollup_config.channel_timeout(safe_head_info.block_info.timestamp),
+            )
+            .await?;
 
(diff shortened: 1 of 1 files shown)


bin/client/src/l1/driver.rs
```diff
@@ -199,8 +199,12 @@ where
         let safe_header = l2_chain_provider.header_by_hash(safe_hash)?;
         let safe_head_info = l2_chain_provider.l2_block_info_by_number(safe_header.number).await?;
 
-        let l1_origin =
-            chain_provider.block_info_by_number(safe_head_info.l1_origin.number).await?;
+        let l1_origin = chain_provider
+            .block_info_by_number(
+                safe_head_info.l1_origin.number +
+                    boot_info.rollup_config.channel_timeout(safe_head_info.block_info.timestamp),
```
refcell, line 205: Doesn't this need to be a subtraction?

clabby:
Closing this in favor of #427 - we'd like to have an optimistic pass for the p99 case, so that we don't need to accept the performance hit.

PR #427, opened 2024-08-13
feat(client): Walkback Channel Timeout in Sad Path

### LINKED ISSUE #417: fix(client): Respect channel timeout

refcell:
**Description**

An alternative to #424 that walks back the `channel_timeout` in the sad path when the pipeline fails to decode a batch from ready channel data.

**Metadata**

Fixes #417

### DIFF
--- a/bin/client/src/l1/driver.rs
+++ b/bin/client/src/l1/driver.rs
@@ -67,6 +67,10 @@ where
     l2_safe_head_header: Sealed<Header>,
     /// The inner pipeline.
     pipeline: OraclePipeline<O, B>,
+    /// The channel timeout to walk back the pipeline origin.
+    walkback: Option<u64>,
+    /// Handle to the l1 chain provider used to fetch the new pipeline origin.
+    chain_provider: OracleL1ChainProvider<O>,
 }
 
 impl<O, B> DerivationDriver<O, B>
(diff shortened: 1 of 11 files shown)


refcell:
* **#427** <a href="https://app.graphite.dev/github/pr/ethereum-optimism/kona/427?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

bin/client/src/l1/driver.rs
```diff
@@ -158,6 +165,15 @@ where
                 StepResult::StepFailed(e) => {
                     warn!(target: "client_derivation_driver", "Failed to step derivation pipeline: {:?}", e)
                 }
+                StepResult::FailedToDecodeBatch => {
+                    if self.walkback.is_some() {
+                        let origin =
+                            self.l2_safe_head.block_info.timestamp - self.walkback.unwrap();
```
clabby, line 171: style nit - Rather than `is_some` + `unwrap`, just `if let Some(_)`?

crates/derive/src/stages/channel_reader.rs
```diff
@@ -96,18 +96,18 @@ where
...
-            .as_mut()
-            .expect("Cannot be None")
-            .next_batch(self.cfg.as_ref())
-            .ok_or(StageError::NotEnoughData)
-        {
+        match self.next_batch.as_mut().expect("Cannot be None").next_batch(self.cfg.as_ref()) {
             Ok(batch) => Ok(batch),
-            Err(e) => {
+            Err(BatchReaderError::FailedToDecodeBatch) => {
+                // TODO: here we want to bubble up a stage error
```
clabby, line 102: Is this still a TODO?

clabby (commented):
This looks good overall. Curious about that `TODO`.

refcell:
This does not prevent a malicious batcher from posting multiple valid frames within the channel window where the first frame is missed because it is posted prior to the l1 origin....

refcell:
Closing for now. Holocene should fix and if we absolutely need, can always re-open.

PR #446, opened 2024-08-22
feat: serde Serialization/Deserialization to BootInfo struct

zobront:
**Description**

Since we're passing the rollup config, we'll be able to use the BootInfo struct directly throughout our program (instead of the custom struct we'd previously defined, which didn't include the rollup config).

This requires serialization / deserialization.

Note that I'd usually put this behind a feature flag, but the `client` program already imports serde with the `derive` feature active, so didn't think it was necessary. Let me know if you'd prefer that I do.

**Tests**

N/A

**Additional context**

N/A

**Metadata**

N/A


refcell (commented):
This LGTM - once ci is passing gtg

refcell:
Replaced by #448

PR #461, opened 2024-08-27
chore(bin): Remove `kt`

clabby:
## Overview

Removes the `kt` binary from `kona`. We are pivoting a bit to a different test format. This tool may return in the future, but for now, is just adding extra compile time etc.


refcell (approved):

refcell:
RIP `ethereum-optimism/tests` you were magnificent while you lasted. A true steward of testing. A lone vector in the night.

PR #481, opened 2024-09-04
fix: fix compile error about derive crate with online feature

raindust:
Fix https://github.com/anton-rs/kona/issues/480


refcell:
This should be fixed by https://github.com/anton-rs/kona/pull/482

PR #484, opened 2024-09-04
fix: Deprecate All Flag

refcell:
### Description

Remove the `--all` flag that is deprecated in favor of `--workspace`.


clabby (approved):
TIL

Screenshot 2024-09-04 at 1 55 43 PM

PR #491, opened 2024-09-05
fix(OrderedListWalker): Support empty tries

hashcashier:
`OrderedListWalker::fetch_leaves` returns an error when traversing an empty trie, since the root node is of type `TrieNode::Empty`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,6 @@ target/
 
 # Environment Variables
 .env
+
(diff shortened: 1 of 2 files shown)


clabby:
Thanks! It looks like our CI is currently rejecting external contributions due to the `PAT_TOKEN` - re-opened in #493 and fixed CI in #494 for the future.

PR #492, opened 2024-09-05
fix(primitives): op-alloy Frame Type

### LINKED ISSUE #487: fix(primitives): Remove Frame
### Description

Remove the [`frame`](https://github.com/anton-rs/kona/blob/main/crates/primitives/src/frame.rs) module in `kona-primitives` and re-export the [`Frame`](https://github.com/alloy-rs/op-alloy/blob/main/crates/protocol/src/frame.rs#L60) type from `op-alloy-protocol`.

A future ticket will remove all of `kona-primitives` and drop `op-alloy-protocol` in-place.

refcell:
### Description

Removes the `frame` module from `kona-primitives` in favor of using the one from upstream [`op-alloy-protocol`](https://github.com/alloy-rs/op-alloy/tree/main/crates/protocol).

Will put up an issue to remove the `patch.crates-io` once the `op-alloy-protocol` pr lands and the version is bumped.

Closes #487


refcell:
> Looks like there's other types (`Channel`, `BlockInfo`, etc.) that we can migrate as well - can we do that now? [alloy-rs/op-alloy@`main`/crates/protocol/src](https://github.com/alloy-rs/op-alloy/tree/main/crates/protocol/src?rgh-link-date=2024-09-05T13%3A38%3A37Z)

Working on this in follow-on prs atm. They're more complicated.

clabby (approved):

PR #506, opened 2024-09-06
feat(driver): Don't panic by default on disputed payload derivation failure

hashcashier:
This PR enables the client to customize how it handles a failure by `DerivationDriver::produce_disputed_payload` due to `StageError::NotEnoughData`. This would enable a `no_std` client to handle that an output cannot be derived from the available L1 data without panic handling.

### DIFF
--- a/bin/client/src/kona.rs
+++ b/bin/client/src/kona.rs
@@ -58,10 +58,8 @@ fn main() -> Result<()> {
             l2_provider.clone(),
         )
         .await?;
-        let L2AttributesWithParent { attributes, .. } = driver
-            .produce_disputed_payload()
-            .await?
-            .expect("Failed to derive payload attributes");
+        let L2AttributesWithParent { attributes, .. } =
+            driver.produce_disputed_payload().await?.expect("Failed to derive payload attributes");
 
         let mut executor = StatelessL2BlockExecutor::builder(&boot.rollup_config)
             .with_parent_header(driver.take_l2_safe_head_header())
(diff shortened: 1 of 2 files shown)


bin/client/src/l1/driver.rs
```diff
@@ -151,7 +151,7 @@ where
 
     /// Produces the disputed [L2AttributesWithParent] payload, directly after the starting L2
     /// output root passed through the [BootInfo].
-    pub async fn produce_disputed_payload(&mut self) -> Result<L2AttributesWithParent> {
+    pub async fn produce_disputed_payload(&mut self) -> Result<Option<L2AttributesWithParent>> {
```
refcell, line 154: I'd prefer to flatten this error type then.

e.g.
```rust
	pub async fn produce_disputed_payload(&mut self) -> Result<L2AttributesWithParent> {
		...

		attributes.ok_or("Failed to derive payload attributes")
	}
```

bin/client/src/kona.rs
```diff
@@ -58,7 +58,8 @@ fn main() -> Result<()> {
             l2_provider.clone(),
         )
         .await?;
-        let L2AttributesWithParent { attributes, .. } = driver.produce_disputed_payload().await?;
+        let L2AttributesWithParent { attributes, .. } =
+            driver.produce_disputed_payload().await?.expect("Failed to derive payload attributes");
```
refcell, line 62: If we keep the signature `fn() -> Result<L2AttributesWithParent>` as I suggest, we shouldn't need to change this since the error will bubble up :)

refcell (changes requested):
Small change

refcell:
This is fixed by #545 - closing for now. Please open an issue if you dislike the solution from #545. Thanks!

PR #535, opened 2024-09-18
fix(preimage): Improve error differentiation in preimage servers

clabby:
## Overview

Improves the differentiation of errors in the preimage server routes. This allows for us to throw on non-pipe related errors, rather than silently fail and assume that the exit is related to a broken pipe.

Also fixes an edge case where if the `HintReaderServer` receives bad UTF-8, it will block forever.

### DIFF
--- a/bin/host/src/server.rs
+++ b/bin/host/src/server.rs
@@ -8,9 +8,12 @@ use crate::{
     },
 };
 use anyhow::{anyhow, Result};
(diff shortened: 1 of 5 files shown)


clabby:
* **#535** <a href="https://app.graphite.dev/github/pr/anton-rs/kona/535?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

refcell (approved):

PR #541, opened 2024-09-20
feat(mpt): Migrate to `thiserror`

clabby:
## Overview

Migrates `kona-mpt` to typed errors.

### DIFF
--- a/crates/mpt/src/errors.rs
+++ b/crates/mpt/src/errors.rs
@@ -1,0 +1,64 @@
+//! Errors for the `kona-derive` crate.
+
+use alloc::string::String;
+use thiserror::Error;
+
+/// A [Result] type alias where the error is [TrieNodeError].
+pub type TrieNodeResult<T> = Result<T, TrieNodeError>;
+
+/// An error type for [TrieNode] operations.
+///
+/// [TrieNode]: crate::TrieNode
+#[derive(Error, Debug, PartialEq, Eq)]
+pub enum TrieNodeError {
+    /// Invalid trie node type encountered.
+    #[error("Invalid trie node type encountered")]
+    InvalidNodeType,
+    /// Failed to decode trie node.
+    #[error("Failed to decode trie node: {0}")]
+    RLPError(alloy_rlp::Error),
+    /// Key does not exist in trie.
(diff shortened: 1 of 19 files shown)


crates/mpt/src/errors.rs
```diff
@@ -0,0 +1,64 @@
...
+    /// Key does not exist in trie.
+    #[error("Key does not exist in trie. Encountered {0} node.")]
+    KeyNotFound(String),
+    /// Trie node is not a leaf node.
+    #[error("Trie provider error: {0}")]
+    Provider(String),
+}
+
+/// A [Result] type alias where the error is [TrieDBError].
+pub type TrieDBResult<T> = Result<T, TrieDBError>;
```
refcell, line 29: It seems off to me that we've removed the `DB` abbreviation in other types like `TrieProvider` and `TrieHinter` but use it here. Would it not make sense to call this `TrieResult` as well? Would favor consistency if possible, otherwise it's difficult for grugbrain.dev to understand.

refcell, line 29: It seems off to me that we've removed the `DB` abbreviation in other types like `TrieProvider` and `TrieHinter` but use it here. Would it not make sense to call this `TrieResult` as well? Would favor consistency if possible, otherwise it's difficult for grugbrain.dev to understand.

crates/mpt/src/fetcher.rs
```diff
@@ -42,20 +46,23 @@ pub trait TrieDBFetcher {
...
+    fn header_by_hash(&self, hash: B256) -> Result<Header, Self::Error>;
 }
 
-/// The [TrieDBHinter] trait defines the synchronous interface for hinting the host to fetch trie
+/// The [TrieHinter] trait defines the synchronous interface for hinting the host to fetch trie
 /// node preimages.
-pub trait TrieDBHinter {
+pub trait TrieHinter {
+    /// The error type for hinting trie node preimages.
+    type Error: Display + ToString;
```
refcell, line 56: Do we actually need the error type here to be `Display + ToString`?

We could use the unit type for the `Noop` impls if not.

refcell (changes requested):

refcell (approved):

Issue #559, opened 2024-09-23
[HOLOCENE] feat(derive): BatchStream Stage


refcell:
### Description

> [!WARNING]
>
> It is highly encouraged to [read the specs](https://specs.optimism.io/protocol/holocene/overview.html) prior to working on this implementation.

As part of the Holocene hardfork, span batch validity is changed to use a `forward-invalidation` model rather than the pre-holocene `backwards-invalidation`. Instead of allowing a span batch to be partially derived and then be invalidated by a single batch after those batches are derived, the entire span batch is effectively buffered in a new stage in the derivation pipeline. If an invalid batch is found as part of the span batch, the span batch is dropped.

The rules for span batch validation are now as follows.

- If the span batch L1 origin check is not part of the canonical L1 chain, the span batch is invalid.
- A failed parent check invalidates the span batch.
- If span_start.timestamp > next_timestamp, the span batch is invalid, because we disallow gaps due to the new strict batch ordering rules.
- If span_end.timestamp < next_timestamp, the span batch is invalid, as it doesn't contain any new batches (this would also happen if applying timestamp checks to each derived singular batch individually).
- Note that we still allow span batches to overlap with the safe chain (span_start.timestamp < next_timestamp).

### Output

- [x] Introduce a new stage in the derivation pipeline called the `BatchStream`.
  - This stage sits before the `BatchQueue`.
  - It is only activated post-holocene and should "pass-through" single batches prior to holocene, effectively acting as a noop.
- [x] Implement the `BatchQueueProvider` trait on the `BatchStream` similar to how the [`ChannelReader` implements the trait](https://github.com/anton-rs/kona/blob/main/crates/derive/src/stages/channel_reader.rs#L94).
  - Notice, now both the `ChannelReaderStage` and `BatchStream` stage implement the `BatchQueueProvider` traits. This is intentional so the `BatchStream` can "slot in" between the `ChannelReader` and `BatchQueue`. Internally, the `BatchStream` should hold a `prev: P where P: BatchQueueProvider` so it both accepts and implements the `BatchQueueProvider`.
- [x] Implement span batch validation rules when holocene is active.

refcell:
This is complete in #592

PR #566, opened 2024-09-24
feat(derive): New BatchStream Stage for Holocene

refcell:
### Description

Adds a new `BatchStream` stage for holocene.

Makes progress towards #559

### DIFF
--- a/crates/derive/src/stages/batch_stream.rs
+++ b/crates/derive/src/stages/batch_stream.rs
@@ -50,12 +50,9 @@ where
 
     /// Returns if the [BatchStream] stage is active based on the
     /// origin timestamp and holocene activation timestamp.
-    pub fn is_active(&self) -> bool {
-        // If the origin cannot be fetched, cannot be active.
-        let Some(origin) = self.prev.origin() else {
-            return false;
-        };
-        self.config.is_holocene_active(origin.timestamp)
+    pub fn is_active(&self) -> PipelineResult<bool> {
+        let origin = self.prev.origin().ok_or(PipelineError::MissingOrigin.crit())?;
+        Ok(self.config.is_holocene_active(origin.timestamp))
     }
 
     /// Gets a [SingleBatch] from the in-memory buffer.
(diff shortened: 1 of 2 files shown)


clabby (commented):
This is a good start. Small preference for `BatchStream` rather than `BatchSpan`, as a dumb naming nit.

I think there's a few things we need to think about architecturally here, though. The way I think about this is that we need to be able to stream single batches from the span batch into the `BatchQueue`, one at a time. This means that, from holocene forward, every batch sent into the `BatchQueue` is a `SingleBatch`. We need to retain `next_batch(...) -> PipelineResult<Batch>`, to keep that backwards compatibility, but the new stage should enforce that.

What this looks like in my head is something like:
Untitled-2022-11-03-2344

The things I'm trying to optimize for here is:
* The `BatchQueue` currently performs the `check_batch` call. But it does it on `Batch`, which wraps both `SingleBatch` + `SpanBatch`. After holocene, basically, the `BatchQueue` should never receive a `SpanBatch`. This way, we can do the [span batch prefix check](https://specs.optimism.io/protocol/holocene/derivation.html#span-batches) in the `BatchStream`, and just reuse the `SingleBatch` check in the BQ.
* The `BatchStream` stage basically acts as a buffer of `SingleBatch`es, derived from a `SpanBatch`. So it holds all of the `SingleBatch`es from a span batch in-memory, and gets drained by the `BatchQueue` as `SingleBatch`es are read. It can eventually land on `Eof`, signaling that it needs to fetch a new batch from the `ChannelReader`.

Notably this means that the stage will always be present - no need for a `active` flag, just forwards batches from the `ChannelReader` directly pre-holocene. Otherwise, it acts as an in-memory buffer of `SingleBatch`es, and also owns the job of validating the span batch's prefix per the spec.

crates/derive/src/stages/batch_stream.rs
```diff
@@ -0,0 +1,169 @@
...
+    /// Create a new [BatchStream] stage.
+    pub fn new(prev: P, config: Arc<RollupConfig>) -> Self {
+        Self { prev, span: None, buffer: Vec::new(), config }
+    }
+
+    /// Returns if the [BatchStream] stage is active based on the
+    /// origin timestamp and holocene activation timestamp.
+    pub fn is_active(&self) -> bool {
+        // If the origin cannot be fetched, cannot be active.
+        let Some(origin) = self.prev.origin() else {
```
clabby, line 55: This should be a critical error, if the origin is missing

clabby (approved):
Good start! Let's get this in and keep iterating.

PR #572, opened 2024-09-25
feat(derive): Holocene Channel Bank Checks

### LINKED ISSUE #558: [HOLOCENE] feat(derive): ChannelBank Strict Ordering
### Description

> [!WARNING]
>

As part of [Strict Batch Ordering][specs], the [channel-bank] is greatly simplified since there can only ever be a single `Channel` buffered at a given time.

[specs]: https://specs.optimism.io/protocol/holocene/derivation.html#strict-frame-and-batch-ordering
[channel-bank]: https://specs.optimism.io/protocol/holocene/derivation.html#channel-bank

Since the `ChannelBank` needs to maintain backwards-compatibility, the in-memory channel buffer cannot be simplified to only hold a single `Channel`. Instead, we can only add a check inside the [`ChannelBank::ingest_frame`](https://github.com/anton-rs/kona/blob/main/crates/derive/src/stages/channel_bank.rs#L82) method that validates there are no `Channel`s in the `channel_queue` before creating a new one.

### Output

Within [`ChannelBank::ingest_frame`](https://github.com/anton-rs/kona/blob/main/crates/derive/src/stages/channel_bank.rs#L85-L90) add a check that the `ChannelBank`'s `channel_queue` is empty before instantiating a new `Channel`.

Use [`RollupConfig::is_holocene_active`](https://github.com/alloy-rs/op-alloy/blob/main/crates/genesis/src/rollup.rs#L207-L210) to check if holocene is active when performing this check.

An example could look like the following.

```rust
// -- snip --

// Get the channel for the frame, or create a new one if it doesn't exist.
let current_channel = match self.channels.entry(frame.id){
	Some(c) => c,
	None => {
		if self.cfg.is_holocene_active(timestamp) && !self.channel_queue.is_empty() {
			// In holocene, channels are strictly ordered.
			// If the previous frame is not the last in the channel
			// and a starting frame for the next channel arrives,
			// the previous channel/frames are removed and a new channel is created.
			self.channel_queue.clear();

			// TODO: metrice
		}
		let channel = Channel::new(frame.id, origin);
    	self.channel_queue.push_back(frame.id);
    	channel
	}
};

// -- snip --
```

Notice, add metrics in the above if an out-of-order channel made it into the `ChannelBank` without being filtered out in the `FrameQueue`. Also add an `error!` log with `tracing` since this is a critical error in the implementation of Holocene changes in the `FrameQueue`.

refcell:
### Description

Adds checks to the `ChannelBank` for post-holocene strict ordering that only one channel may only be buffered at a given point in time within the channel queue.

Closes #558

### DIFF
--- a/crates/derive/src/stages/channel_bank.rs
+++ b/crates/derive/src/stages/channel_bank.rs
@@ -41,13 +41,13 @@ where
     P: ChannelBankProvider + OriginAdvancer + OriginProvider + ResettableStage + Debug,
 {
     /// The rollup configuration.
-    cfg: Arc<RollupConfig>,
+    pub cfg: Arc<RollupConfig>,
     /// Map of channels by ID.
-    channels: HashMap<ChannelId, Channel>,
+    pub channels: HashMap<ChannelId, Channel>,
     /// Channels in FIFO order.
-    channel_queue: VecDeque<ChannelId>,
+    pub channel_queue: VecDeque<ChannelId>,
     /// The previous stage of the derivation pipeline.
-    prev: P,
+    pub prev: P,
 }
 
 impl<P> ChannelBank<P>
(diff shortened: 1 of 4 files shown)


refcell:
* **#572** <a href="https://app.graphite.dev/github/pr/anton-rs/kona/572?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/derive/src/stages/channel_bank.rs
clabby, line 1: Curious your take on the `MAX_CHANNEL_BANK_SIZE` change. Should we be invalidating that on holocene activation, and explicitly enforcing that a single channel exists within the bank?

clabby (approved):
This LGTM. It'd be nice to add a quick unit test that covers the case where the channel queue is cleared, when a new frame is sent during an existing channel's buffer period.

crates/derive/src/stages/channel_bank.rs
refcell: With this PR we are strictly enforcing only a single channel ever exists within the bank.

The specs _do not_ say to remove the channel bank size, but rather it isn't effective since it is greater than the channel size.

```
The MAX_CHANNEL_BANK_SIZE parameter becomes ineffective and the size of a channel is bounded, as before, during decompression to be at most of size MAX_RLP_BYTES_PER_CHANNEL.
```

Given this, I would follow the specs by not removing / de-activating this logic post-holocene.

Issue #553, opened 2024-09-23
feat(common): Improve native IO API


clabby:
## Overview

Right now, the IO API for the native platform in `kona-common` is inherently unsafe. Though, because it is behind the `BasicKernelInterface` trait, the API is not actually marked as `unsafe` lexically.

The reason the API is currently unsafe is because of our use of raw file descriptors. The `FileDescriptor` enum for the FPVM target is constant, and it is also `Clone + Copy`. However, on the native platform, file descriptors cannot be safely cloned or copied without checking for potential issues with creating a second handle to the fd.

This has resulted in a hacky workaround, where we hold onto the raw file descriptor, expect the host process to keep them alive for long enough, and [construct + forget](https://github.com/anton-rs/kona/blob/main/crates/common/src/io.rs#L78-L85) the file for read and write ops. While this works, the API is very unclear, and this invisible requirement of having the host keep the file descriptors alive is a bit annoying.

Instead, we should treat the `FileDescriptor` enum as an _owned_ type. Cloning or copying it should be a falliable operation, i.e. [OwnedFd::try_clone](https://doc.rust-lang.org/stable/std/os/fd/struct.OwnedFd.html#method.try_clone).

This will help improve the API to expose a truly safe native IO API, and prevent us from having to leak the `File`s in `NativeIO`.

malik672:
@clabby this should work if a try_clone() is implemented right ?

clabby:
This is fixed with in-process execution of the client program in the host - we use native channels now, so the native IO implementation doesn't leak into the FPVM IO abstractions. This logic has been moved into the platform-specific `kona-std-fpvm` crate.

Issue #398, opened 2024-07-24
fix(common): remove ZkvmIO


zobront:
**Is your feature request related to a problem? Please describe.**
To allow for compilation of `kona-common` in SP1, it is required to stop NativeIO from compiling. The current workaround is to implement a `ZkvmIO` alternative which is triggered when `target_os = "zkvm"`, which contains only no-ops.

@clabby would like to make sure this workaround is removed, so adding this ticket to ensure that gets done.

**Describe the solution you'd like**
TBD

**Describe alternatives you've considered**
The obvious answer seems to be to use feature flags. I had a WIP version that used a default `io` feature flag, which could be turned off when importing into a zkvm environment. However, because `kona` is set up as a workspace, it was difficult to get the `io` feature to turn off across the board without adding a feature to many of the crates with lots of messy interdependencies.

**Additional context**
N/A

PR #573, opened 2024-09-25
feat: Switch to `std::process::Command` when invoking the program binary

ratankaliani:
The use of [command-fds](https://crates.io/crates/command-fds) with tokio::process::Command causes rare, deterministic race conditions that lead to witness generation failures. These issues consistently occur for specific block ranges on OP Sepolia, Base, and OP Mainnet in [op-succinct](https://github.com/succinctlabs/op-succinct).

We should switch to std::process::Command to avoid these race conditions. This will increase resource usage for concurrent witness generation, and we can move back to `tokio` once the race condition issues are resolved.

I'm not exactly certain the source of the issue, but the Safety section of `tokio::process::Command` mentions some issues arising from file descriptors: https://tikv.github.io/doc/tokio/process/struct.Command.html#safety

### DIFF
--- a/bin/host/src/lib.rs
+++ b/bin/host/src/lib.rs
@@ -23,9 +23,10 @@ use std::{
     io::{stderr, stdin, stdout},
     os::fd::{AsFd, AsRawFd},
     panic::AssertUnwindSafe,
+    process::Command,
     sync::Arc,
 };
-use tokio::{process::Command, sync::RwLock, task};
+use tokio::{sync::RwLock, task};
 use tracing::{debug, error, info};
 use util::Pipe;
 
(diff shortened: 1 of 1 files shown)


clabby:
It's a bit concerning that we have to move to a blocking task to fix this. I'd prefer if we found out the root cause of the issue before making it such that the future that spawns this command cannot be cancelled. That can cause hanging behavior in ways we may not anticipate, if the host program panics.

I think a better approach here, at least for the native run, is to move towards just executing the client program from within the host. It'll require a bit of a larger refactor (likely including some or all of #398 + #361 + #553), but would remove the child process.

PR #565, opened 2024-09-24
feat(derive): Holocene Reset Type

refcell:
### Description

Refactors stage resets into a type to easily slot in a new Holocene hardfork reset type.

Closes #556

PR #574, opened 2024-09-25
feat(derive): Holocene Activation

### LINKED ISSUE #556: [TRACKING] Holocene Hardfork Changes
### Description

> [!WARNING]
>

This is a tracking issue for [Holocene Hardfork](https://specs.optimism.io/protocol/holocene/overview.html) changes to kona and its dependency set.

### LINKED ISSUE #562: [HOLOCENE] feat(derive): Pipeline Reset on HF Activation
### Description

> [!WARNING]
>

When the L1 origin of the derivation pipeline is advanced to the timestamp at which the Holocene hardfork first activates, the [derivation pipeline needs to be reset](https://specs.optimism.io/protocol/holocene/derivation.html#activation) so any lingering frames, channels, and batches are removed.

Without inline code switches, alternate stage implementations need to be used instead as detailed in the specs. Since the Holocene hardfork in `kona` is implemented with inline toggles using the rollup config, we don't have alternate stage implementations so this note is mute.

### Output

This ticket will entail a larger derivation pipeline refactor.

Currently, the derivation pipeline implements the [`Pipeline::reset`](https://github.com/anton-rs/kona/blob/main/crates/derive/src/traits/pipeline.rs#L30-L31) method across all stages.

Since holocene introduces a new way of resetting the pipeline, we like to avoid adding custom methods across the entire array of stages for this hardfork and going forward.

This ticket is to refactor the `reset` method to be a more generic way to pass a message down the pipeline.

- [ ] Introduce a `ResetType` enum that can be either `FULL` or `HOLOCENE` for now.
- [ ] Update the [`Pipeline::reset`](https://github.com/anton-rs/kona/blob/main/crates/derive/src/traits/pipeline.rs#L30-L31) method to accept the `ResetType`.
- [ ] Update each stage reset method to perform the current reset if the type is `FULL` and the holocene reset for `HOLOCENE`.
- [ ] Update the derivation pipeline drivers in `client` and `trusted-sync` to reset the pipeline when the holocene L1 timestamp is reached.

refcell:
### Description

Introduces logic for holocene activation.

Replaces #565

Closes #562

### DIFF
--- a/crates/derive/src/errors.rs
+++ b/crates/derive/src/errors.rs
@@ -104,7 +104,7 @@ impl PipelineError {
 }
 
 /// A reset error
-#[derive(Error, Debug, Eq, PartialEq)]
+#[derive(Error, Clone, Debug, Eq, PartialEq)]
 pub enum ResetError {
     /// The batch has a bad parent hash.
     /// The first argument is the expected parent hash, and the second argument is the actual
@@ -127,6 +127,16 @@ pub enum ResetError {
     /// Attributes builder error variant, with [BuilderError].
     #[error("Attributes builder error: {0}")]
     AttributesBuilder(#[from] BuilderError),
+    /// A Holocene activation temporary error.
+    #[error("Holocene activation reset")]
+    HoloceneActivation,
+}
+
+impl ResetError {
+    /// Wrap [self] as a [PipelineErrorKind::Reset].
+    pub const fn reset(self) -> PipelineErrorKind {
+        PipelineErrorKind::Reset(self)
+    }
 }
 
 /// A decoding error.
@@ -159,7 +169,7 @@ pub enum BatchDecompressionError {
 /// An [AttributesBuilder] Error.
 ///
 /// [AttributesBuilder]: crate::traits::AttributesBuilder
-#[derive(Error, Debug, PartialEq, Eq)]
+#[derive(Error, Clone, Debug, PartialEq, Eq)]
 pub enum BuilderError {
     /// Mismatched blocks.
     #[error("Block mismatch. Expected {0:?}, got {1:?}")]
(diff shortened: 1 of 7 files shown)


crates/derive/src/errors.rs
```diff
@@ -127,6 +127,16 @@ pub enum ResetError {
     /// Attributes builder error variant, with [BuilderError].
     #[error("Attributes builder error: {0}")]
     AttributesBuilder(#[from] BuilderError),
+    /// A Holocene activation temporary error.
+    #[error("Holocene activation error")]
```
clabby, line 131: smol nit
```suggestion
    #[error("Holocene activation reset")]
```

crates/derive/src/stages/l1_traversal.rs
```diff
@@ -108,8 +108,19 @@ impl<F: ChainProvider + Send> OriginAdvancer for L1Traversal<F> {
...
+
         self.block = Some(next_l1_origin);
         self.done = false;
+
+        // If the prev block is not holocene, but the next is, we need to flag this
+        // so the pipeline driver will reset the pipeline for holocene activation.
+        let prev_block_holocene =
+            self.rollup_config.is_holocene_active(self.block.unwrap().timestamp);
+        let next_block_holocene = self.rollup_config.is_holocene_active(next_l1_origin.timestamp);
+        if !prev_block_holocene && next_block_holocene {
```
clabby, line 120: There's a bug here - we're setting `self.block` to `next_l1_origin`, then comparing it with itself.

clabby (changes requested):

crates/derive/src/stages/l1_traversal.rs
```diff
@@ -108,8 +108,21 @@ impl<F: ChainProvider + Send> OriginAdvancer for L1Traversal<F> {
         }
 
         crate::set!(ORIGIN_GAUGE, next_l1_origin.number as i64);
+
+        let prev_block_holocene =
+            self.rollup_config.is_holocene_active(self.block.unwrap().timestamp);
```
clabby, line 113: This is safe, but can we pls add a comment for why we can `unwrap` here (or just use `expect` w/ a message that it's infallible due to the first check.)

clabby (approved):

PR #575, opened 2024-09-26
feat(derive): Holocene Buffer Flushing

refcell:
### Description

Small PR to add buffer flushing to the batch stream.

### DIFF
--- a/crates/derive/src/stages/batch_queue.rs
+++ b/crates/derive/src/stages/batch_queue.rs
@@ -23,6 +23,10 @@ pub trait BatchQueueProvider {
     ///
     /// [ChannelReader]: crate::stages::ChannelReader
     async fn next_batch(&mut self) -> PipelineResult<Batch>;
+
+    /// Allows the [BatchQueue] to flush the buffer in the [crate::stages::BatchStream]
+    /// if an invalid single batch is found. Pre-holocene hardfork, this will be a no-op.
+    fn flush(&mut self);
 }
 
 /// [BatchQueue] is responsible for o rdering unordered batches
@@ -146,6 +150,9 @@ where
                     remaining.push(batch.clone());
                 }
                 BatchValidity::Drop => {
+                    // If we drop a batch, flush previous batches buffered in the BatchStream
+                    // stage.
+                    self.prev.flush();
                     warn!(target: "batch-queue", "Dropping batch with parent: {}", parent.block_info);
                     continue;
                 }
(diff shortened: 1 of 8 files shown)


refcell:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/anton-rs/kona/575?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>
> <a href="https://graphite.dev/docs/merge-pull-requests">Learn more</a>

* **#575** <a href="https://app.graphite.dev/github/pr/anton-rs/kona/575?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#567** <a href="https://app.graphite.dev/github/pr/anton-rs/kona/567?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#566** <a href="https://app.graphite.dev/github/pr/anton-rs/kona/566?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

crates/derive/src/stages/batch_queue.rs
```diff
@@ -146,6 +150,9 @@ where
                     remaining.push(batch.clone());
                 }
                 BatchValidity::Drop => {
+                    // If we drop a batch, flush previous batches buffered in the BatchStream
+                    // stage.
+                    self.prev.flush();
```
clabby, line 155: I _think_ this is right, but would be good to triple check. The stage should only be flushed on partial span batch invalidity, I wonder if there are other batch drop cases where this shouldn't be hit?

crates/derive/src/stages/channel_reader.rs
```diff
@@ -99,6 +99,9 @@ impl<P> BatchQueueProvider for ChannelReader<P>
 where
     P: ChannelReaderProvider + OriginAdvancer + OriginProvider + ResettableStage + Send + Debug,
 {
+    fn flush(&mut self) { /* noop */
```
clabby, line 102: Wonder if we should make a new set of traits here, since we plan on always including the `BatchStream`? This doesn't need to be here, but I'm cool w/ keeping the no-op.

clabby (approved):

crates/derive/src/stages/batch_queue.rs
refcell: My thinking:
- Pre-holocene this flush will be a no-op.
- Post-holocene, single batches are either actually single, or derived from a span batch.
  - If they are single, there shouldn't be a span batch in the batch stream due to strict ordering.
  - If they are derived from a span batch the whole span batch and it's channel need to be flushed.

PR #579, opened 2024-09-27
feat(derive): Holocene Frame Queue

### LINKED ISSUE #557: [HOLOCENE] feat(derive): FrameQueue Strict Ordering
### Description

> [!WARNING]
>

As part of [Strict Batch Ordering][specs], the [frame-queue] must be modified with the following constraints:

[specs]: https://specs.optimism.io/protocol/holocene/derivation.html#strict-frame-and-batch-ordering
[frame-queue]: https://specs.optimism.io/protocol/holocene/derivation.html#frame-queue

- If a non-first frame (i.e., a frame with index >0) decoded from a batcher transaction is out of order, it is immediately dropped, where the frame is called out of order if
  - its frame number is not the previous frame's plus one, if it has the same channel ID, or
  - the previous frame already closed the channel with the same ID, or
  - the non-first frame has a different channel ID than the previous frame in the frame queue.
- If a first frame is decoded while the previous frame isn't a last frame (i.e., is_last is false), all previous frames for the same channel are dropped and this new first frame remains in the queue.

These changes must be added to the `FrameQueue` stage in `kona-derive`, with backwards-compatibility.

### Output

Add out-of-order changes to the [FrameQueue](https://github.com/anton-rs/kona/blob/main/crates/derive/src/stages/frame_queue.rs#L76-L77) after parsing frames. This should be placed within a branch that is hit if holocene is active. Use [`RollupConfig::is_holocene_active`](https://github.com/alloy-rs/op-alloy/blob/main/crates/genesis/src/rollup.rs#L207-L210).

Out of order validation rules:
- Since the Frame queue is empty, the start of the frame array can trimmed while the frame number is non-zero, effectively dropping the frames. Add a metric to record the number of dropped frames due to out of order frame number.
- Validate the channel ids are strictly ordered, dropping and metricing out of order frames.
- Validate frames are not present with already-closed channels.

Add a validation rule for when the first frame is decoded while the previous channel is open, drop all previous frames and start with the current channel.

For reference, the `Frame` type lives [in op-alloy](https://github.com/alloy-rs/op-alloy/blob/main/crates/protocol/src/frame.rs#L88).

Alternatively, we can place the validation rules within the `Frame` type `impl` in `op-alloy` such that in the `FrameQueue` we could write the following, where `Frames` is a wrapper type around `Vec<Frame>`.

```rust
// -- snip --
                    
if let Ok(frames) = Frame::parse_frames(&data.into()) {
	crate::inc!(DERIVED_FRAMES_COUNT, frames.len() as f64, &["success"]);
	self.queue.extend(Frames(frames).validate(&cfg, timestamp));
}

// -- snip --
```

refcell:
### Description

Implements the Holocene FrameQueue.

### Metadata

Closes #557.

### DIFF
--- a/crates/derive/src/stages/frame_queue.rs
+++ b/crates/derive/src/stages/frame_queue.rs
@@ -9,7 +9,8 @@ use alloc::{boxed::Box, collections::VecDeque};
 use alloy_primitives::Bytes;
 use async_trait::async_trait;
 use core::fmt::Debug;
-use op_alloy_genesis::SystemConfig;
+use alloc::sync::Arc;
+use op_alloy_genesis::{RollupConfig, SystemConfig};
 use op_alloy_protocol::{BlockInfo, Frame};
 use tracing::{debug, error, trace};
 
@@ -38,6 +39,8 @@ where
     pub prev: P,
     /// The current frame queue.
     queue: VecDeque<Frame>,
+    /// The rollup config.
+    rollup_config: Arc<RollupConfig>,
 }
 
 impl<P> FrameQueue<P>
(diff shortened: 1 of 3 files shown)


refcell:
* **#579** <a href="https://app.graphite.dev/github/pr/anton-rs/kona/579?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

crates/derive/src/stages/frame_queue.rs
```diff
@@ -47,9 +49,94 @@ where
...
-        Self { prev, queue: VecDeque::new() }
+        Self { prev, queue: VecDeque::new(), rollup_config: cfg }
+    }
+
+    /// Returns if holocene is active.
+    pub fn is_holocene_active(&self) -> bool {
+        let Some(origin) = self.prev.origin() else {
+            warn!(target: "frame-queue", "Failed to get origin");
+            return false;
+        };
```
clabby, line 62: The origin missing is a critical error, `PipelineErrors::MissingOrigin.crit()`

crates/derive/src/stages/frame_queue.rs
```diff
@@ -47,9 +49,94 @@ where
...
+        let data = match self.prev.next_data().await {
+            Ok(data) => data,
+            Err(e) => {
+                debug!(target: "frame-queue", "Failed to retrieve data: {:?}", e);
+                // SAFETY: Bubble up potential EOF error without wrapping.
+                return Err(e);
+            }
+        };
+
+        //
```
clabby, line 124:
```suggestion
        // <some comment / remove comment>
```

crates/derive/src/stages/frame_queue.rs
```diff
@@ -47,9 +49,94 @@ where
...
+    }
+
+    /// Prunes frames if Holocene is active.
+    pub fn prune(&mut self) {
+        if !self.is_holocene_active() {
+            return;
+        }
+
+        let mut i = 0;
+        while i < self.queue.len() - 1 {
```
clabby, line 73: Hmm.. It seems like looping through the full queue every time (we call `self.prune()` every `load_frames` call post-holocene is a bit inefficient. Can we move towards only pruning when the new frame comes in?

It looks like [the spec](https://specs.optimism.io/protocol/holocene/derivation.html#frame-queue) mentions that most of these cases where we drop a frame happen atomically as a frame is absorbed, i.e. the "non-first frame" cases.

For the "first frame, where prev.is_last == false" case, we should be able to just drain the full queue and add the new first frame, without having to iterate through every single one of them.

clabby (changes requested):

crates/derive/src/stages/frame_queue.rs
refcell: oops

clabby (approved):

crates/derive/src/stages/frame_queue.rs
clabby: Following up on this

PR #583, opened 2024-09-28
chore(derive): [Holocene] Drain previous channel in one iteration

clabby:
## Overview

Drains the entire previous channel at one time, rather than over multiple iterations, if the following case is met:

> If a first frame is decoded while the previous frame isn't a last frame (i.e., is_last is false), all previous frames for the same channel are dropped and this new first frame remains in the queue.

Also fixes a small issue w/ the `DERIVED_FRAMES_COUNT` metric's accounting, where it was counting dropped frames.

### Tests

Added a new test for this functionality, when there is a valid, closed channel earlier on in the frame queue @ `test_holocene_replace_channel`.

### DIFF
--- a/crates/derive/src/stages/frame_queue.rs
+++ b/crates/derive/src/stages/frame_queue.rs
@@ -67,33 +67,40 @@ where
 
         let mut i = 0;
         while i < self.queue.len() - 1 {
-            let frame = &self.queue[i];
+            let prev_frame = &self.queue[i];
             let next_frame = &self.queue[i + 1];
+            let extends_channel = prev_frame.id == next_frame.id;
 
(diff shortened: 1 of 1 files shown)


clabby:
* **#583** <a href="https://app.graphite.dev/github/pr/anton-rs/kona/583?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

<h2></h2>

refcell (commented):
I think this makes sense to me. I would like to verify that it passes tests that cover the sad path ruleset where frames with differing channels are interleaved in https://github.com/anton-rs/kona/pull/585.

Maybe we can merge https://github.com/anton-rs/kona/pull/585 and then rebase this on-top?

PR #587, opened 2024-09-30
fix(derive): Pipeline Parameters & Max Channel Bank Size

refcell:
### Description

Cleans up pipeline parameters and fixes the channel bank to use the post-fjord max channel bank size check.


crates/derive/src/stages/l1_traversal.rs
```diff
@@ -146,27 +146,22 @@ impl<F: ChainProvider + Send> ResettableStage for L1Traversal<F> {
...
-    use crate::{
-        errors::PipelineErrorKind,
-        params::{CONFIG_UPDATE_EVENT_VERSION_0, CONFIG_UPDATE_TOPIC},
-        traits::test_utils::TestChainProvider,
-    };
+    use crate::{errors::PipelineErrorKind, traits::test_utils::TestChainProvider};
     use alloc::vec;
     use alloy_consensus::Receipt;
     use alloy_primitives::{address, b256, hex, Bytes, Log, LogData, B256};
+    use op_alloy_genesis::system::{CONFIG_UPDATE_EVENT_VERSION_0, CONFIG_UPDATE_TOPIC};
```
clabby, line 153: not an issue with this PR, but these should probably be in `op-alloy-protocol`

or are they there just because of the `SystemConfig` type?

clabby (approved):

crates/derive/src/stages/l1_traversal.rs
refcell: Yes just because of system config type. genesis is lower-level than protocol.

PR #592, opened 2024-09-30
feat(derive): Span batch prefix checks

clabby:
## Overview

Adds the checks for the span batch prefix.

### DIFF
--- a/crates/derive/src/stages/batch_stream.rs
+++ b/crates/derive/src/stages/batch_stream.rs
@@ -107,6 +107,7 @@ where
 {
     fn flush(&mut self) {
         if self.is_active().unwrap_or(false) {
+            self.prev.flush();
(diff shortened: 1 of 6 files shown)


crates/derive/src/stages/batch_stream.rs
```diff
@@ -124,12 +128,34 @@ where
...
+                        .await;
+
+                    match validity {
+                        BatchValidity::Accept => self.span = Some(b),
+                        BatchValidity::Drop => {
+                            // Flush the stage.
+                            self.flush();
+
+                            // Send a signal to drop the channel that the batch was derived from.
+                            // TODO
```
refcell, line 151:
```suggestion
                            self.prev.flush();
```

refcell (approved):
This looks good!

PR #612, opened 2024-10-03
feat(derive): Holocene flush signal

clabby:
## Overview

Adds the channel flushing functionality for Holocene.


refcell (approved):
Very clean

PR #620, opened 2024-10-03
chore: Span Batch Prefix: no BatchValidity::Future 

refcell:
### Description

TODO: add a test to assert `span_batch_prefix` check never returns a batch validity future.

prop test this?


clabby:
I like the idea of this. Maybe we change this branch up to where `BatchValidity::Future` is a crit?

PR #622, opened 2024-10-04
fix(executor): Holocene EIP-1559 params in Header

clabby:
## Overview

Fixes the inclusion of the current block's base fee parameters within the Header. As-is, it always includes the payload attributes' EIP-1559 parameters in the header, but this is not the desired result.

Instead, we should be including the EIP-1559 parameters that were selected for the block. That is, if the parent header's `nonce` field is `B64::ZERO`, we should be including the canyon base fee parameters in the `Header`, not what was sent through the `PayloadAttributes`.


refcell (approved):
Good catch!

refcell (approved):

PR #645, opened 2024-10-07
chore(mpt): Noop Provider Refactor

refcell:
### Description

Pulls `Noop` provider and hinter into the `test-utils` feat flag.

This increases code coverage % since these noop impls were not, and don't need to be covered as they are test utilities, shared by the benchmark code.

### DIFF
--- a/crates/mpt/src/test_utils.rs
+++ b/crates/mpt/src/test_utils.rs
@@ -12,8 +12,7 @@ use reqwest::Url;
 const RPC_URL: &str = "https://docs-demo.quiknode.pro/";
 
 /// Grabs a live merkleized receipts list within a block header.
-pub async fn get_live_derivable_receipts_list(
-) -> Result<(B256, BTreeMap<B256, Bytes>, Vec<ReceiptEnvelope>)> {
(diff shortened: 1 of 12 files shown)


crates/mpt/src/test_utils.rs
```diff
@@ -118,15 +118,62 @@ pub(crate) async fn get_live_derivable_transactions_list(
...
+    }
+
+    fn header_by_hash(&self, _hash: B256) -> Result<Header, Self::Error> {
+        Ok(Header::default())
+    }
+}
+
+/// The default, no-op implementation of the [TrieHinter] trait, used for testing.
+#[derive(Debug, Clone, Copy)]
+pub struct NoopTrieHinter;
```
clabby, line 143: I don't think these belong in test utils unfortunately. Succinct and other downstream implementers use this - non FPVM targets don't have hints.

clabby (changes requested):

crates/mpt/src/test_utils.rs
clabby: Specifically the hinter - if we really want to we can keep the provider behind the test utils, but may as well also distribute it if we have the noop hinter distributed.

refcell:
Replaced by https://github.com/anton-rs/kona/pull/649

PR #648, opened 2024-10-07
feat(executor): Update EIP-1559 configurability

clabby:
## Overview

Updates `kona-executor`'s implementation of configurable EIP-1559 parameters, per the specification update in https://github.com/ethereum-optimism/specs/pull/418.

Instead of using the parameters within the payload for the current block, we use the parameters encoded in the parent block header (or the canyon constants, if the parent header's nonce == 0). When encoding the parameters for placing in the header, we use the `eip1559Params` within the payload (or the canyon constants, if payload.eip1559Params == 0).

### DIFF
--- a/crates/executor/src/lib.rs
+++ b/crates/executor/src/lib.rs
@@ -784,17 +784,14 @@ mod test {
 
     #[test]
     fn test_encode_canyon_1559_params() {
-        let cfg = RollupConfig { 
+        let cfg = RollupConfig {
             canyon_base_fee_params: BaseFeeParams {
                 max_change_denominator: 32,
                 elasticity_multiplier: 64,
             },
             ..Default::default()
         };
-        assert_eq!(
-            encode_canyon_base_fee_params(&cfg),
-            b64!("0000002000000040")
-        );
+        assert_eq!(encode_canyon_base_fee_params(&cfg), b64!("0000002000000040"));
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


crates/executor/src/lib.rs
```diff
@@ -306,19 +310,25 @@ where
...
-                let mut encoded_params = B64::ZERO;
-                encoded_params[0..4].copy_from_slice(
-                    (base_fee_params.max_change_denominator as u32).to_be_bytes().as_ref(),
-                );
-                encoded_params[4..8].copy_from_slice(
-                    (base_fee_params.elasticity_multiplier as u32).to_be_bytes().as_ref(),
-                );
-                encoded_params
+                let payload_params =
+                    payload.eip_1559_params.ok_or(ExecutorError::MissingEIP1559Params)?;
```
refcell, line 324: Can we pull this into a separate method and test it for codecov 🙏

crates/executor/src/lib.rs
```diff
@@ -696,6 +713,18 @@ where
     }
 }
 
+/// Encodes the canyon base fee parameters, per Holocene spec.
+///
+/// <https://specs.optimism.io/protocol/holocene/exec-engine.html#eip1559params-encoding>
+fn encode_canyon_base_fee_params(config: &RollupConfig) -> B64 {
```
refcell, line 719: test pls

refcell (approved):
Logic is sound

cody-wang-cb (commented):
LGTM

PR #650, opened 2024-10-07
chore(executor): Test Coverage over Executor Utilities

refcell:
### Description

Adds test coverage over utilities.


refcell:
These utils will be upstreamed into op-alloy, but for now let's have coverage.

clabby (approved):

PR #651, opened 2024-10-08
chore(executor): Use Upstreamed op-alloy Methods 

refcell:
### Description

Cleans up `op-alloy` type use and custom functions that are upstreamed.


crates/executor/src/util.rs
```diff
@@ -29,8 +29,8 @@ pub(crate) fn receipt_envelope_from_parts<'a>(
         OpTxType::Eip1559 => {
             OpReceiptEnvelope::Eip1559(ReceiptWithBloom { receipt: inner_receipt, logs_bloom })
         }
-        OpTxType::Eip4844 => {
-            OpReceiptEnvelope::Eip4844(ReceiptWithBloom { receipt: inner_receipt, logs_bloom })
+        OpTxType::Eip7702 => {
+            OpReceiptEnvelope::Eip7702(ReceiptWithBloom { receipt: inner_receipt, logs_bloom })
```
clabby, line 33: We cannot use 7702, Pectra isn't active - this should throw an error.

clabby (approved):

PR #667, opened 2024-10-09
chore(host): Support environment variables for `kona-host` flags

clabby:
## Overview

> [!NOTE]
> Depends on #666 
>
> need `st` to remove this warning 😢 

Allows for CLI flags to be passed as environment variables in `kona-host`. Some CLI flags are excluded from this:
* `-v[vvvv]`
* `--exec`
* `--server`

mostly because these explicitly select the mode that the host runs in, or for trace verbosity, doesn't really make sense to define within the environment.


refcell (approved):

refcell:
Very nice

PR #670, opened 2024-10-09
chore(derive): Test and Clean Batch Types

refcell:
### Description

Tests and cleans derivation batch types.


clabby (approved):
very nice mr allen

PR #674, opened 2024-10-09
feat: Bump Codecov Threshold 

refcell:
### Description

Bumps the CODECOV threshold to 90%


refcell:
I'm thinking #638 will bring kona to around ~95% which will give us enough of a buffer to avoid regressions hitting the threshold.

clabby (approved):

PR #684, opened 2024-10-10
feat(derive): `Past` batch validity variant

clabby:
## Overview

Adds a new variant to the `BatchValidity` enum that activates with Holocene. When a batch is outdated, after Holocene, the `Past` variant is returned. In the `BatchQueue`, this instructs a silent drop, _without_ flushing the active span batch and channel.

We should follow this PR up with removing the additional check within `add_batch`, deferring to the validity checks in `derive_next_batch`.

### DIFF
--- a/crates/derive/src/batch/validity.rs
+++ b/crates/derive/src/batch/validity.rs
@@ -15,12 +15,20 @@ pub enum BatchValidity {
     Undecided,
     /// The batch may be valid, but cannot be processed yet and should be checked again later
     Future,
+    /// Introduced in Holocene, a special variant of the Drop variant that signals not to flush
+    /// the active batch and channel.
+    BatchOutdated,
 }
 
 impl BatchValidity {
     /// Returns if the batch is dropped.
     pub const fn is_drop(&self) -> bool {
         matches!(self, Self::Drop)
+    }
+
+    /// Returns if the batch is outdated.
+    pub const fn is_outdated(&self) -> bool {
+        matches!(self, Self::BatchOutdated)
     }
 
     /// Returns if the batch is future.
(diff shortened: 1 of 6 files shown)


crates/derive/src/batch/validity.rs
```diff
@@ -15,6 +15,9 @@ pub enum BatchValidity {
     Undecided,
     /// The batch may be valid, but cannot be processed yet and should be checked again later
     Future,
+    /// Introduced in Holocene, a special variant of the Drop variant that signals not to flush
+    /// the active batch and channel.
+    BatchOutdated,
```
sebastianst, line 20: Following the others enum names, wouldn't you just name this `Outdated`? Note that I name the validity `past` in the specs. Or do you strongly prefer `outdated` over `past`? I thought that `past` pairs better with `future`.

crates/derive/src/stages/batch_queue.rs
```diff
@@ -261,6 +270,9 @@ where
         if drop {
             self.prev.flush();
```
sebastianst, line 271: So do you now also flush pre-holocene?

crates/derive/src/batch/validity.rs
clabby: Not opinionated. We can standardized on the `Past` name.

crates/derive/src/stages/batch_queue.rs
clabby: Good catch - no, we shouldn't be.

crates/derive/src/batch/validity.rs
sebastianst: Ok, also using `past` in the Go implementation now.

crates/derive/src/stages/batch_queue.rs
sebastianst: does this also prevent flushing in the batchqueue itself?

refcell:
This LGTM - rip coverage

crates/derive/src/stages/batch_queue.rs
refcell: Yes it does - is that incorrect?

sebastianst: It also needs to be flushed if a batch is found to be invalid (`drop`).

PR #686, opened 2024-10-10
feat: Published thiserror with no_std support

refcell:
### Description

Use a published version of `thiserror` with no_std support so kona crates can be released.


refcell:
Opened an issue to track this: https://github.com/anton-rs/kona/issues/720

PR #692, opened 2024-10-14
fix(derive): SpanBatch element limit + channel RLP size limit

clabby:
## Overview

The enforcements on span batch size for Fjord was implemented incorrectly. There is a separate protocol constant, `MaxSpanBatchElements`, in the reference implementation that limits span batch element size. This is strictly different than the channel size limitation.

We also did not enforce maximum channel RLP size within the `BatchReader` for zlib. This has been fixed.


crates/derive/src/stages/batch_queue.rs
```diff
@@ -523,7 +523,7 @@ mod tests {
         let file_contents = &(&*file_contents)[..file_contents.len() - 1];
         let data = alloy_primitives::hex::decode(file_contents).unwrap();
         let bytes: alloy_primitives::Bytes = data.into();
-        BatchReader::from(bytes)
+        BatchReader::new(bytes, MAX_RLP_BYTES_PER_CHANNEL_FJORD as usize)
```
refcell, line 526: What about pre-fjord here?

refcell (approved):
Generally happy with this as it removes a bunch of hardfork-specific logic. Really think this demonstrates the need to upstream logic into op-alloy

PR #693, opened 2024-10-15
feat(derive): Stage multiplexer

clabby:
## Overview

Implements a stage multiplexer macro, which creates a shim above several stage implementations and has the ability to re-link the pipeline at runtime. Upon configured fork activation, a new stage is created at runtime and the ownership of the previous stage is transferred to the new stage.

This method of stage multiplexing:
1. Preserves the use of generic types over dynamic dispatch.
2. Requires no duplication of the pipeline; It is re-linked at runtime.
3. Is generic enough to work with other stages, due to the multiplexer being implemented as a macro.
    - Doesn't make the nested generics worse, since the sub-stages are contained within the multiplexer, and only generic over the previous stage.

Downsides of this method:
1. ~~With the current macro implementation, all stages must have the same `new` signature (`rollup_cfg, previous stage`).~~ e: changed, can now also be used for other stages that need more parameters / generics and can benefit from multiplexing (`BatchQueue`)
2. This macro is currently partially unhygenic, but that should be fine for the sake of `kona-derive`.

## `ChannelAssembler`

To test this abstraction, I implemented the `ChannelAssembler` - a simplified `ChannelBank` that only has to contend with a single channel, activating upon Holocene activation. This implementation is similar to the method upstream went with to simplify the pipeline along with Holocene.

## Visual

Untitled (16)


crates/derive/src/stages/channel/channel_provider.rs
refcell, line 1: We should be able to add simple tests to this using test stage providers.

refcell: Part of the reason to have codecov is to get into the habit of adding tests as we make changes.

crates/derive/src/stages/multiplexed.rs
refcell, line 1: Generally this makes it harder to follow along what's happening in the pipeline. There's no test coverage. Why not do away with the macro and just implement on the actual stages?

clabby: For reusability. This will also be used for other multiplexed stages, i.e. the `BatchQueue`, if you're a fan of the pattern.

crates/derive/src/stages/channel/channel_provider.rs
clabby: I wanted your feedback on the pattern before committing - kept in draft for that reason

refcell (approved):
I'll approve this, but I could see this macro biting us down the road. Hopefully it will be easy to modify if that happens

PR #703, opened 2024-10-17
feat(derive): `BatchValidator` stage

clabby:
## Overview

Adds a new stage to take place of the `BatchQueue` after Holocene activation, the `BatchValidator`. This stage's only role is to consume the `SingleBatch`es from the `BatchStream`, validate them, and send them off to the `AttributesQueue`.

This PR does not hook this stage up to a multiplexer stage with the `BatchQueue`, and that will be done in a follow-up.

### DIFF
--- a/crates/derive/src/stages/batch/batch_validator.rs
+++ b/crates/derive/src/stages/batch/batch_validator.rs
@@ -51,13 +51,25 @@ where
     }
 
     /// Returns `true` if the pipeline origin is behind the parent origin.
+    ///
+    /// ## Takes
+    /// - `parent`: The parent block of the current batch.
+    ///
+    /// ## Returns
+    /// - `true` if the origin is behind the parent origin.
     fn origin_behind(&self, parent: &L2BlockInfo) -> bool {
         self.prev.origin().map_or(true, |origin| origin.number < parent.l1_origin.number)
     }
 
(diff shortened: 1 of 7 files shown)


crates/derive/src/stages/batch/batch_validator.rs
```diff
@@ -0,0 +1,622 @@
...
+        // preserve existing l1 blocks to verify the epochs of the next candidate batch.
+        if !self.l1_blocks.is_empty() && parent.l1_origin.number > self.l1_blocks[0].number {
+            for (i, block) in self.l1_blocks.iter().enumerate() {
+                if parent.l1_origin.number == block.number {
+                    self.l1_blocks.drain(0..i);
+                    debug!(target: "batch-validator", "Advancing internal L1 epoch");
+                    break;
+                }
+            }
+            // If the origin of the parent block is not included, we must advance the origin.
```
refcell, line 115: ?

refcell (approved):
Looks good but it seems like there's a lot of duplicate logic with the `batch_queue`. Is it not possible to de-dup into a split out validator that holds the l1 blocks? Happy to allow any enhancements to be follow on since this works.

crates/derive/src/stages/batch/batch_validator.rs
clabby: This comment is correct - maybe better put above the loop, though?

PR #717, opened 2024-10-19
feat(workspace): Distribute pipeline, not providers

clabby:
## Overview

Experiment to distribute `DerivationPipeline` types, as `kona-providers-alloy` is doing today. Our abstractions are a bit off right now, with `kona-providers` holding traits specific to `kona-derive`, without ever needing to be used alongside it. However, `kona-providers-alloy` is already following a great pattern by distributing the pipeline itself with a constructor for the configuration as `new_online_pipeline`, as the abstractions intended.

This pattern keeps cyclical dependencies away, and because we're distributing the pipeline, there's no real need for `kona-providers`. We can also extend this pattern further, with a good cleanup of the oracle-backed pipeline into a new `kona-derive-oracle` crate.

This is also how modifications like alt DA intend to ship their pipeline type variants as well.


README.md
```diff
@@ -59,11 +59,13 @@ see the [SDK section of the book](https://anton-rs.github.io/kona/sdk/intro.html
 
 - [`common`](./crates/common): A suite of utilities for developing `client` programs to be run on top of Fault Proof VMs.
 - [`common-proc`](./crates/common-proc): Proc macro for the `client` program entrypoint.
-- [`primitives`](./crates/primitives): Primitive types for use in `kona` crates.
 - [`preimage`](./crates/preimage): High level interfaces to the [`PreimageOracle`][fpp-specs] ABI
+
+**Protocol**
```
refcell, line 64: Pipeline

crates/providers-alloy/README.md
refcell, line 1: Seems like not all of `providers-alloy` was removed here?

crates/derive/src/sources/blobs.rs
refcell, line 1: As part of this pr can we merge the `sources` module into `traits`? sources are just a specific subset of traits that `kona-derive` needs. We don't need to specify them as "sources"

refcell (commented):
I'm a proponent of this refactor. Left a few comments that I'd like to see addressed prior to merging.

refcell:
I'd also like to get this in first: https://github.com/anton-rs/kona/pull/713

README.md
clabby: Also has the implementation of the executor + Ethereum MPT. Not just pipeline

refcell (approved):

PR #733, opened 2024-10-24
fix(derive): Holocene action tests / fixes

clabby:
## Overview

This PR updates the action tests in `kona` to the latest version of the monorepo. It also contains small fixes. If throughout getting action tests updated, there are bigger issues, I'll stack the fixes on top of this PR.


justfile
```diff
@@ -42,7 +42,7 @@ action-tests test_name='Test_ProgramAction':
   export KONA_CLIENT_PATH="{{justfile_directory()}}/target/release-client-lto/kona"
 
   cd monorepo/op-e2e/actions/proofs && \
-    go test -run "{{test_name}}" -v ./...
+    go test -run "{{test_name}}" -v -count=1 ./...
```
refcell, line 45: Why do we set `count` to 1 here?

justfile
```diff
@@ -151,7 +151,7 @@ build-client-prestate-asterisc-image kona_tag asterisc_tag out='./prestate-artif
 
 # Clones and checks out the monorepo at the commit present in `.monorepo`
 monorepo:
-  [ ! -d monorepo ] && git clone https://github.com/ethereum-optimism/monorepo
+  ([ ! -d monorepo ] && git clone https://github.com/ethereum-optimism/monorepo) || exit 0
```
refcell, line 154: I think this change makes it so if the repo already exists, it exits... Should we not leave it so it checks out the commit?

refcell (approved):
Two questions, but otherwise changes look good

justfile
clabby: To make sure the tests always run with no cached results.

justfile
clabby: Right now it will exit still, but it won't be able to check out. If the `&&` fails, we `exit 1`. Just wanted there to be no visible error.

PR #755, opened 2024-10-30
chore(executor): Remove Upstream Utility

refcell:
### Description

Small PR to use the upstream receipt constructor in `op-alloy` added in https://github.com/alloy-rs/op-alloy/pull/165


crates/executor/src/executor/mod.rs
```diff
@@ -205,7 +205,7 @@ where
             cumulative_gas_used += result.gas_used();
 
             // Create receipt envelope.
-            let receipt = receipt_envelope_from_parts(
+            let receipt = OpReceiptEnvelope::<Log>::from_parts(
```
clabby, line 208: This is a bit dangerous, as upstream alloy allows 7702 txs through as Ithica needs it for Odyssey, but we explicitly need to disallow them. We need to make sure that we add enforcement for fork activation.

clabby (changes requested):

clabby (approved):

Issue #724, opened 2024-10-21
fix(client): Invalidate Oracle Cache on Pipeline Reorg


refcell:
### Description

Currently, `kona-client`'s caching oracle is not invalidate on reorg. This should be fixed in the [`kona_client::DerivationDriver`](https://github.com/anton-rs/kona/blob/main/bin/client/src/l1/driver.rs#L74) when a temporary error is found to match `ResetError::ReorgDetected(_, _)`, the `CachingOracle` is flushed.

refcell:
Closed by https://github.com/anton-rs/kona/pull/756

PR #756, opened 2024-10-30
feat: flush oracle cache on reorg #724

piotr-roslaniec:


bin/client/src/l1/driver.rs
```diff
@@ -280,6 +287,9 @@ where
                                         .signal(),
                                     )
                                     .await?;
+                            } else if matches!(e, ResetError::ReorgDetected(_, _)) {
+                                self.caching_oracle.as_ref().flush();
+                                self.pipeline.signal(Signal::FlushChannel).await?;
```
refcell, line 292: We will need to perform a full reset on a reorg and not just a channel flush.
Basically, we need to check inside the `else` branch below if `e` matches the `ResetError::ReorgDetected(_, _)`, then flush the caching oracle and resume with sending a reset signal.

e.g.
```rust
  } else {
      if matches!(e, ResetError::ReorgDetected(_, _)) {
            self.caching_oracle.as_ref().flush();
      }
      // Reset the pipeline to the initial L2 safe head and L1 origin,
      // and try again.
      self.pipeline
          .signal(
              ResetSignal {
                  l2_safe_head: self.l2_safe_head,
                  l1_origin: self
                      .pipeline
                      .origin()
                      .ok_or_else(|| anyhow!("Missing L1 origin"))?,
                  system_config: Some(system_config),
              }
              .signal(),
          )
          .await?;
  }
```

refcell (changes requested):
Nice work, nearly there with a few changes

refcell:
Great work @piotr-roslaniec!

clabby (approved):

PR #757, opened 2024-10-30
chore(derive): Cleanup Exports

refcell:
### Description

Cleans up exports in `kona-derive`, part of a stack.


refcell:
## 📚 $\text{Stack Overview}$

* #757 👈
* `main`

crates/derive/src/attributes/mod.rs
```diff
@@ -1,616 +1,4 @@
-//! The [`AttributesBuilder`] and it's default implementation.
+//! Module containing `AttributeBuilder` implementations.
```
clabby, line 1:
```suggestion
//! Module containing [AttributesBuilder] trait implementations.
```

clabby (approved):
lgtm. always has felt weird to me that it's called **Stateful**AttributesBuilder. The `prepare_payload_attributes` doesn't reference any local state except for the rollup cfg.

PR #759, opened 2024-10-30
chore: Derivation Import Cleanup

refcell:
### Description

Small PR to clean up imports from `kona_derive` to use the `kona_derive::prelude`.

### DIFF
--- a/crates/derive-alloy/src/pipeline.rs
+++ b/crates/derive-alloy/src/pipeline.rs
@@ -1,14 +1,6 @@
 //! Helper to construct a [DerivationPipeline] using online types.
 
-use kona_derive::{
-    attributes::StatefulAttributesBuilder,
-    pipeline::{DerivationPipeline, PipelineBuilder},
-    sources::EthereumDataSource,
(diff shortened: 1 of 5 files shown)


crates/derive-alloy/src/pipeline.rs
```diff
@@ -1,14 +1,6 @@
...
-use kona_derive::{
-    attributes::StatefulAttributesBuilder,
-    pipeline::{DerivationPipeline, PipelineBuilder},
-    sources::EthereumDataSource,
-    stages::{
-        AttributesQueue, BatchProvider, BatchStream, ChannelProvider, ChannelReader, FrameQueue,
-        L1Retrieval, L1Traversal,
-    },
-};
+use kona_derive::prelude::*;
```
clabby, line 3: Not a huge fan of importing `*`, it causes a lot more things to be linked with this module, making resulting binaries a bit bigger. Makes it a good bit harder for LLVM to do dead code elimination.

clabby (changes requested):

Issue #721, opened 2024-10-21
feat(derive): Metrics


refcell:
### Description

Metrics in `kona-derive` are a set of prometheus types that are spread across the pipeline stages using helper macros. The helper macros handle feature flagging since `metrics` are not available by default.

We want to provide a more stable and structured way for pipeline consumers to handle metrics.
Each stage should define its own metrics through a trait abstraction.
By default, pipeline construction will provide `no-op` implementations.

### Output

- [ ] Remove current metrics in `kona-derive`.
- [ ] Introduce stage-wise metricer traits with Noop Implementations.
- [ ] Introduce a `metrics` module with prometheus-backed metric collection.

steph-rs:
hey, can I take it? 🙏

refcell:
> hey, can I take it? 🙏

That would be excellent @steph-rs! We've already removed metrics from `kona-derive` and `kona-derive-alloy`. (See linked PRs below). The next step is basically to introduce a metricer trait that is able to consume metrics. I'd suggest to start by adding metrics to the top-level [`DerivationPipeline`](https://github.com/anton-rs/kona/blob/main/crates/derive/src/pipeline/core.rs#L21). Some useful metrics may be:
- Metrice the step results in the [`DerivationPipeline::step`](https://github.com/anton-rs/kona/blob/main/crates/derive/src/pipeline/core.rs#L150) method.
- Metrice the number of times the pipeline received [Reset signal](https://github.com/anton-rs/kona/blob/main/crates/derive/src/pipeline/core.rs#L94).
- Metrice the number of times the pipeline received a [FlushChannel signal](https://github.com/anton-rs/kona/blob/main/crates/derive/src/pipeline/core.rs#L119).

- Remove metrics from `kona-derive-alloy`: https://github.com/anton-rs/kona/pull/748
- Remove metrisc from `kona-derive`: https://github.com/anton-rs/kona/pull/743

PR #760, opened 2024-10-30
feat: `PipelineMetrics` common scheme, `no-op` version

steph-rs:
### Summary of Metrics Abstraction

In this PR, I've introduced a `PipelineMetrics` struct that contains trait objects for metrics (e.g., `Arc<dyn DerivationPipelineMetrics + Send + Sync>`). This allows us to pass a single metrics object through the pipeline without adding generics for each metric trait, simplifying type signatures and reducing complexity.

I implemented `DerivationPipelineMetrics` for `PipelineMetrics` by delegating the methods to the inner `derivation_pipeline_metrics` field. We also have a `PipelineMetrics::no_op()` method that provides standard no-op implementations.

### Question

Could you please review this abstraction scheme? I initially tried using generics for each metric, but this made the code too complicated. What do you think of this model? Any suggestions for improvement are welcome.


steph-rs:
subtask from #721

crates/derive/.idea/.gitignore
refcell, line 1: Can we gitignore the `.idea` directory here and in `mpt`. Probably can add a `**/.idea` to the top-level `.gitignore`.

steph-rs: oops, sorry I'll fix that

crates/derive/src/pipeline/core.rs
```diff
@@ -95,6 +106,8 @@ where
         match signal {
             s @ Signal::Reset(ResetSignal { l2_safe_head, .. }) |
             s @ Signal::Activation(ActivationSignal { l2_safe_head, .. }) => {
+                self.metrics.inc_reset_signals();
```
refcell, line 109: What do you think about using the `record_` syntax for most metrics?

e.g. instead of `inc_reset_signals()` and `inc_flush_channel_signals()`, we could use a more generic `record_signal(Signal)` method like you did below with `record_step_result(StepResult)`

refcell: This allows the metrics implementation to have more flexibility with the metric types (whether it is an increment or an observation etc etc). It may also allow for better forwards compatibility; imagine if there is a new `Signal` variant, the metrics implementation may use a histogram with the signal variants as labels, requiring no changes to be made to neither the metrics interface nor the metrics impl.

crates/derive/src/pipeline/builder.rs
```diff
@@ -127,7 +136,9 @@ where
         let l2_chain_provider = builder.l2_chain_provider.expect("chain_provider must be set");
         let dap_source = builder.dap_source.expect("dap_source must be set");
         let attributes_builder = builder.builder.expect("builder must be set");
+        let metrics = builder.metrics.expect("metrics must be set");
 
+        // todo: add metrics to the stages
```
refcell, line 141: Preference to create a sub-ticket in github issues over leaving TODO comments in the codebase.

crates/derive/src/metrics/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+
+use alloc::sync::Arc;
+use core::fmt::Debug;
+
+use crate::traits::{DerivationPipelineMetrics, StepResult};
+
+/// Composite metrics struct containing metrics for all stages.
+pub struct PipelineMetrics {
+    pub(crate) derivation_pipeline_metrics: Arc<dyn DerivationPipelineMetrics + Send + Sync>,
+    // todo: add more metrics here for each stage
```
refcell, line 13: Preference to create a sub-ticket in github issues over leaving TODO comments in the codebase.

crates/derive/src/metrics/noop.rs
```diff
@@ -0,0 +1,28 @@
+use crate::{metrics::PipelineMetrics, pipeline::StepResult, traits::DerivationPipelineMetrics};
+use alloc::sync::Arc;
+
+impl PipelineMetrics {
+    /// No-op implementation for `PipelineMetrics`.
+    pub fn no_op() -> Self {
+        Self {
+            derivation_pipeline_metrics: Arc::new(NoopDerivationPipelineMetrics),
+            // todo: add more metrics here for each stage
```
refcell, line 9: Preference to create a sub-ticket in github issues over leaving TODO comments in the codebase.

steph-rs: I'll finish the no-op metrics for all stages in this PR once you've approved the abstractions schema

crates/derive/src/metrics/mod.rs
```diff
@@ -0,0 +1,34 @@
...
+mod noop;
+
+use alloc::sync::Arc;
+use core::fmt::Debug;
+
+use crate::traits::{DerivationPipelineMetrics, StepResult};
+
+/// Composite metrics struct containing metrics for all stages.
+pub struct PipelineMetrics {
+    pub(crate) derivation_pipeline_metrics: Arc<dyn DerivationPipelineMetrics + Send + Sync>,
```
refcell, line 12: I think we can remove the `Arc` here - it can be placed around the `PipelineMetrics`.

If we do this, we can then make a `PipelineMetrics::NOOP` `const`. We should use a generic here bound to `DerivationPipelineMetrics + Send + Sync`.

e.g.
```rust
impl<M> PipelineMetrics<M> where M: DerivationPipelineMetrics + Send + Sync {
    /// A no-op implementation for derivation `PipelineMetrics`.
    pub const NOOP: Self<M> = Self { 
        inner: NoopDerivationPipelineMetrics,
    };
}
```

steph-rs: I'd love to do that btw, but how should other metrics be implemented?

refcell: I see what you're saying... Let me think about this a little bit and get back to you. What you're currently proposing is to place the various stage metrics inside `PipelineMetrics` and then each stage holds a `PipelineMetrics` which implements any metrics trait required through the internal field for that stage? Is that right?

e.g.
```rust
struct AttributesQueue {
    metrics: PipelineMetrics,
    ...
}
```

```rust
PipelineMetrics {
    attributes_queue_metrics: Arc<dyn AttributesQueueMetrics + Send + Sync>,
    ...
}

impl AttributesQueueMetrics for PipelineMetrics {
   fn some_metrics_method() {
      self.attributes_queue_metrics.some_metrics_method();
   }
}
```

steph-rs: yeah, exactly
but I think there's a typo in first block:
```rust
pub struct AttributesQueue<P, AB, M>
where
    M: AttributesQueueMetrics + Send + Sync,
{
    ...
    metrics: M,
}
```

and then we could use `self.metrics.some_metrics_method()`

I'm not sure about this idea, but it can help us easily add other versions as `no-op`, or `prometheus`. I thought this was a problem we wanted to solve.

Does it make sense?

refcell: Ok, I thought you're intention of placing the various metric types inside `PipelineMetrics` was to be able to clone PipelineMetrics (doing `Arc::clone` for each internal `Arc<dyn ..>` field) and then passing it as a concrete type to each stage. This way, you could just avoid having another generic on each stage, and instead just import the trait that is implemented on the `PipelineMetrics` type.

This is why in my first block above I specific `PipelineMetrics` explicitly. And yes like you said - the problem to solve is having different metrics. since `PipelineMetrics` itself has dynamic fields, you can swap in a prometheus, or no-op, or whatever individual metrics easily, without any api changes.

Let me know if this makes sense. Happy to open a draft PR into this branch to show you a diff of what I'm getting at here.

steph-rs: Hey @refcell!
Could you please check the fixes I just pushed and the way I work with `AttributesQueueMetrics`. If it's okay with you, I'll add other no-op metrics.
But since I don't really see the whole picture of what we actually need in metrics, I will ask you to define the exact methods we want. It will be much appreciated.

steph-rs:
Hi @refcell, it's been a long time for this PR...
I just pushed all the noop metrics for each stage, please check it out.
The use of the metrics still needs some work. Please don't be afraid: all metrics are `record_*`. I did it on purpose so that we can change it and don't forget anything.
Also, I'm not quite sure about the files (I created a few in `src/traits` for each stage). 
Please let me know what you think about all this.

steph-rs:
hey guys, it's been a while. should we close this PR if it's not relevant? 
@refcell

refcell:
Hey @steph-rs super sorry about not following up on this - a lot of work has been de-prioritized and we're not committed to a consensus node yet where these metrics would be used. For now, let's close this PR and use it as a reference in the future when we do need to add metrics. Again, apologies for the confusion and lag here - failure on my part 🙇

PR #764, opened 2024-10-31
Only fill blob data when there is no calldata

lightning-li:


crates/derive/src/sources/blobs.rs
```diff
@@ -170,8 +170,8 @@ impl BlobData {
         blobs: &[Box<Blob>],
         index: usize,
     ) -> Result<(), BlobDecodingError> {
-        // Do not fill if there is no calldata to fill
-        if self.calldata.as_ref().map_or(false, |data| data.is_empty()) {
+        // Do not fill if there is calldata here
+        if self.calldata.as_ref().map_or(false, |data| !data.is_empty()) {
             return Ok(());
         }
```
refcell, line 176: We want to return `Ok(())` if the data is empty. Prior to your change, this did that. If the data is empty, `map_or` returns `true` which enters the branch to `return Ok(())`. The new logic now says if the data is not empty, return Ok(()).

lightning-li: Why returning ok when calldata is empty? If the calldata is not empty, the old codes would fill the blob.data. It means the calldata and data fields are both not empty at the same time. Is this allowed? In the golang code https://github.com/ethereum-optimism/optimism/blob/6f2058c4603c389c6d30d3d8f541508ee1eccf36/op-node/rollup/derive/blob_data_source.go#L157, when the calldata is not nil, it will continue which means that if the calldata isn't empty, we shouldn't fill the blob again

refcell: Thank you for clarifying - I believe I was mistaken here. You're suggestion and relevant fix with incrementing the blob index looks correct.

refcell (approved):

PR #768, opened 2024-11-01
feat(docs): Derivation Docs

### LINKED ISSUE #554: feat(book): `kona-derive` extension docs
## Overview

`kona-derive` is a very extensible crate, allowing for swapping out stages entirely, defining custom data sources, custom DA layers, etc.

For people looking to build new protocol features, it's important for this sort of functionality to be documented at a high level so that consumers can get an idea of what they're working with, without having to crawl through the rust docs etc. The book should have a section under the Kona SDK for things like:
* Creating a new DA provider, i.e. the `DataAvailabilityProvider` trait / `kona-plasma`.
* Creating a new stage implementation that can be swapped in.
* Defining a custom payload attributes builder

These docs should be focused primarily on how to extend the pipeline to create non-canonical OP Stack features. Extending the pipeline for custom data sources is already covered @ https://anton-rs.github.io/kona/sdk/custom-backend.html

Example here being the [block executor docs](https://anton-rs.github.io/kona/sdk/exec-ext.html), though these are largely able to point towards `revm` since we re-use their handler abstractions. The derive docs should showcase the abstractions we provide.

refcell:
### Description

Document `kona-derive` in the kona book including swapping out stages, extending signals, defining custom providers, and more.

### Metadata

Fixes #554


refcell:
## 📚 $\text{Stack Overview}$

* #768 👈
* `main`

clabby (approved):
🚀

PR #787, opened 2024-11-06
feat(TrieProvider): Abstract TrieNode retrieval

hashcashier:
This PR refactors the `TrieProvider` to change the the method for fetching trie node data into one that provides `TrieNode` instances instead of raw rlp-encoded bytes:
**Before**
```rust
fn trie_node_preimage(&self, key: B256) -> Result<Bytes, Self::Error>;
```
**After**
```rust
fn trie_node_by_hash(&self, key: B256) -> Result<TrieNode, Self::Error>;
```

This allows `TrieProvider` implementations to abstract away the details of how the required instance can be created and authenticated (i.e. through RLP decoding pre-image oracle answers on-demand or otherwise). This is useful for providing entire trees as query answers, instead of single nodes with blinded children.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -511,6 +511,7 @@ dependencies = [
  "arrayvec",
  "derive_more",
  "nybbles",
+ "serde",
  "smallvec",
  "tracing",
 ]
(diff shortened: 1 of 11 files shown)


clabby (approved):
Fan of this change. LGTM

PR #794, opened 2024-11-07
feat(driver): Refactor Driver out of Client

refcell:
### Description

Refactors the driver out of `kona-client` and into a new crate with trait abstractions.

### Metadata

Fixes #765

PR #796, opened 2024-11-08
feat(driver): Abstract, Default Pipeline

refcell:
### Description

Abstracts the driver pipeline trait to allow for a default payload production loop.

Avoids driver implementors from footgunning themselves with rolling their own payload loops.

### DIFF
--- a/crates/driver/src/core.rs
+++ b/crates/driver/src/core.rs
@@ -1,12 +1,13 @@
 //! The driver of the Derivation Pipeline.
 
 use alloc::vec::Vec;
-use alloy_consensus::{BlockBody, Header, Sealable, Sealed};
+use alloy_consensus::{BlockBody, Sealable};
 use alloy_primitives::B256;
 use alloy_rlp::Decodable;
 use core::fmt::Debug;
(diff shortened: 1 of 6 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #796 👈
* #794
* `main`

crates/driver/src/core.rs
```diff
@@ -166,7 +164,7 @@ where
             // Update the safe head.
             self.cursor.l2_safe_head = L2BlockInfo::from_block_and_genesis(
                 &block,
-                &self.pipeline.rollup_config().genesis,
+                &self.pipeline.inner().rollup_config().genesis,
```
clabby, line 167: We should expose a `rollup_config` fn so we don't have to `.inner()`.

clabby: same w/ `step`

clabby (commented):
LGTM - to reduce the feeling of nesting, it'd be nice to lift some of the `step` / `inner` etc. functions into the pipeline itself.

clabby (approved):

PR #816, opened 2024-11-14
chore(workspace): Reorganize SDK

clabby:
## Overview

Reorganizes the proof SDK crates into a `proof-sdk` subdirectory, to differentiate them from the protocol. This is in preparation for improving some of the platform abstractions
as noted in https://github.com/anton-rs/kona/issues/815 + https://github.com/anton-rs/kona/issues/553. Most of the client program will be broken out into a more generic library,
and `kona-preimage` will be decoupled from `kona-common` to enable it to be a fault-proof-vm target specific library, and not leak into the proof SDK.

PR #813, opened 2024-11-14
feat(mpt): Extend `TrieProvider` in `kona-executor`

clabby:
## Overview

Cleans up the leaky `TrieProvider` abstraction by extending the `TrieProvider` trait with the added functionality it needs to back a database rather than just a merkle patricia trie.

The `TrieProvider` trait served by `kona-mpt` now has a much simpler interface, just fetching trie node preimages.

PR #817, opened 2024-11-14
feat(preimage): Decouple from `kona-common`

### LINKED ISSUE #398: fix(common): remove ZkvmIO

clabby:
## Overview

Introduces a `Channel` abstraction into `kona-preimage` that decouples it from `kona-common`'s IO abstractions. This allows for `kona-preimage` consumers to not have to integrate with `kona-common`, for example for the sake of witness generation runs in `kona-host`.

This also will allow for `kona-host` to run `kona-client` natively, using in-memory channels for communication rather than an operating system pipe across processes.

Temporarily, the `PipeHandle` has been relocated into `kona-client`. All fault-proof-VM platform specific code will soon be isolated into its own crate.

closes #398 , but needs follow-ups to remove the `kona-client` library's dependency on `kona-common` as well.


clabby:
## 📚 $\text{Stack Overview}$

* #817 👈
* #816
* #813
* `main`

refcell (approved):
Nice, mr allen

PR #821, opened 2024-11-15
feat(workspace): Isolate FPVM-specific platform code

clabby:
## Overview

Isolates the FPVM-specific platform code into separate crates. This solves the issues that zkVM teams are having with the `kona-client` libs being coupled to the `kona-common` IO pattern.

This PR will be followed up with a new `kona-client` library, that just serves the entrypoint, generic over the communication `Channel`. From there, we can run `kona-client` in the host w/ `tokio` channels in the same process, and reuse the same logic for the entry point with a `kona-std-fpvm::FileChannel`


clabby:
## 📚 $\text{Stack Overview}$

* #821 👈
* `main`

bin/client/src/lib.rs
refcell, line 1: Can this be deleted?

clabby: Negative, see upstack.

Though if you'd prefer we could always refactor the code upstack into a new `kona-fault-proof` crate, and have the client bin literally just be the entrypoint.

refcell (approved):

PR #832, opened 2024-11-20
L2ExecutePayloadProof Hint Type

zobront:
In order to use Kona with reth, we need to get around the `debug_dbGet` calls.

Reth has recently implemented the `debug_executePayload` RPC method. This allows us to bulk get the entire witness (codes, states, and keys) for executing a given set of payloads against a prior state.

To minimally modify Kona to support this, this PR:
- Introduces the `L2ExecutePayloadProof` hint type
- Implements this hint in the fetcher to get all the data and add it all to the KV store

This will allow a program that uses Kona as a library to perform a bulk fetch using this hint type before execution. If it succeeds, all L2 state related hints (except the final output root calculation, which doesn't use `debug_dbGet`) will be supplied with an existing KV store value in `get_preimage()` and will never need to call `prefetch()`.

When Kona is upgraded to use this RPC method in the future, it can be integrated more deeply into the existing MPT library (and eventually, when fully trusted, can replace a lot of it). For now, this PR will give users the option to manually use this method to avoid further RPC calls.


bin/host/src/fetcher/mod.rs
```diff
@@ -507,6 +508,38 @@ where
...
+                    .l2_provider
+                    .client()
+                    .request::<(B256, Vec<u8>), ExecutionWitness>(
+                        "debug_executePayload",
+                        (parent_block_hash, payload_attributes),
+                    )
+                    .await
+                    .map_err(|e| anyhow!("Failed to fetch preimage: {e}"))?;
+
+                // TODO: validate what happens if multiple of the same key
```
clabby, line 528: This is fine. The values will be de-duplicated in favor of most recent insertion.

clabby (approved):
lgtm!

bin/host/src/fetcher/mod.rs
```diff
@@ -507,6 +508,38 @@ where
...
+                if hint_data.len() < 32 {
+                    anyhow::bail!("Invalid hint data length: {}", hint_data.len());
+                }
+                let parent_block_hash = B256::from_slice(&hint_data.as_ref()[..32]);
+                let payload_attributes = hint_data[32..].to_vec();
+
+                let execute_payload_response: ExecutionWitness = self
+                    .l2_provider
+                    .client()
+                    .request::<(B256, Vec<u8>), ExecutionWitness>(
```
clabby, line 521: has this been tested? Curious if it's expecting a `Vec<u8>` (json enc: `[0, 1, 2, 3, ...]`) instead of a `String`ified JSON (enc: `"{\"a\": 1, ...}"` for the req. Over in `reth` it's just deserializing the `Attributes` generic verbatim, which should be `OptimismPayloadAttributesV3` under the hood, but just sanity checking.

zobront: Waiting on some help from reth team to get tests working on this endpoint. Will test everything before merging — just wanted to get your feedback early in case there were more changes to make.

Thank you!

tynes:
Can we add the `L2ExecutePayloadProof` to the specs?

zobront:
@clabby I ran a test and it works as expected. Didn't leave the test in there because there aren't any public RPCs implementing this method yet. Let me know if there's another way you'd like to include the test - otherwise, this is good to merge.

refcell (approved):
LGTM pending @clabby's approval

clabby (approved):

PR #836, opened 2024-11-21
feat(host): Run client program natively in process

### LINKED ISSUE #553: feat(common): Improve native IO API

### LINKED ISSUE #815: feat(host): Execute `kona-client` in-process
# Overview

Currently, the `kona-host` binary shells out to invoke the `kona-client` binary, rather than executing it in-process. This was done towards the beginning in order to test the `kona-preimage` API natively in a similar way that the FPVMs use, but there are a few problems:
* The current [`BasicKernelInterface::write` & `BasicKernelInterface::read`](https://docs.rs/kona-common/latest/kona_common/trait.BasicKernelInterface.html) interfaces take an owned [`FileDescriptor`](https://docs.rs/kona-common/latest/kona_common/enum.FileDescriptor.html), which is inflexible for a few reasons:
    * It contains mappings that are invalid on the native architecture or non fault proof VMs (i.e. zkVMs). This is a poor abstraction for re-use.
    * On the `NativeIO` implementation, we have to take a `FileDescriptor` type, which is bad for reasons described in #553.
* The `kona-host` implementation therefore has to keep the pipe files for the client program alive, in order to guarantee that the pipe remains open for the duration of the asynchronous child process' lifetime.

Instead of doing this, we should change `kona-host` up so that `kona-client` runs natively within the process. This will remove the need for the `NativeIO` entirely in `kona-common`, and then we can split `kona-common` into client program utilities and FPVM-specific logic. This should remove the need to solve #553.

clabby:
## Overview

Runs the client program natively in the `kona-host` process rather than in a child process. This change is enabled by the preimage oracle modifications to decouple from the old `kona-common` crate, allowing for the host program to run with in-process channels and the client program to run with `kona-std-fpvm`'s `FileChannel`, per FPVM spec.

### on accelerated precompiles...

This change has also dug up some old tech debt around precompile acceleration. We need our EVM to be able to hold onto an `OracleReader<C: Channel>` and a `HintWriter<C: Channel>` in order to reach out to the preimage oracle for the result of certain precompiles' execution. However, the `HandlerRegister` function in `revm` only consumes a mutable reference to the `EvmHandler`, so we can't pull anything in that's non-static. There is a `BoxedHandlerRegister`, but means we need to capture the variables, and it ends up being a real mess with `Send` + `Sync` bounds throughout
the codebase.

As a follow up, we probably need to upstream some changes to `revm` that allow for precompiles to access the broader `Context`, so that we can place the host<->client communication handles within it for use within the precompiles.

### Meta

closes #553
closes #815


clabby:
## 📚 $\text{Stack Overview}$

* #836 👈
* #821
* `main`

crates/derive/src/attributes/stateful.rs
```diff
@@ -212,8 +212,7 @@ async fn derive_deposits(
         for l in r.logs.iter() {
             let curr_index = global_index;
             global_index += 1;
-            #[allow(clippy::all)]
-            if !l.data.topics().first().map_or(false, |i| *i == DEPOSIT_EVENT_ABI_HASH) {
+            if l.data.topics().first().is_none_or(|i| *i != DEPOSIT_EVENT_ABI_HASH) {
```
refcell, line 215: Thanks for fixing this - failed in nightly i think due to a bad clippy lint being introduced

refcell (approved):

Issue #839, opened 2024-11-22
feat: abstraction to allow customized EVMs


zobront:
Currently `kona-executor` exposes the revm handler in order to customize block execution. However, larger customizations of the EVM are not possible.

This abstraction already exists in reth/revm, where:
- reth implements `BasicBlockExecutor`, which contains a `strategy`.
- in order to `execute()` ([link](https://github.com/paradigmxyz/reth/blob/36db1c24077f7a500048c545c4c5cefb6a27224a/crates/evm/src/execute.rs#L344-L355)), it calls `apply_pre_execution_changes()`, `execute_transactions()`, `apply_post_execution_changes(), and `finish()` on the strategy.
- reth includes the [`OpExecutionStrategy`](https://github.com/paradigmxyz/reth/blob/main/crates/optimism/evm/src/execute.rs) which is a good example of this implementation.
- All strategies contain an `evm_config` which must implement [`ConfigureEvm`](https://github.com/paradigmxyz/reth/blob/36db1c24077f7a500048c545c4c5cefb6a27224a/crates/evm/src/lib.rs#L44).

This `evm_config` is what allows customization, as it defines three functions that replace some of the hardcoded actions in Kona:
- `fill_cfg_env()` replaces `self.evm_cfg_env()`
- `fill_block_env()` replaces `Self::prepare_block_env()`
- `fill_tx_env()` replaces `Self::prepare_tx_env()`

I believe the best option to accomplish this is to refactor Executor to rely on a struct implementing the `ConfigureEvm` trait in order to perform the above actions. For "normal" Kona, we can break the existing functions into a new type that implements the `ConfigureEvm` trait. For other customized chains like Odyssey, we can simply inherit whatever `ConfigureEvm` struct they are passing to reth for execution.

I want to talk through this and make sure it's the right strategy before touching any code, so let's discuss. Also available for a live call on Monday if that's easier.

clabby:
> I believe the best option to allow Kona more flexibility in execution is to refactor Executor to rely on a struct implementing the ConfigureEvm trait in order to perform the above actions. For "normal" Kona, we can break the existing functions into a new type that implements the ConfigureEvm trait. For other customized chains like Odyssey, we can simply inherit whatever ConfigureEvm struct they are passing to reth for execution.

[...]

PR #843, opened 2024-11-26
wip: better custom evm abstraction

zobront:
addresses #839 

to do:
- [x] redo Executor w `ConfigureEvmEnv` abstraction 
- [x] new ExecutorBuilder
- [x] `DefaultEvmConfig` impl of `ConfigureEvmEnv`
- [x] get rest of kona native compiling with these changes
- [ ] fix dependencies to ensure cannon & asterisc can compile
- [ ] [add CI checks on Reth side](https://github.com/paradigmxyz/reth/pull/12925) for all imported dependencies
- [ ] tests passing, addl test on customized chain


clabby:
I'm a fan of the direction of this. If the reth dependency is removed in favor of these traits being upstreamed in alloy, down to merge!

bin/client/src/lib.rs
```diff
@@ -116,7 +117,15 @@ where
         l1_provider.clone(),
         l2_provider.clone(),
     );
-    let executor = KonaExecutorConstructor::new(&cfg, l2_provider.clone(), l2_provider, None);
+    let executor = KonaExecutorConstructor::new(
+        &cfg,
+        DefaultEvmConfig::new(
+            // ZTODO: Use custom chainspec here, not defaulting to OP Mainnet
+            Arc::new(OpChainSpecBuilder::optimism_mainnet().build()),
+        ),
```
zobront, line 125: @clabby Is this client only for OP Mainnet (ie is this safe) or should it work for any chain that uses FPVM? If the latter, I'll construct ChainSpec out of data in the RollupConfig.

crates/executor/src/syscalls/eip4788.rs
```diff
@@ -5,60 +5,23 @@ use crate::{
...
-fn apply_beacon_root_contract_call<F, H>(
+pub(crate) fn apply_beacon_root_contract_call<F, H, C: ConfigureEvm>(
     config: &RollupConfig,
+    evm_config: &C,
     timestamp: u64,
     block_number: u64,
     parent_beacon_block_root: Option<B256>,
-    evm: &mut Evm<'_, (), &mut State<&mut TrieDB<F, H>>>,
+    // ZTODO: understand default external context better, is it ok to allow any here?
+    evm: &mut Evm<'_, C::DefaultExternalContext<'_>, &mut State<&mut TrieDB<F, H>>>,
```
zobront, line 24: @clabby I think this change is fine to allow the Evm to be parameterized by any DefaultExternalContext rather than forcing it to be `()`. Do you see any problem with that?

crates/executor/src/executor/mod.rs
```diff
@@ -166,10 +162,9 @@ where
             }
 
             // Modify the transaction environment with the current transaction.
-            evm = evm
-                .modify()
-                .with_tx_env(Self::prepare_tx_env(&transaction, raw_transaction)?)
-                .build();
+            let reth_tx = TransactionSigned::decode_2718(&mut raw_transaction).map_err(ExecutorError::RLPError)?;
```
zobront, line 165: @clabby This decoding is failing in the tests. Having trouble figuring out why. If you have any instincts, I'd love your input as I keep digging.

bin/client/src/lib.rs
clabby: It should work for any chain with a `RollupConfig`, yep!

crates/executor/src/syscalls/eip4788.rs
clabby: This is all good. Right now the enforcement on `()` limits a degree of freedom that consumers might want.

crates/executor/src/executor/mod.rs
clabby: Likely because some of these txs are deposits. I assume `TransactionSigned` may not support the deposit (0x7e) variant, given that no features are enabled for `reth-primitives`.

clabby:
Thanks for the effort here - looks to have spawned some new work from @mattsse et. co on something that will accomplish a similar goal. Going to go ahead and close this, but we'll revisit the issue (https://github.com/op-rs/kona/issues/839) soon after `ithacaxyz/evm` gets open-sourced.

PR #846, opened 2024-11-26
fix(`IndexedBlobHash`): Change `index` to `u64` 

ratankaliani:
On 64-bit architectures the behavior of `to_be_bytes` for `usize` and `u32` is the same, but in 32-bit architectures (e.g. SP1), `to_be_bytes` for a `usize` outputs 4 bytes instead of the expected 8 bytes.

This leads to the following error when getting blobs:

```shell
stderr: thread '<unnamed>' panicked at kona-d482f77e14858878/fd084b6/crates/proof-sdk/proof/src/l1/blob_provider.rs:41:31:
stderr: source slice length (4) does not match destination slice length (8)
stderr: stack backtrace:
stderr: note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
2024-11-26T19:26:01.928084Z  INFO execute: close time.busy=84.2s time.idle=20.0µs
thread 'main' panicked at /home/ubuntu/deployment-op-succinct/scripts/prove/src/lib.rs:44:10:
called `Result::unwrap()` on an `Err` value: execution failed with exit code 1
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Asterisc is a 64-bit architecture, which is why I think this issue wasn't caught: https://github.com/ethereum-optimism/asterisc.

Can we add CI for running a `kona` program in a 32-bit architecture? Happy to make a PR to ensure that consumers of `kona` (e.g. `op-succinct`) aren't impacted.

### DIFF
--- a/bin/host/src/fetcher/mod.rs
+++ b/bin/host/src/fetcher/mod.rs
@@ -187,7 +187,7 @@ where
                 let timestamp = u64::from_be_bytes(timestamp_data_bytes);
 
                 let partial_block_ref = BlockInfo { timestamp, ..Default::default() };
-                let indexed_hash = IndexedBlobHash { index: index as usize, hash };
+                let indexed_hash = IndexedBlobHash { index, hash };
 
                 // Fetch the blob sidecar from the blob provider.
                 let mut sidecars = self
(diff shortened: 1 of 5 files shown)


clabby (approved):
Thanks!

Yeah, let's add it to CI. Best way to do that is probably to add a platform matrix to the native run of the client / host program. Also glad to add a 32-bit target to the build job.

PR #850, opened 2024-11-27
feat(driver): Executor Interface

refcell:
### Description

Proofs is a specific use case for kona's driver.

This PR cleans up the executor interface that the kona driver uses such that the proofs use case of re-creating the stateless block executor is abstracted away.


refcell:
## 📚 $\text{Stack Overview}$

* #850 👈
* `main`

clabby (approved):

PR #852, opened 2024-11-27
feat(client): Invalidate impossibly old claims

clabby:
## Overview

Adds a check into the fault proof program entrypoint that invalidates any claims where the claimed L2 block number is < the safe L2 head's block number.

The short-circuit check for trace extension, if agreed == disputed, is preserved.


clabby:
## 📚 $\text{Stack Overview}$

* #852 👈
* `main`

refcell (approved):
Good catch

PR #853, opened 2024-11-27
fix: use non problematic hashmap fns

mattsse:
with alloy::map we need to be a bit careful with which hashmap functions we are using because some fns enforce certain hashers.

for example `From` is only implemented for `Randomstate`, which itself calls `from_iter` which only enforces ` S: BuildHasher + Default,`


refcell (approved):
thank you

PR #862, opened 2024-12-03
add fraemwork

bxue-l2:


refcell:
Kona is built so that trait abstractions provide a way to add new implementations through separate crates. The changes you've made here should go in a new crate, in a new repository that pulls in kona crates as dependencies. This way, new implementations don't need to be merged into kona, but can live outside of the kona repository while still building ontop of kona. See [hilo](https://github.com/anton-rs/hilo) for example, which pulls in `kona-derive` and `kona-driver` as dependencies, and builds off the trait abstractions in kona to use the derivation pipeline and driver.

bxue-l2:
Hi refcell, I intend it to upstream this change to my fork for others to review, but made a mistake to upstream here, sorry about it. Thanks so much for giving feedbacks. Will take a look at the example you gave.

refcell:
> Hi refcell, I intended to upstream this change to my fork for others to review, but made a mistake to upstream here, sorry about it. Thanks so much for giving feedbacks. Will take a look at the example you gave.

Oh no worries at all, if there is a change you'd like to make to any of the trait abstractions or internal pipeline/proofs logic please do open a pr here!!

samlaf:
@refcell are you picturing a world where every altda layer has their own hilo-like crate in their own repo? Or one where there's a single altda crate as part of anton-rs where all the altda providers would upstream their integration?

PR #868, opened 2024-12-11
feat(build): Adjust RV target - `riscv64g` -> `riscv64ima`

clabby:
## Overview

> [!NOTE]
> I validated locally that this target works out fine for running on asterisc, but will need to cut a release of the `asterisc-builder` cross compilation image for this to pass CI. Will do after vaca.
>
> For anyone who wants to experiment with this, head over to the `build` dir, and build the asterisc image locally w/ `just asterisc`. Then, head over to the root `justfile`, and adjust the `build-asterisc` recipe to use the local `asterisc-pipeline:latest` image that was built in the previous step.
>
> Once this is done, head over to `bin/client`, and run kona on asterisc w/ `just run-client-asterisc` to use the new `riscv64ima` binary.
>
> **e: A new image was manually published. This is good to go.**

Adjusts the build pipeline to use `riscv64imac-unknown-none-elf` as the target rather than `riscv64gc`. We still turn off `c`ompressed instructions, but this target forces LLVM to use soft-float operations for any `f`loating point / `d`oubleword floating point operations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ see the [SDK section of the book](https://anton-rs.github.io/kona/sdk/intro.html
 **Build Pipelines**
 
 - [`cannon`](./build/cannon): Docker image for compiling to the bare-metal `mips-unknown-none` target.
-- [`asterisc`](./build/asterisc): Docker image for compiling to the bare-metal `riscv64gc-unknown-none-elf` target.
+- [`asterisc`](./build/asterisc): Docker image for compiling to the bare-metal `riscv64imac-unknown-none-elf` target.
 
 **Protocol**
 - [`mpt`](./crates/mpt): Utilities for interacting with the Merkle Patricia Trie in the client program.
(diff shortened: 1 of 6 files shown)


zigtur:
I wonder if there is a way to exclude Zicsr instructions from the build.

[The LLVM docs are unclear](https://releases.llvm.org/17.0.1/docs/RISCVUsage.html#riscv-i2p1-note) about it:
> **zicntr, zicsr, zifencei, zihpm**
Between versions 2.0 and 2.1 of the base I specification, a backwards incompatible change was made to remove selected instructions and CSRs from the base ISA. These instructions were grouped into a set of new extensions, but were no longer required by the base ISA. This change is partially described in “Preface to Document Version 20190608-Base-Ratified” from the specification document (the zicntr and zihpm bits are not mentioned). LLVM currently implements version 2.1 of the base specification. To maintain compatibility, instructions from these extensions are accepted without being in the -march string. LLVM also allows the explicit specification of the extensions in an -march string.

It would make sense to also exclude these instructions as they are not supported by Asterisc.

PR #869, opened 2024-12-11
tooling: make client justfile's commands take an optional rollup_config_path

samlaf:
Needed to target a devnet which doesnt have a standard chain_id. Might need to make the same change for the offline client target as well?

We use this to test the client against the docker devnet in op's monorepo:
```
# Run the client program natively with the host program attached, against the op-devnet.
run-client-native-against-devnet verbosity='' block_number='' rollup_config_path='':
  #!/usr/bin/env bash
  L1_RPC="http://127.0.0.1:8545"
  L1_BEACON_RPC="http://127.0.0.1:5052"
  L2_RPC="http://127.0.0.1:9545"
  ROLLUP_NODE_RPC="http://127.0.0.1:7545"
  ROLLUP_CONFIG_PATH="../../../optimism/.devnet/rollup.json"

  if [ -z "{{block_number}}" ]; then
    BLOCK_NUMBER=$(cast block finalized --json --rpc-url $L2_RPC | jq -r .number | cast 2d)
  else
    BLOCK_NUMBER={{block_number}}
  fi

  just run-client-native $BLOCK_NUMBER \
    $L1_RPC $L1_BEACON_RPC $L2_RPC $ROLLUP_NODE_RPC \
    $ROLLUP_CONFIG_PATH {{verbosity}}
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,6 @@ monorepo/
 
 # Environment Variables
 .env
+
+# kona-host data-dir
+data/
--- a/bin/client/justfile
+++ b/bin/client/justfile
@@ -63,13 +63,21 @@ run-client-asterisc block_number l1_rpc l1_beacon_rpc l2_rpc rollup_node_rpc ver
     {{verbosity}}
 
 # Run the client program natively with the host program attached.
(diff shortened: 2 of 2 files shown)


clabby (approved):
Thanks :)

PR #875, opened 2024-12-20
chore: remove unused function in OnlineBlobProvider

samlaf:


rust-toolchain.toml
refcell, line 1: Can we not commit this and add it to the .gitignore?

refcell (changes requested):

rust-toolchain.toml
samlaf: Did you see my comment though?
> channel = "nightly" # needed so that vscode formats using rustmfmt.toml that uses nightly features
Otherwise how do you guys setup your vscode so that it formats using the rules you setup in rustfmt.toml?

clabby: Neither of us use vscode idt. I'm okay with adding this if it's needed for good devex for contributors who use vscode, though.

samlaf: Not going to lie haven’t been using rust for long enough to understand fully its whole toolchain. Adding this fixed my Va code formatting issue which I found weird because I thought specifying the toolchain in the vscode specific settings.json would make more sense.

Are you guys aware of the pros and cons of committing this toolchain file? If there are no known cons then I’d be in favor of committing it to make better devex for vscode devs. But otherwise I’m fine with removing it.

clabby: @samlaf The bad part about doing this is that we will always use nightly. This isn't often desired for normal builds.

You can work around this by just formatting using the CLI, i.e. `cargo +nightly fmt --all` or `just f`, rather than relying on VSCode's formatter hook.

clabby: @samlaf this might be a better solution: https://github.com/rust-lang/vscode-rust/issues/438#issuecomment-1003671382

samlaf: Nice, thanks. Removed the toolchain.toml file and added the vscode settings.json file instead: [04358b3](https://github.com/anton-rs/kona/pull/875/commits/04358b3c3c66a673ae1f320285f198311b6c6fdb).

There are still issues with this such that it doesn't work with vscode workspaces which is a bit annoying. I'm sure there's a way around this but don't feel like figuring it out now so I'll just add this image here for anyone else reading this in the future.
image

Also somehow feels like the toolchain.toml file should still be checked in to make for more reproducible environments, especially something like mentioning the MSRV (see https://github.com/anton-rs/kona/pull/867).

refcell (approved):
Good catch actually - the host doesn't even need the logic provided by the `BlobProvider` trait so this can be removed.

PR #876, opened 2024-12-22
fix(derive): `BatchStream` Past batch handling

clabby:
## Overview

Fixes the handling of the `BatchValidity::Past` condition. It's currently throwing a temporary `Eof`, but the desired behavior is to continue reading old batches until all past batches have been read, validated, and discarded due to their timestamp.


Inphi:
is/will there be an action test covering this case?

clabby:
> is/will there be an action test covering this case?

Not yet, but set a reminder to get one in once I'm back to work.

refcell (approved):
Good catch thanks

PR #881, opened 2025-01-03
chore: Isthmus Withdrawals Root

refcell:
### Description

Updates kona's stateless block executor to set the withdrawals root to the L2 to L1 message passer account when isthmus is active.

Specs: https://specs.optimism.io/protocol/isthmus/exec-engine.html#l2tol1messagepasser-storage-root-in-header

### DIFF
--- a/crates/executor/src/executor/mod.rs
+++ b/crates/executor/src/executor/mod.rs
@@ -59,6 +59,20 @@ where
         StatelessL2BlockExecutorBuilder::new(config, provider, hinter)
     }
 
+    /// Fetches the L2 to L1 message passer account from the cache or underlying trie.
+    fn message_passer_account(&mut self) -> Result<B256, TrieDBError> {
+        match self.trie_db.storage_roots().get(&L2_TO_L1_BRIDGE) {
+            Some(storage_root) => {
+                storage_root.blinded_commitment().ok_or(TrieDBError::RootNotBlinded)
+            }
+            None => Ok(self
+                .trie_db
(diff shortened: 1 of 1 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #881 👈
* `main`

crates/executor/src/executor/mod.rs
```diff
@@ -103,6 +117,18 @@ where
...
+        // root hash.
+        let mut withdrawals_root = self
+            .config
+            .is_canyon_active(payload.payload_attributes.timestamp)
+            .then_some(EMPTY_ROOT_HASH);
+        // If the Isthmus hardfork is active, the withdrawals root is the L2 to L1 message passer
+        // account.
+        if self.config.is_isthmus_active(payload.payload_attributes.timestamp) {
+            withdrawals_root = Some(self.message_passer_account()?);
+        }
```
clabby, line 130: This always needs to go after the state root is sealed up, post-execution. Code looks good though, just need to move it back to where it used to be.

clabby (changes requested):

crates/executor/src/executor/mod.rs
refcell: The issue is an exclusive mutable reference to the trie db is given to the `State` through the builder earlier in the method (right below this added section) - see `&mut self.trie_db` which is needed to get the message passer account for the withdrawals root.

clabby (approved):

PR #884, opened 2025-01-06
chore: Add emhane as a CODEOWNER

refcell:
### Description

Updates the github codeowners to include @emhane


emhane (approved):
ci should will pass after https://github.com/op-rs/kona/pull/883 is released

clabby (approved):

PR #886, opened 2025-01-07
chore: Enable EIP-7702 txs in Derivation Pipeline

refcell:
### Description

Enables derivation over EIP-7702 transactions.

Notice, the fault proof program still explicitly _disallows_ EIP-7702 transactions on L2. See [here](https://github.com/op-rs/kona/blob/992b1a15733c3b30dfe9e6a0d9d86c5e3ca133b3/crates/executor/src/executor/mod.rs#L237-L240).

### DIFF
--- a/crates/derive/src/sources/calldata.rs
+++ b/crates/derive/src/sources/calldata.rs
@@ -51,6 +51,7 @@ impl<CP: ChainProvider + Send> CalldataSource<CP> {
                     TxEnvelope::Legacy(tx) => (tx.tx().to(), tx.tx().input()),
                     TxEnvelope::Eip2930(tx) => (tx.tx().to(), tx.tx().input()),
                     TxEnvelope::Eip1559(tx) => (tx.tx().to(), tx.tx().input()),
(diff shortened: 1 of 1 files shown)


crates/derive/src/sources/calldata.rs
```diff
@@ -51,6 +51,7 @@ impl<CP: ChainProvider + Send> CalldataSource<CP> {
                     TxEnvelope::Legacy(tx) => (tx.tx().to(), tx.tx().input()),
                     TxEnvelope::Eip2930(tx) => (tx.tx().to(), tx.tx().input()),
                     TxEnvelope::Eip1559(tx) => (tx.tx().to(), tx.tx().input()),
+                    TxEnvelope::Eip7702(tx) => (tx.tx().to(), tx.tx().input()),
```
clabby, line 54: Super important that we never consider these txs just yet. We just want the ability to observe 7702 txs on L1, but skip over them, without the pipeline crashing due to the tx type being unknown.

PR #888, opened 2025-01-08
feat(driver): Multi-block derivation

### LINKED ISSUE #795: feat(driver): Multi-block derivation
### Description

`DerivationDriver::advance_to_target` currently only derives a single L2 block and exits even if the `BootInfo` points to a claimed output that is more than a single block away from the agreed l2 head.

Enabling multi-block derivation would allow non-interactive fault proofs that cover an entire proposal while remaining compatible with the superchain.

hashcashier:
This PR adds an optional `PipelineCursor` instance into `OracleL2ChainProvider` which enables the forward movement of the reference L2 block hash used to verifiably query L2 block data by number.

Using this cursor in the provider enables providing authenticated data to the pipeline beyond what could previously be validating only using the commitments in `BootInfo`.

This allows `Driver::advance_to_target` to iteratively process multiple sequential blocks, extending the usability of Kona in ZK fault proofs.

The PR also reduces some boilerplate by introducing convenience methods in `HintType` and `OracleL2ChainProvider`.

Closes #795


bin/client/src/lib.rs
```diff
@@ -147,33 +146,3 @@ where
...
-        )
-        .await
-        .map_err(OracleProviderError::Preimage)?;
-
-    let safe_hash =
-        output_preimage[96..128].try_into().map_err(OracleProviderError::SliceConversion)?;
-    l2_chain_provider
-        .header_by_hash(safe_hash)
-        .map(|header| Sealed::new_unchecked(header, safe_hash))
-}
```
refcell, line 179: This de-duplication is very nice

crates/proof-sdk/proof/src/sync.rs
```diff
@@ -38,5 +40,7 @@ where
     let mut cursor = PipelineCursor::new(channel_timeout, origin);
     let tip = TipCursor::new(safe_head_info, safe_header, boot_info.agreed_l2_output_root);
     cursor.advance(origin, tip);
-    Ok(cursor)
+
+    // Wrap the cursor in a shared read-write lock
+    Ok(Arc::new(RwLock::new(cursor)))
```
refcell, line 45: Is there a reason you chose to return the `Arc<RwLock<PipelineCursor>>` instead of doing this when it's consumed?

refcell (approved):
I support this - one pedantic nit is holding the RwLock across multiple asynchronous operations, but since it will not deadlock, it's acceptable.

crates/proof-sdk/proof/src/sync.rs
hashcashier: There's just one usage of the method in client so I thought not to clog up the client's `run` code.

hashcashier:
@clabby I've resolved the conflicts with the recent refactor

clabby:
> @clabby I've resolved the conflicts with the recent refactor

Amazing. Getting in #910, then will merge this today!

PR #897, opened 2025-01-12
Move temporary error logs to lower level than WARN

### LINKED ISSUE #896: chore(derive): Remove redundant warning logs
## Overview

Right now, there are some redundant warning logs that get emitted mentioning that the derivation step fails. These are a bit misleading, since a `Temporary` error kind is basically a signal to block the pipeline stage until new data arrives.

Image

We should move `Temporary` errors here to `debug!` traces, and still `warn!` / `error!` on `Critical` errors.

sergerad:
Closes #896

Refactors logging around some `PipelineErrorKind::TemporaryError`s to be lower level than WARN.

### DIFF
--- a/crates/derive/src/pipeline/core.rs
+++ b/crates/derive/src/pipeline/core.rs
@@ -177,7 +177,7 @@ where
                     StepResult::AdvancedOrigin
                 }
                 PipelineErrorKind::Temporary(_) => {
-                    debug!(target: "pipeline", "Attributes queue step failed temporarily: {:?}", err);
+                    trace!(target: "pipeline", "Attributes queue step failed temporarily: {:?}", err);
                     StepResult::StepFailed(err)
                 }
                 _ => {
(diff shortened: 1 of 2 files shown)


sergerad:
@clabby My changes have captured the lines that you have shared in image attached to the issue.

Just wanted to check whether we wanted to handle the general case of `::Temporary(_)` with a trace log:
https://github.com/op-rs/kona/blob/24e73bae9e87586420d5ff399fe11a47bae8bd58/crates/derive/src/pipeline/core.rs#L108

And whether we want to consider any other temporary errors here:
https://github.com/op-rs/kona/blob/24e73bae9e87586420d5ff399fe11a47bae8bd58/crates/derive/src/stages/batch/batch_queue.rs#L356

and

https://github.com/op-rs/kona/blob/24e73bae9e87586420d5ff399fe11a47bae8bd58/crates/derive/src/stages/channel/channel_bank.rs#L201

clabby:
> @clabby My changes have captured the lines that you have shared in image attached to the issue.
> 
> Just wanted to check whether we wanted to handle the general case of `::Temporary(_)` with a trace log:
> 
> [...]

Yeah, this looks good to me. Thank you for the change!

clabby (approved):

PR #899, opened 2025-01-13
feat(workspace): `kona-interop` crate

clabby:
## Overview

Adds the initial version of the `kona-interop` crate, containing core types for interop (messages, `SuperRoot`, message graph) as well as core functionality for extracting messages from logs + resolving a set of interdependent blocks' validity per [message validity rules](https://specs.optimism.io/interop/messaging.html#invalid-messages).

This crate should eventually live over in `maili`, but it's subject to change while the initial interop proof is being laid out, so it's easier for it to live in `kona` for the first phase of its life.

### DIFF
--- a/crates/interop/src/message.rs
+++ b/crates/interop/src/message.rs
@@ -1,0 +1,114 @@
+//! Interop message primitives.
+//!
+//! <https://specs.optimism.io/interop/messaging.html#messaging>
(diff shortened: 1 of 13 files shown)


crates/interop/src/message.rs
refcell, line 1: These we should have in `maili-protocol`, e.g. [`MessageIdentifier`](https://docs.rs/maili-protocol/0.1.4/maili_protocol/struct.MessageIdentifier.html)

PR #902, opened 2025-01-14
feat(workspace): `kona-proof-interop` crate

clabby:
## Overview

Adds a new extension to the `kona-proof` crate that supports the interop proof program.

PR #903, opened 2025-01-14
feat(client): Interop binary

clabby:
## Overview

Adds a new entrypoint to `kona-client` for the interop binary


clabby:
## 📚 $\text{Stack Overview}$

* #904
* #903 👈
* #902
* `main`

crates/proof-sdk/proof/src/sync.rs
```diff
@@ -26,17 +28,16 @@ where
...
-        l1_origin_number = boot_info.rollup_config.genesis.l1.number;
+    if l1_origin_number < rollup_config.genesis.l1.number {
+        l1_origin_number = rollup_config.genesis.l1.number;
     }
     let origin = chain_provider.block_info_by_number(l1_origin_number).await?;
 
     // Construct the cursor.
     let mut cursor = PipelineCursor::new(channel_timeout, origin);
-    let tip = TipCursor::new(safe_head_info, safe_header, boot_info.agreed_l2_output_root);
+    let tip = TipCursor::new(safe_head_info, safe_header, B256::ZERO);
```
clabby, line 40: note to reader - this type is pretty weird. We should refactor this at some point - it's mainly a way to retrieve the OR after the block was computed, but nothing that uses it depends on a starting output root at all.

refcell (approved):
Different binary entrypoints look good.

PR #904, opened 2025-01-14
feat(host): Support multiple modes

clabby:
## Overview

Reworks the host program to support multiple modes via subcommands. This will be useful for the interop proof + maintaining the legacy single-chain proof.


clabby:
## 📚 $\text{Stack Overview}$

* #904 👈
* #903
* #902
* `main`

bin/host/src/single/cli.rs
```diff
@@ -0,0 +1,282 @@
...
+use tokio::sync::RwLock;
+use tracing::error;
+
+/// The host binary CLI application arguments.
+#[derive(Default, Parser, Serialize, Clone, Debug)]
+#[command(styles = cli_styles())]
+pub struct SingleChainHostCli {
+    /// Verbosity level (0-2)
+    #[arg(long, short, action = ArgAction::Count)]
+    pub v: u8,
```
refcell, line 31: Doesn't the host cli itself have this field? Is this so the verbosity positioning is graceful?

refcell (approved):

bin/host/src/single/cli.rs
clabby: good catch - forgot to remove

Issue #849, opened 2024-11-27
chore(ci): Build and test on 32-bit platforms


clabby:
## Overview

Users of kona run on both 32-bit and 64-bit platforms, and there's currently no job in kona's CI to build and test on 32-bit platforms. This can cause us to easily break the build for downstream consumers without knowing.

We should add a 32-bit linux platform target for the client / host job

https://github.com/anton-rs/kona/blob/89ed2099547a65f66ff87bed3120d1ce25d4d993/.github/workflows/client_host.yaml#L10-L75

and also add a build job for a 32-bit linux patform in

https://github.com/anton-rs/kona/blob/89ed2099547a65f66ff87bed3120d1ce25d4d993/.github/workflows/rust_ci.yaml#L81-L111

emhane:
blocked by https://github.com/actions/checkout/issues/1681, ref https://github.com/op-rs/kona/pull/907#issuecomment-2612821340

refcell:
Initial stab taken in https://github.com/op-rs/kona/pull/907

PR #907, opened 2025-01-15
chore(ci): Run client + host workflow on 32 bit platform

emhane:
Closes https://github.com/op-rs/kona/issues/849


emhane:
seems like no runner wants to pick up this job...
```
Requested labels: linux-ubuntu-latest-32bit
Job defined at: op-rs/kona/.github/workflows/client_host.yaml@refs/pull/907/merge
Waiting for a runner to pick up this job...
Requested labels: linux-ubuntu-latest-32bit
Job defined at: op-rs/kona/.github/workflows/client_host.yaml@refs/pull/907/merge
Waiting for a runner to pick up this job...
```

emhane:
not sure why the syntax isn't being parsed, feel free to take over
```
The workflow is not valid. .github/workflows/client_host.yaml (Line: 18, Col: 9): Unrecognized named-value: 'matrix'. Located at position 1 within expression: matrix.isa == '32'
```

refcell:
It appears this won't work because github actions checkout steps require a node version only available in 64 bit builds.

emhane:
ah ok, I was looking at this https://github.com/yuki-kimoto/SPVM-Regex/blob/master/.github/workflows/linux-ubuntu-latest-32bit.yml @refcell

clabby:
It looks like @refcell's fix does the trick, but the image we're using does not have all of the tools that are preloaded into GitHub's [`runner-images`](https://github.com/actions/runner-images). To complete this I think we'd need to add an extra step to install all required dependencies.

emhane:
ah I thought rustc was clever enough to infer the target in this case. is that what you mean? @clabby

clabby:
> ah I thought rustc was clever enough to infer the target in this case. is that what you mean? @clabby

Ah nope - this isn't to do with rustc, the image that the actual runner is executing the job on doesn't have the tools that are required. GitHub's runner images come pre-loaded with a bunch of stuff (gnu utils, docker, nodejs, etc.) that this new i386 image doesn't have.

emhane:
@janjakubnanista had some insights on this, it could be a GitHub Actions bug :eyes:

janjakubnanista:
@emhane indeed, this is the root cause I believe https://github.com/actions/checkout/issues/1681

refcell:
Going to ice this for now and reference in the issue.

Feel free to re-open as desired to work on it 😄

PR #913, opened 2025-01-19
feat(client): Interop consolidation sub-problem

clabby:
## Overview

Implements the happy-path variant of the consolidation sub-problem for the interop FPP.

This will be followed up with a larger overhaul that allows for re-executing the optimistic blocks with deposits only, in the case where dependencies were found to be invalid.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2593,6 +2593,7 @@ dependencies = [
  "rand",
  "serde",
  "serde_json",
+ "spin",
  "tracing",
 ]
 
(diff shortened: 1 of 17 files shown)


refcell (approved):
Still a few todos - also beware i saw a mix of `INTEROP: ` and `TODO: ` comment prefixes used over your interop prs. Those will need to be fixed.

clabby:
Thanks @refcell - those todos are left there intentionally. I'll comb through them this morning and translate to tickets if I'm not able to get to them today.

PR #915, opened 2025-01-22
feat(host): Modular components

clabby:
## Overview

Cleans up some of the duplication in the host by tackling https://github.com/op-rs/kona/issues/369. The host's preimage server is now served with generics over the `Fetcher` + communication `Channel`. The `HostOrchestrator` trait allows for creating a custom host program with its own CLI with most of the bootstrapping auto-implemented.

These abstractions are also used in the `kona-host` binary now, sharing a ton more code between the single-chain and interop modes.


clabby:
## 📚 $\text{Stack Overview}$

* #915 👈
* `main`

bin/host/src/interop/orchestrator.rs
```diff
@@ -0,0 +1,105 @@
...
+impl HostOrchestrator for InteropHostCli {
+    type Providers = InteropProviders;
+
+    async fn create_providers(&self) -> Result<Option<Self::Providers>> {
+        if self.is_offline() {
+            return Ok(None);
+        }
+
+        let l1_provider =
+            http_provider(self.l1_node_address.as_ref().ok_or(anyhow!("Provider must be set"))?);
```
emhane, line 37: lets define a proper error enum for the errors thrown in this impl, and deriving `thiserror::Error` for it, has no_std support https://github.com/dtolnay/thiserror/blob/0f532e326e9a4cc6c6e30ee19ab00cb9eeb44362/Cargo.toml#L17-L28

bin/host/src/main.rs
```diff
@@ -1,21 +1,31 @@
...
+
 #[tokio::main(flavor = "multi_thread")]
 async fn main() -> Result<()> {
     let cfg = HostCli::parse();
     init_tracing_subscriber(cfg.v)?;
 
     match cfg.mode {
         HostMode::Single(cfg) => {
-            cfg.run().await?;
+            cfg.start().await?;
```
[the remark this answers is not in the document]
clabby: np, can change back.

bin/host/src/orchestrator.rs
```diff
@@ -0,0 +1,132 @@
...
+    ) -> Option<Arc<RwLock<impl Fetcher + Send + Sync + 'static>>>;
+
+    /// Runs the client program natively and returns the exit code.
+    async fn run_client_native(
+        hint_reader: HintWriter<NativeChannel>,
+        oracle_reader: OracleReader<NativeChannel>,
+    ) -> Result<()>;
+
+    /// Starts the host and client program in-process.
+    async fn start(&self) -> Result<()> {
```
refcell, line 66: This doesn't really just "start" the host and client though - it persistently runs them

bin/host/src/orchestrator.rs
```diff
@@ -0,0 +1,132 @@
...
+        &self,
+        providers: Option<Self::Providers>,
+        kv_store: SharedKeyValueStore,
+    ) -> Option<Arc<RwLock<impl Fetcher + Send + Sync + 'static>>>;
+
+    /// Runs the client program natively and returns the exit code.
+    async fn run_client_native(
+        hint_reader: HintWriter<NativeChannel>,
+        oracle_reader: OracleReader<NativeChannel>,
+    ) -> Result<()>;
```
refcell, line 63: There must be a cleaner way to abstract this

bin/host/src/interop/orchestrator.rs
clabby: Glad to do - though the entire host currently uses `anyhow`. I'll create a ticket for nuking `anyhow` from the codebase, the host is the last step.

clabby: https://github.com/op-rs/kona/issues/916

bin/host/src/orchestrator.rs
clabby: What would you suggest?

The idea here is to allow for customization of the host, and these are the main configuration params. The providers can be different depending on the host, the key value store can be customized to allow for certain local keys / pre-loaded data (useful for zkVMs that want to customize witness collection & our interop/single modes that have unique local KV stores), and the fetcher can be different to allow handling a custom set of hints.

refcell (approved):

bin/host/src/orchestrator.rs
clabby: Talked offline - we'll comb over these as we keep working on the interop host, and see if this causes any pain. If it does, will revisit.

Issue #949, opened 2025-01-24
feat(executor): EIP-7685 no-op


clabby:
## Overview

L1 pectra introduces [EIP-7685](https://github.com/ethereum/EIPs/blob/7f01d7b262fd2acda7ccb9a7b02aa1914d42e649/EIPS/eip-7685.md?plain=1#L6), which adds a general purpose message bus to the block header. Following isthmus activation, this value will be set to the `EMPTY_ROOT_HASH`.

https://github.com/op-rs/kona/blob/4b00849696c8dffe07272505a99c93346bc998a6/crates/executor/src/executor/mod.rs#L340

PR #951, opened 2025-01-24
feat: use empty requests hash when isthmus enabled

meyer9:
If Isthmus is enabled, pass empty SHA-256 hash as the requests root. Checked that this matches op-geth when passing in an empty array.

Wasn't totally sure where to put the constant for empty sha-256, but let me know if I should move it.

Spec: https://eips.ethereum.org/EIPS/eip-7685#block-header

Fixes https://github.com/op-rs/kona/issues/949


clabby (approved):
Looks good to me. Thanks @meyer9! I'll handle moving the constant soon, not super important.

PR #957, opened 2025-01-27
feat(kona-host): export `HostCli`, fix compilation

ratankaliani:
The modules defined in `main.rs` are not accessible outside the crate because `main.rs` is the binary entry point, not a library file. By default, code in `main.rs` is only available within the binary itself.

To make the CLI types (e.g. `HostCli`) available externally (e.g. in https://github.com/succinctlabs/op-succinct), they need to be:

- Defined in `lib.rs` (or a module exported by lib.rs).
- Marked with pub visibility.

Once I included exposed the modules in `lib.rs`, modules within `cli`, `single`, etc., need to import types from `kona-host` with `crate::`.

### DIFF
--- a/bin/host/src/lib.rs
+++ b/bin/host/src/lib.rs
@@ -19,3 +19,8 @@ pub use preimage::{
 
 mod server;
 pub use server::PreimageServer;
+
+pub mod cli;
+pub mod eth;
+pub mod interop;
+pub mod single;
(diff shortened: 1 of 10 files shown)


bin/host/src/lib.rs
```diff
@@ -19,3 +19,8 @@ pub use preimage::{
 
 mod server;
 pub use server::PreimageServer;
+
+pub mod cli;
+pub mod eth;
+pub mod interop;
+pub mod single;
```
ratankaliani, line 26: If we want to restrict the accessible types from these crates, happy to apply the `pub use` pattern above.

ratankaliani:
@clabby Would it be possible to export `SingleChainHostCli` from `kona-host`? Or was this the intended behavior for it to only be used by binaries in `kona-host`?

PR #958, opened 2025-01-28
feat: fill eip 7702 tx env with auth list

### LINKED ISSUE #948: feat(executor): Allow EIP-7702 transactions after Isthmus
## Overview

Right now, the executor explicitly disallows EIP-7702 transactions. With Isthmus, we'll need to conditionally activate them.

Transaction env filling:
https://github.com/op-rs/kona/blob/4b00849696c8dffe07272505a99c93346bc998a6/crates/executor/src/executor/env.rs#L238

Remove panic:
https://github.com/op-rs/kona/blob/4b00849696c8dffe07272505a99c93346bc998a6/crates/executor/src/executor/mod.rs#L237-L240

meyer9:
Fills tx env for EIP 7702 transactions using the same format as EIP 1559, but with authorization list added.

Fixes #948 

TODO: do we need a switch for Isthmus here? Or is it OK to be active all the time - guessing the batcher shouldn't include 7702 txs until Isthmus is active?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2070,15 +2070,15 @@ dependencies = [
 
 [[package]]
 name = "httparse"
-version = "1.9.5"
+version = "1.10.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7d71d3574edd2771538b901e6549113b4006ece66150fb69c0fb6d9a2adae946"
+checksum = "f2d708df4e7140240a16cd6ab0ab65c972d7433ab77819ea693fde9c43811e2a"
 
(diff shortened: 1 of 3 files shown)


emhane (approved):
nice. makes most sense to have a switch to check if isthmus is active in scope of caller of `prepare_tx_env`.

refcell:
I've updated the calling context of `prepare_tx_env` to prevent EIP-7702 txs if isthmus is _not_ active as previous behavior.

refcell (approved):

meyer9:
awesome, thanks!

PR #960, opened 2025-01-28
feat(client): EIP-2537 BLS12-381 Curve Precompile Acceleration

### LINKED ISSUE #946: feat(client): Accelerate BLS12-381 precompile
## Overview

The BLS12-381 precompile is being added in L1's pectra hardfork, and inherited by isthmus. It's an expensive cryptographic precompile, which means we'll want to accelerate it within the fault prover.

These are registered in the client program [here](https://github.com/op-rs/kona/blob/main/bin/client/src/precompiles/mod.rs), and the BLS precompile will also need to be added into the host [here](https://github.com/op-rs/kona/blob/main/bin/host/src/eth/precompiles.rs#L9-L15).

We'll also need to be cognizant of the ability to reproduce the precompile call on L1 via the `PreimageOracle` contract's [`loadPrecompilePreimagePart`](https://github.com/ethereum-optimism/optimism/blob/82784fad99be3cca81fe81ee7c9e3c678abd2c37/packages/contracts-bedrock/src/cannon/PreimageOracle.sol#L403) function. The BLS12-381 precompile [accepts variable-length inputs, so it'll have variable gas](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2537.md#gas-schedule-clarifications-for-the-variable-length-input), and we'll need to cap this within both the client software and the proof. We did [the same thing for the `ecPairing` precompile in the Granite hardfork](https://specs.optimism.io/protocol/granite/exec-engine.html#bn256pairing-precompile-input-restriction), since it's also accelerated in the proof + has dynamic gas consumption based on the input.

refcell:
### Description

Adds support for the [EIP-2537](https://eips.ethereum.org/EIPS/eip-2537) BLS12-381 curve precompile using FPVM acceleration in kona.

Closes #946


refcell:
## 📚 $\text{Stack Overview}$

* #960 👈
* `main`

bin/client/src/precompiles/bls12.rs
```diff
@@ -0,0 +1,108 @@
...
+
+    #[test]
+    fn test_fpvm_bls12_out_of_gas() {
+        let input = Bytes::from(vec![0u8; INPUT_LENGTH * 2]);
+        let gas_limit = PAIRING_MULTIPLIER_BASE - 1;
+        assert_eq!(
+            fpvm_bls12_pairing(&input, gas_limit),
+            Err(PrecompileError::OutOfGas.into())
+        );
+    }
```
emhane, line 107: can we get tests for all the possible errors returned from `fpvm_bls12_pairing`?

emhane (commented):
looking good!

bin/client/src/precompiles/bls12.rs
refcell: Added another test for the offset. Note that other precompiles do not have tests. Can you write up a ticket for adding tests to those precompiles?

emhane (approved):
amazing

bin/client/src/precompiles/bls12.rs
```diff
@@ -0,0 +1,115 @@
...
+/// Offset gas fee for BLS12-381 pairing operation.
+const PAIRING_OFFSET_BASE: u64 = 37700;
+
+/// The address of the BLS12-381 pairing precompile.
+pub(crate) const FPVM_BLS12_PAIRING: PrecompileWithAddress =
+    PrecompileWithAddress(BLS12_PAIRING_CHECK, Precompile::Standard(fpvm_bls12_pairing));
+
+/// Performs an FPVM-accelerated BLS12-381 pairing check.
+fn fpvm_bls12_pairing(input: &Bytes, gas_limit: u64) -> PrecompileResult {
+    let input_len = input.len();
```
clabby, line 40: We need to make a diff to the L1 precompile that limits the amount of gas it can consume, similar to what we did in granite for bn128 pairing. Have we spec'd this yet?

clabby: e: i see the discussion over in https://github.com/ethereum-optimism/specs/pull/553/files#r1931103051, looks like the right direction. This is good to go once that limit has been put in place in both the accelerated precompile and op-revm.

emhane: is there a way to catch this in a test?

refcell: Oh good callout - this is not spec'd out yet thanks

bin/client/src/precompiles/bls12.rs
emhane: believe @meyer9 has been testing this elsewhere?

meyer9: Yeah I copied some test cases from here for the end-to-end tests in the Optimism monorepo: https://eips.ethereum.org/assets/eip-2537/test-vectors

I would think that revm should include the test vectors from here as part of the tests for those precompiles, but I don't see them. What do you think about adding these to revm?

emhane: that's a great idea! do you want to open an issue in revm? think it makes sense if you open it since you have the most context here

meyer9: sure, will do!

meyer9: https://github.com/bluealloy/revm/issues/2034

meyer9: ah nvm, those tests are already implemented as part of the `execution-spec-tests` repo. https://github.com/bluealloy/revm/blob/4c773541ddf5a4966a512b4f0dde3da617ba9962/tests/pectra_devnet5/state_tests/prague/eip2537_bls_12_381_precompiles/bls12_g1msm/valid.json#L1

bin/client/src/precompiles/mod.rs
```diff
@@ -43,6 +44,10 @@ pub(crate) fn fpvm_handle_register<F, H>(
             ctx_precompiles.extend([bn128_pair::FPVM_ECPAIRING_GRANITE]);
         }
 
+        if spec_id.is_enabled_in(SpecId::ISTANBUL) {
```
clabby, line 47:
```suggestion
        if spec_id.is_enabled_in(SpecId::ISTHMUS) {
```

clabby (approved):
Nice. Will also want to land this limit in `revm`.

refcell:
> Nice. Will also want to land this limit in `revm`.

Will handle this

refcell:
> > Nice. Will also want to land this limit in `revm`.
> 
> Will handle this

https://github.com/bluealloy/revm/pull/2038

Issue #950, opened 2025-01-24
feat(executor): Pectra system calls


clabby:
## Overview

Pectra adds 3 new system calls, similar to EIP-4788:
* EIP-2935
* EIP-7002
* EIP-7251

Like the pre-block beacon root call, we'll need to support all 3 of these in `kona`'s executor, enabled by Isthmus activation.

https://github.com/op-rs/kona/blob/4b00849696c8dffe07272505a99c93346bc998a6/crates/executor/src/executor/mod.rs#L122-L130

Reference implementations over in reth:
* [`EIP-2935` system call](https://github.com/paradigmxyz/reth/blob/203fed0f6455c0afb5bd989992299f1002ec73e9/crates/evm/src/system_calls/eip2935.rs)
* [`EIP-7002` system call](https://github.com/paradigmxyz/reth/blob/203fed0f6455c0afb5bd989992299f1002ec73e9/crates/evm/src/system_calls/eip7002.rs)
* [`EIP-7251` system call](https://github.com/paradigmxyz/reth/blob/203fed0f6455c0afb5bd989992299f1002ec73e9/crates/evm/src/system_calls/eip7251.rs)

PR #963, opened 2025-01-28
feat(executor): EIP-2935 Syscall Support [ISTHMUS]

refcell:
### Description

The `Isthmus` hardfork introduces the [EIP-2935](https://eips.ethereum.org/EIPS/eip-2935) contract to serve historical block hashes from state.

This PR adds support to kona's executor to syscall the contract.

Makes progress on #950

### DIFF
--- a/crates/executor/src/syscalls/eip2935.rs
+++ b/crates/executor/src/syscalls/eip2935.rs
@@ -1,0 +1,183 @@
+//! Contains the [EIP-2935][eip-2935] syscall.
+//!
+//! [eip-2935]: https://eips.ethereum.org/EIPS/eip-2935
+
+use crate::{
+    db::TrieDB,
+    errors::{ExecutorError, ExecutorResult},
+    TrieDBProvider,
+};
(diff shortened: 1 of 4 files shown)


crates/executor/src/syscalls/eip2935.rs
```diff
@@ -0,0 +1,183 @@
...
+/// Fill transaction environment with the system caller and the system contract address and message
+/// data.
+///
+/// This is a system operation and therefore:
+///  * the call must execute to completion
+///  * the call does not count against the block’s gas limit
+///  * the call does not follow the EIP-1559 burn semantics - no value should be transferred as part
+///    of the call
+///  * if no code exists at the provided address, the call will fail silently
+fn fill_tx_env_with_system_contract_call(
```
clabby, line 141: can this be shared with EIP-4788 + EIP-7251? This code should already exist in `eip4788.rs`.

clabby (approved):
lgtm, can de-dup the TxEnv filling logic between these new system calls though.

PR #973, opened 2025-01-29
fix(client): Interop transition rules

clabby:
## Overview

Small fix to the interop transition step, where:
1. No-op transitions were wrongfully rejected.
2. The validity of the claimed timestamp wasn't checked.


bin/client/src/interop/transition.rs
```diff
@@ -160,7 +162,8 @@ where
 fn transition_and_check(
     pre_state: PreState,
     optimistic_block: Option<OptimisticBlock>,
-    expected: B256,
+    expected_post_state: B256,
+    timestamps: Option<(u64, u64)>,
```
refcell, line 166: Is there a reason to make timestamp optional?

refcell (approved):

PR #974, opened 2025-01-29
fix(executor): Withdrawals root

clabby:
## Overview

Fixes the withdrawals root feature of Isthmus. It's not currently enabled, but there was also a logic flaw where it fetched the L2ToL1MessagePasser's storage root for the parent block, not the current block.


mattsse (commented):
btw

the trieaccount looks identical to 

https://github.com/alloy-rs/trie/blob/main/src/account.rs

clabby:
> btw
> 
> the trieaccount looks identical to
> 
> https://github.com/alloy-rs/trie/blob/main/src/account.rs

Nice. We'll de-dup that.

clabby:
@mattsse https://github.com/op-rs/kona/issues/975

refcell (approved):

PR #977, opened 2025-01-29
chore(executor): De-duplicate `TrieAccount` type

### LINKED ISSUE #975: chore(executor): De-duplicate `TrieAccount`
## Overview

Alloy offers the `TrieAccount` type in `alloy-trie` [here](https://github.com/alloy-rs/trie/blob/main/src/account.rs), and we've re-defined it in `kona-executor` for the sake of the in-memory trie DB:

https://github.com/op-rs/kona/blob/b3ce5eec7838aaaffc36e79c0b941fe820f6ec17/crates/executor/src/db/account.rs#L7-L18

We can de-duplicate this type and use `alloy-trie`'s.

clabby:
## Overview

De-duplicates the `TrieAccount` type in favor of using `alloy-trie`'s.

closes #975


refcell (approved):
Very nice

Issue #985, opened 2025-01-31
feat(kona-proof-interop/host): Load in all `RollupConfig`s for the pre-state.


clabby:
## Overview

The client program only needs access to a single `RollupConfig` for the first sub-problem, but in the second problem, it may need to re-execute multiple blocks from different L2 chains. To support this, we'll need to update the `LocalKeyValueStore` in `kona-host` to pass through a serialized array of all of the various rollup configurations. We will also need to adjust the `BootInfo` type in `kona-proof-interop` to parse the serialized array in to a hashmap, keying each rollup config by L2 chain ID.

PR #986, opened 2025-01-31
feat(proof-interop): Support multiple `RollupConfigs` in boot routine

clabby:
## Overview

Adds support for loading multiple `RollupConfig`s into the client program during the boot routine.

closes https://github.com/op-rs/kona/issues/985

### DIFF
--- a/bin/client/src/interop/transition.rs
+++ b/bin/client/src/interop/transition.rs
@@ -31,31 +31,42 @@ pub(crate) async fn sub_transition<P, H>(
         >,
     >,
     boot: BootInfo,
-    pre: PreState,
 ) -> Result<(), FaultProofProgramError>
 where
     P: PreimageOracleClient + Send + Sync + Debug + Clone,
     H: HintWriterClient + Send + Sync + Debug + Clone,
 {
     // Check if we can short-circuit the transition, if we are within padding.
-    if let PreState::TransitionState(ref transition_state) = pre {
+    if let PreState::TransitionState(ref transition_state) = boot.agreed_pre_state {
         if transition_state.step >= transition_state.pre_state.output_roots.len() as u64 {
             info!(
                 target: "interop_client",
                 "No derivation/execution required, transition state is already saturated."
             );
 
-            return transition_and_check(pre, None, boot.claimed_post_state, None);
+            return transition_and_check(boot.agreed_pre_state, None, boot.claimed_post_state, None);
         }
     }
 
     // Fetch the L2 block hash of the current safe head.
-    let safe_head_hash = fetch_l2_safe_head_hash(oracle.as_ref(), &pre).await?;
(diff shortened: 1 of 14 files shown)


clabby:
## 📚 $\text{Stack Overview}$

* #986 👈
* `main`

bin/client/src/interop/transition.rs
```diff
@@ -31,31 +31,42 @@ pub(crate) async fn sub_transition<P, H>(
...
+    let active_l2_chain_id = boot
+        .agreed_pre_state
+        .active_l2_chain_id()
+        .ok_or(FaultProofProgramError::StateTransitionFailed)?;
+    let rollup_config = boot
+        .rollup_configs
+        .get(&active_l2_chain_id)
+        .cloned()
+        .map(Arc::new)
+        .ok_or(FaultProofProgramError::StateTransitionFailed)?;
```
refcell, line 64: Seems like we could pull this into a function on the `boot` itself.

e.g.
```rust
impl BootInfo {
    /// Returns the [RollupConfig] for the active L2 chain of the boot info.
    pub fn active_rollup_config(&self) -> Result<RollupConfig, BootInfoError> {
        let active_l2_chain_id = self.agreed_pre_state.active_l2_chain_id().ok_or(BootInfoError::MissingActiveChainId)?;
        self.rollup_configs.get(&active_l2_chain_id).cloned().ok_or(BootInfoError::MissingRollupConfig(active_l2_chain_id)
    }
}
```

bin/host/src/interop/local_kv.rs
```diff
@@ -37,14 +38,11 @@ impl KeyValueStore for LocalKeyValueStore {
...
             L2_ROLLUP_CONFIG_KEY => {
                 let rollup_configs = self.cfg.read_rollup_configs().ok()?;
-                let active_rollup_config = rollup_configs
-                    .get(&self.cfg.active_l2_chain_id().ok()?)
-                    .cloned()
-                    .unwrap_or_default();
-                let serialized = serde_json::to_vec(&active_rollup_config).ok()?;
+                let serialized =
+                    serde_json::to_vec(&rollup_configs.into_iter().collect::<HashMap<_, _>>())
+                        .ok()?;
```
refcell, line 45: Feels like this could also be squashed into a method on the `InteropHostCli`.

e.g. `pub fn config_vec(&self) -> Result<Vec<RollupConfig>, ...> { ... }`

bin/client/src/interop/transition.rs
clabby: Yeah will do :)

bin/host/src/interop/local_kv.rs
clabby: The `InteropHostCli` does already have `read_rollup_configs`, which returns a `HashMap` keying the L2 chain ID to the `RollupConfig`s.

That said I was originally trying to pass these as just an array of values - no need to iter -> collect again. Changed it back up to just serialize the map verbatim.

PR #990, opened 2025-01-31
fix(executor): Removes EIP-7002 and EIP-7251 Pre-block Calls

refcell:
### Description

Removes EIP-7002 and EIP-7251 pre-block calls as per the op stack specs.


refcell:
## 📚 $\text{Stack Overview}$

* #990 👈
* `main`

PR #1001, opened 2025-02-04
feat(mpt): Copy-on-hash

clabby:
## Overview

Adjusts the MPT crate to copy-on-hash rather than hash-in-place when computing the root hash of the trie.

This is less efficient, but is needed for completing execution of Optimistic blocks in the interop proof. Doing this enables for us to not rely solely on the `L2AccountProof` hint, which depends on the block the proof is fetched within being _canonical_.

### Benchmarks

The slowdown isn't crazy significant, but more `keccak256` operations may not be amazing for users running `kona` on zkVMs.


clabby:
## 📚 $\text{Stack Overview}$

* #1001 👈
* `main`

refcell:
IMO this hit is not significant enough to be a blocker for the interop workstream.

clabby:
> IMO this hit is not significant enough to be a blocker for the interop workstream.

Yeah, agree. @ratankaliani & @hashcashier, please report back in the issues if the additional cache is needed for your usecases.

Issue #921, opened 2025-01-22
[TRACKING] Interop support


clabby:
## Overview

This ticket serves as a high-level tracker for all tasks related to the completion of interop support within `kona`.

For a high-level ELI20 explainer of how interop proofs will work, see hackmd.io/@clabby/byf1flivje (official specs coming soon.)

clabby:
Also linking github.com/op-rs/maili/issues/27. Once this ticket has been mostly closed out and the proof has stabilized, we will be moving the `kona-interop` crate over to [`maili`](https://github.com/op-rs/maili).

PR #1003, opened 2025-02-05
feat(host): Reintroduce `L2BlockData` hint

clabby:
## Overview

Reintroduces the `L2BlockData` hint to the interop host. This was removed temporarily during the refactor, but is now supported by the L2 chain ID in all hints feature.

PR #1004, opened 2025-02-05
feat(client): Superchain Consolidation

clabby:
## Overview

Implements support for sub-problem #2 of the interop proof, superchain consolidation.

In this stage of the program, all pending results are aggregated together into the `SuperchainConsolidator`, which
attempts resolution of the `MessageGraph`. If resolution fails, blocks are re-executed with only their deposit transactions,
guaranteeing a block with no invalid executing messages.

Finally, the `SuperRoot` is re-formed, hashed, and compared with the disputed post-state commitment.

closes https://github.com/op-rs/kona/issues/921


clabby:
## 📚 $\text{Stack Overview}$

* #1004 👈
* #1003
* #1001
* `main`

PR #1010, opened 2025-02-06
feat(host): Accelerate all BLS12-381 Precompiles

### LINKED ISSUE #1000: feat: ensure BLS precompiles are all accelerated
Also need size limits for the variable-sized input precompiles.

refcell:
### Description

Updates the `kona-host` to accelerate all BLS12-381 precompiles.

Fixes #1000


clabby (approved):
Looks good for the host. Feel free to follow up in another PR or do it here, but this is only the host end of the equation - we'll also need to add accelerated versions of the precompiles in the client software (with limits on input size enforced.)

https://github.com/op-rs/kona/blob/2419b29fe7107f01835b0f1681321943dad1bad8/bin/client/src/precompiles/mod.rs#L47-L49

Looks like BLS pairing is already done :)

clabby (approved):
This LGTM. Would love to see action tests covering these precompiles across `op-program` + `kona` as a part of the Isthmus acceptance testing, but no need to rush those in. Nice work.

PR #1011, opened 2025-02-06
fix(client): No-op sub-transitions in Superchain STF

clabby:
## Overview

Fixes the logic for no-op'ing sub-transitions within the Superchain STF. The superchain STF happens once every second, regardless of if each individual chain within
the dependency set has actually progressed. In this case, we need to be able to short-circuit sub-problem #1.

### DIFF
--- a/bin/client/src/interop/mod.rs
+++ b/bin/client/src/interop/mod.rs
@@ -78,6 +78,12 @@ where
         }
     };
 
+    // Short-circuit the state transition if the agreed pre-state matches the claimed post-state.
+    if boot.agreed_pre_state_commitment == boot.claimed_post_state {
+        info!(target: "client_interop", "No-op transition, short-circuiting.");
+        return Ok(());
+    }
+
     // Load in the agreed pre-state from the preimage oracle in order to determine the active
     // sub-problem.
     match boot.agreed_pre_state {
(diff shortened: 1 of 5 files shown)


clabby:
## 📚 $\text{Stack Overview}$

* #1011 👈
* `main`

clabby (commented):
@ajsutton This still confuses me. Our `claimed_l2_timestamp` in the local boot info is never actually consulted, since we:
1. Check if we can no-op the transition based off of if `pre == post`
    - This does not use the claimed timestamp, as it both `pre` + `post` commit to the timestamp (and `pre` is always assumed to be trusted, so we can assume it's a well-formed commitment over the output root.)
3. Check if we can no-op the transition based off of whether or not the chain in sub-problem # 1 was able to progress in the 1s since the prestate.
    - Notably this does _not_ used the claimed timestamp, for this check to be correct we have to infer the next timestamp from the safe head timestamp + chain block time.

Still not entirely sure why we need the program to know about a claimed timestamp at all.

clabby:
@refcell Above comment can be discussed out of band; The fixes in this PR should still hold up. Good for review 👍

refcell (approved):

ajsutton:
The game proposal timestamp is needed to differentiate between the two cases where:

1. Agreed state == super root at proposal timestamp -> post state is therefore _also_ the super root at proposal timestamp (the state transition should not go past the proposal timestamp)
2. Agreed state = super root at a timestamp boundary _before_ the proposal timestamp -> post state is a transition state at step 1 (with appropriate pending progress etc).

If we don't have that, then every game will eventually transition to the invalid hash because you can keep moving right in the top half of the game such that the super root timestamp is in the future and there won't be sufficient batch data on L1 to reach that block because it may not even exist yet.

PR #1012, opened 2025-02-06
chore(client): Dedup MSM Required Gas Fn

refcell:
### Description

Small cleanup PR to dedup the MSM required gas calculation utility method.

This will need to be removed and revm's utility used once it's made public in the next revm release.

PR #1013, opened 2025-02-06
chore(client): Remove Hardcoded Precompile Addresses

refcell:
### Description

Updates `kona-client` precompiles to use the addresses defined in `revm-precompiles`.

Reduces the chance that precompile addresses are incorrectly input.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -9901,3 +9901,8 @@ dependencies = [
  "cc",
  "pkg-config",
 ]
+
+[[patch.unused]]
+name = "revm"
+version = "20.0.0-alpha.1"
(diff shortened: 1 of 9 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1013 👈
* #1012
* `main`

refcell:
Darn rust-analyzer must have automatically enabled `blst` feature... forgot `bls12_381` module in revm is hidden behind `blst`. This is blocking. Once solution may be to PR into revm and try to hoist these constants to the parent module so they can be used without `blst` being enabled.

refcell:
FYI - tracking this in revm [here](https://github.com/bluealloy/revm/issues/2081)

refcell:
Going to ice this for now since this will need to wait on revm.

PR #1017, opened 2025-02-07
feat(host): Proactive hints

clabby:
## Overview

Adds a feature to the `OnlineHostBackend` that allows for "proactive hints," which fetch data from the remote sources and populate the key-value store immediately rather than upon the next request for a preimage.

With this, the interop host sets the `L2BlockData` hint to a proactive hint in order to ensure that the host fetches the optimistic block data without another hint interrupting it.

### DIFF
--- a/bin/host/src/backend/online.rs
+++ b/bin/host/src/backend/online.rs
@@ -7,15 +7,16 @@ use kona_preimage::{
     errors::{PreimageOracleError, PreimageOracleResult},
     HintRouter, PreimageFetcher, PreimageKey,
 };
-use std::{hash::Hash, str::FromStr, sync::Arc};
+use kona_proof::{errors::HintParsingError, Hint};
+use std::{collections::HashSet, hash::Hash, str::FromStr, sync::Arc};
 use tokio::sync::RwLock;
-use tracing::{error, trace, warn};
+use tracing::{debug, error, trace};
 
 /// The [OnlineHostBackendCfg] trait is used to define the type configuration for the
 /// [OnlineHostBackend].
 pub trait OnlineHostBackendCfg {
     /// The hint type describing the range of hints that can be received.
-    type Hint: FromStr + Hash + Eq + PartialEq + Send + Sync;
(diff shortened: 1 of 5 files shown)


clabby:
## 📚 $\text{Stack Overview}$

* #1017 👈
* `main`

bin/host/src/backend/online.rs
```diff
@@ -86,8 +96,19 @@ where
     /// Set the last hint to be received.
     async fn route_hint(&self, hint: String) -> PreimageOracleResult<()> {
         trace!(target: "host-backend", "Received hint: {hint}");
-        let mut hint_lock = self.last_hint.write().await;
-        hint_lock.replace(hint);
+
+        let parsed_hint =
+            hint.parse::<Hint<C::HintType>>().map_err(|_| PreimageOracleError::KeyNotFound)?;
```
refcell, line 101: Is there a downside to modifying the `HintType` itself to describe whether or not the `HintType` is proactive?

bin/host/src/backend/online.rs
```diff
@@ -86,8 +96,19 @@ where
...
+        if self.proactive_hints.contains(&parsed_hint.ty) {
+            debug!(target: "host-backend", "Proactive hint received; Immediately fetching {hint}");
+            H::fetch_hint(parsed_hint, &self.cfg, &self.providers, self.kv.clone())
+                .await
+                .map_err(|e| PreimageOracleError::Other(e.to_string()))?;
```
refcell, line 106: This is now blocking the router right? Should this not be spawned?

refcell (approved):

bin/host/src/backend/online.rs
clabby: Yeah, mostly that the `HintType` would need to hold onto some extra metadata that leaks info about the `HintRouter`. We could do that, but felt a bit nicer to just register proactive hints with the backend, since it's more of an implementation detail of the `HintRouter` trait implementation.

bin/host/src/backend/online.rs
refcell: Nice didn't know from which context the router was being spawned

PR #1022, opened 2025-02-10
feat(interop): Support full timestamp invariant

clabby:
## Overview

Adds support for the full timestamp invariant for the interop hardfork, enforcing that a message is invalid if the Interop activation time had not happened at the time the initiating message was sent.

### DIFF
--- a/crates/interop/src/errors.rs
+++ b/crates/interop/src/errors.rs
@@ -31,6 +31,11 @@ pub enum MessageGraphError<E> {
     /// Invalid messages were found
     #[error("Invalid messages found on chains: {0:?}")]
     InvalidMessages(Vec<u64>),
+    /// Missing a [RollupConfig] for a chain ID
+    ///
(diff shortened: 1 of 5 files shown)


crates/interop/src/errors.rs
```diff
@@ -31,6 +31,11 @@ pub enum MessageGraphError<E> {
     /// Invalid messages were found
     #[error("Invalid messages found on chains: {0:?}")]
     InvalidMessages(Vec<u64>),
+    /// Missing a [RollupConfig] for a chain ID
+    ///
+    /// [RollupConfig]: maili_registry::RollupConfig
```
refcell, line 36: this should be from `maili_genesis`, I think the type is re-exported from `maili_registry` just fyi

```suggestion
    /// [RollupConfig]: maili_genesis::RollupConfig
```

clabby: Ah nice - updated

PR #1034, opened 2025-02-13
feat(client): Wire up `L2PayloadWitness` hint for single-chain proof

clabby:
## Overview

Experimental wiring of the `L2PayloadWitness` hint, allowing `op-reth` to be used as the backing L2 execution client for `kona-host`.


refcell (approved):
Codecov hurts me soul

PR #1052, opened 2025-02-18
feat(client): support cannon mips64r2

pcw109550:
Luckily we only need DCLZ support after ruint patch https://github.com/recmo/uint/pull/431 is applied. So closing below monorepo PR and plan to reopen with minimal diff only for DCLZ
- Cannon supporting minimal mips64r2: https://github.com/pcw109550/optimism/tree/pcw109550/cannon-mipsevm-full-mips64r2


pcw109550:
Closing this because https://github.com/op-rs/kona/pull/1054 is merged.

PR #1057, opened 2025-02-19
fix(std-fpvm): Allow non-const fn with mut ref

clabby:
## Overview

Fixes the lint routine for the `cannon-builder` image. These images use the nightly toolchain, rather than the MSRV, to build.

Also loosens the requirements for the build images' platforms, since they're now distributed as cross-platform builds. Rejoice, Apple laptop users :D


refcell (approved):
Very nice

PR #1062, opened 2025-02-20
chore(book): Move the Monorepo Doc to Archives

refcell:
### Description

Moves the monorepo doc to the archives.

PR #1063, opened 2025-02-20
feat(book): Umbrella Crate RFC

refcell:
### Description

Introduces an RFC that discusses adding one or multiple "umbrella" crates to re-export `kona` crates.

Since `kona` has a growing number of crates that all have independent versions,
providing an easier way to consume `kona` as a dependency is increasingly important.

The `Proposed Solution` section is not yet filled out because I am looking for opinions / preferences
here before suggesting something.


refcell:
## 📚 $\text{Stack Overview}$

* #1063 👈
* #1062
* `main`

refcell:
Example of the single umbrella crate: https://github.com/op-rs/kona/pull/1059

Issue #1009, opened 2025-02-05
feat(ci): Add OP Succinct Cycle Tracking


ratankaliani:
To confirm that modifications to `kona` aren't greatly increasing the costs of running the cycle counts of consumer crates (e.g. OP Succinct), add a CI job which outputs the cycle count of OP Succinct with the corresponding changes from Kona.

ratankaliani:
https://github.com/op-rs/kona/pull/1001

PR #1064, opened 2025-02-20
feat: add a check on OP Succinct cycle count in CI

leruaa:
Here a first draft PR that handle the following issue: #1009. 

I added a new binary that build an oracle in memory from the test data and send it to the client program that run in the SP1 zkVM.


refcell:
Closing this for now since it's stale. Feel free to re-open if required

PR #1071, opened 2025-02-20
fix: Maili Shadows

refcell:
### Description

Fixes up `maili` shadow crates.

PR #1072, opened 2025-02-21
fix: Remove Maili Shadows from Workspace

refcell:
### Description

Since the `maili` shadow crates are very tiny and intentionally unchanging in config,
I realized we can actually just remove them from the workspace and reference the published
crate versions directly. We can then publish these on a one-off basis. I expect we won't
need these soon since we can port downstream users to `kona` crates regardless.

Alternative to #1071


refcell:
## 📚 $\text{Stack Overview}$

* #1072 👈
* #1071
* `main`

mattsse (approved):
long live kona

PR #1075, opened 2025-02-21
fix(registry): Use `superchain-registry` as a submodule

clabby:
## Overview

In the move over to the monorepo, `superchain-registry` was vendored rather than used as a submodule. This PR deletes the vendored version of the repository and replaces it with a submodule.


refcell:
Ah good catch thank you

refcell:
FYI, once https://github.com/op-rs/kona/pull/1060 is merged, this can be rebased on `main` and the `no_std` checks will pass. Somehow in fixing the merge conflicts, the maili shadow crates were not removed from the `no_std` checks.

refcell (approved):

PR #1088, opened 2025-02-21
feat(proof): EIP-2935 lookback

### LINKED ISSUE #934: feat(client): Use EIP-2935 for history-by-number lookups
## Overview

The history-by-number lookups we use right now are very inefficient, being a linear-time commit-reveal walkback:

https://github.com/op-rs/kona/blob/3e8844749515becf132181eaca01fa649374d08a/crates/proof-sdk/proof/src/l2/chain_provider.rs#L65-L68

With [`EIP-2935`](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2935.md) coming in Isthmus (L2 Pectra), we have the opportunity to drastically improve the performance of this operation. Once this has landed, we should refactor our oracle-backed history-by-number implementations to take advantage of the feature.

This is pretty critical for the performance of #927, where remote message lookups have to use this route.

clabby:
## Overview

Adds the ability for both the `OracleL2ChainProvider` and the `OracleInteropProvider` to utilize the [EIP-2935](github.com/ethereum/eips/blob/master/eips/eip-2935.md) contract for more efficient history-by-number lookups, following the activation of the Isthmus hardfork.

Behavior:
- providers, during history-by-number lookups, will attempt to use the EIP-2935 contract if Isthmus is active.
- If, for any reason, the lookup fails, the legacy linear walkback is used.
    - In practice, the only reason the lookup should ever fail is during the first 4-5 hours worth of blocks, while the contract's ring buffer is being saturated. However, for safety, the linear lookup fallback is retained for any possible error in the EIP-2935 lookup.

Tests have been added for the lookup function, using a mock state / account trie and inserting history accumulation information.

closes #934


crates/proof/proof-interop/src/provider.rs
```diff
@@ -136,9 +143,37 @@ where
             return Err(OracleProviderError::BlockNumberPastHead(number, header.number));
         }
 
+        // Set the chain ID for the trie hints, and explicitly drop the lock.
+        let mut chain_id_lock = self.chain_id.write();
+        *chain_id_lock = Some(chain_id);
+        drop(chain_id_lock);
```
clabby, line 149: Using this spin-locked variable in the provider itself to communicate with the `TrieHinter` implementation is ugly, but serves to not break the `TrieHinter` interface for now. I'll throw up a ticket to improve this API.

crates/proof/proof/src/lib.rs
```diff
@@ -3,7 +3,7 @@
 #![deny(unused_must_use, rust_2018_idioms)]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
 #![cfg_attr(not(test), warn(unused_crate_dependencies))]
-#![no_std]
+#![cfg_attr(not(test), no_std)]
```
refcell, line 6: I'd prefer we not do this if we can avoid it.

If we need `std` in tests, just make the test itself as `#[cfg(feature = "std")]` or whatever feature flag allows for `std`. Here we have an `std` feature flag for `kona-proof` so i think we can use that.

refcell: The reason why is this leaks `std` into all tests including `no_std` tests. It loosens the testing environment constraints.

refcell (approved):

Issue #916, opened 2025-01-22
chore(host): Remove `anyhow` dependency


clabby:
## Overview

Right now, the host uses `anyhow` as a dependency for error handling. Anyhow is not great for being able to pattern match on errors as the consumer.

Remove `anyhow` from the host (and the whole workspace) in favor of `thiserror`.

einar-oplabs:
[Motivation](https://momori.dev/posts/rust-error-handling-thiserror-anyhow/)

PR #1093, opened 2025-02-23
chore(host): Replace anyhow with thiserror

Levi0804:
This PR is a work in progress addressing #916.

I am aware that some of the changes in this PR introduce breaking changes in other parts of the workspace. I created this PR to take reviews on whether my approach looks good before proceeding further. Please let me know if these changes look good to you.


bin/host/src/server.rs
```diff
@@ -20,6 +19,16 @@ pub struct PreimageServer<P, H, B> {
...
 
+#[derive(Debug, thiserror::Error)]
+enum PreimageServerError {
+    #[error("Failed to serve preimage request: {0}")]
+    PreimageRequestFailed(PreimageOracleError),
+    #[error("Failed to route hint: {0}")]
+    RouteHintFailed(PreimageOracleError),
+    #[error("Join error: {0}")]
+    ExecutionError(#[from] tokio::task::JoinError),
+}
```
mattsse, line 30: this is used it pub api and should also be pub

a one-line doc also wouldn't hurt here

mattsse (changes requested):
smol style nit

bin/host/src/server.rs
refcell: Agree with @mattsse here

Levi0804:
Will it be okay to allow `unnameable_types` in this case?

bin/host/src/interop/cfg.rs
```diff
@@ -98,9 +97,28 @@ pub struct InteropHost {
     pub rollup_config_paths: Option<Vec<PathBuf>>,
 }
 
+/// An error that can occur when handling interop hosts
+#[derive(Debug, thiserror::Error)]
```
refcell, line 101: Add [`derive_more::From`](https://docs.rs/derive_more/latest/derive_more/derive.From.html) here - see comments below.

bin/host/src/server.rs
```diff
@@ -20,6 +19,17 @@ pub struct PreimageServer<P, H, B> {
     backend: Arc<B>,
 }
 
+/// An error that can occur when handling preimage requests
+#[derive(Debug, thiserror::Error)]
+pub enum PreimageServerError {
```
refcell, line 24: This will cause ci to error because the type is only visible within the `server` module. It has to be visible at the top-level of the crate so linting will pass.

If you run `just l` locally which runs linting, you'll see a visibility error.

The solution is in the host's `lib.rs` file, make the `PreimageServerError` visible by adding it to the `pub use server::` line [here](https://github.com/op-rs/kona/blob/main/bin/host/src/lib.rs#L5).

You'll also need to do this for the other error types introduced in this PR.

bin/host/src/server.rs
```diff
@@ -20,6 +19,17 @@ pub struct PreimageServer<P, H, B> {
...
+    #[error("Failed to serve preimage request: {0}")]
+    PreimageRequestFailed(PreimageOracleError),
+    #[error("Failed to route hint: {0}")]
+    RouteHintFailed(PreimageOracleError),
+    #[error("Join error: {0}")]
+    ExecutionError(#[from] tokio::task::JoinError),
```
refcell, line 30: Our linting rules are fairly strict as well so these enum variants also need documentation.

bin/host/src/single/cfg.rs
```diff
@@ -109,9 +108,24 @@ pub struct SingleChainHost {
...
+    #[error("Error handling preimage request: {0}")]
+    PreimageServerError(#[from] PreimageServerError),
+    #[error("IO error: {0}")]
+    IOError(#[from] std::io::Error),
+    #[error("Failed deserializing RollupConfig: {0}")]
+    ParseError(#[from] serde_json::Error),
+    #[error("Join error: {0}")]
+    ExecutionError(#[from] tokio::task::JoinError),
+    #[error("Error: {0}")]
+    Other(&'static str),
```
refcell, line 123: These enum variants will need docs

bin/host/src/interop/cfg.rs
```diff
@@ -98,9 +97,28 @@ pub struct InteropHost {
...
+    #[error("Failed deserializing RollupConfig: {0}")]
+    ParseError(#[from] serde_json::Error),
+    #[error("Join error: {0}")]
+    ExecutionError(#[from] tokio::task::JoinError),
+    #[error("Rpc Error: {0}")]
+    RcpError(#[from] alloy_transport::RpcError<alloy_transport::TransportErrorKind>),
+    #[error("No provider found for chain ID: {0}")]
+    RootProviderError(u64),
+    #[error("Error: {0}")]
+    Other(&'static str),
```
refcell, line 116: These enum variants will need docs

refcell:
Few more comments for CI, otherwise this is looking good. Thanks for taking this on!

Levi0804:
> Few more comments for CI, otherwise this is looking good. Thanks for taking this on!

Happy to contribute! Let me know if these changes look good. I squashed and rebased to clean up some mistakes...

refcell (approved):
Nice work!

PR #1095, opened 2025-02-24
chore(deps): Bump reth

emhane:
Based on https://github.com/op-rs/kona/pull/1094

- Bumps reth to v1.2.0
- Updates docs to reflect that `ChainProvider` is for L1 chain

### DIFF
--- a/.github/workflows/rust_ci.yaml
+++ b/.github/workflows/rust_ci.yaml
@@ -145,7 +145,7 @@ jobs:
       - name: Install Rust stable toolchain
         uses: dtolnay/rust-toolchain@stable
         with:
-          toolchain: 1.82
+          toolchain: 1.85
       - uses: Swatinem/rust-cache@v2
         with:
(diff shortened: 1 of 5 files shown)


clabby:
Our MSRV in `kona` is rustc `1.81`, needed to be able to support R0/SP-1 zkVM targets. @ratankaliani - have you guys been able to upgrade your target yet?

refcell:
The rust MSRV needs to be bumped here to `1.82`. We need to check in with Succinct and other downstream kona users to ensure this will work for them.

ratankaliani:
We now support Rust 1.82, should be good to go to bump.

refcell:
@clabby updated the MSRV in https://github.com/op-rs/kona/pull/1097.

Please rebase on trunk to include these changes :)

clabby (approved):

refcell (approved):

PR #1100, opened 2025-02-24
feat(bin-utils): Add prometheus server initializer

clabby:
## Overview

Adds a shared prometheus server initializer to `kona-bin-utils`, taken from `kona-nexus`.

PR #1099, opened 2025-02-24
chore(nexus): Use `kona-bin-utils`

clabby:
## Overview

Uses the `kona-bin-utils` crate for `kona-nexus`. The `init_tracing_subscriber` function in `kona-bin-utils` has been altered to accept an optional `EnvFilter`. The other logic, for "is-tty," should be handled internally by the FMT subscriber already.

PR #1098, opened 2025-02-24
chore(bin): Split up bin utilities

clabby:
## Overview

Splits up the binary utilities in `kona-host` into a new crate, `kona-bin-utils`. This crate is pretty thin for now, but as we grow the number of applications in the monorepo, we'll likely have a good bit of logic that's not worth duplicating.

Closes #1108

PR #1101, opened 2025-02-24
chore(workspace): Adjust build recipes

clabby:
## Overview

Adjusts the build recipes for `cannon` and `asterisc` to just build the `kona-client` binary. We have the `check-no-std.sh` script running in CI now, so it's really only important that we check `kona-client`'s ability to compile to these targets in CI now.


clabby:
## 📚 $\text{Stack Overview}$

* #1101 👈
* #1100
* #1099
* #1098
* `main`

refcell (approved):

Issue #1103, opened 2025-02-25
[Feature] Batch Submitter Support


ratankaliani:
Noticed on Base Sepolia that `kona` does not work because the `RollupConfig` from `kona-genesis` does not support the `batchSubmitter` field on `RollupConfig`, which is what chains use to define their `batchSubmitter`, not the genesis-defined `batcherAddress`.

Note that on most chains, `batcherAddress` = `batchSubmitter`, so this is not an issue, but it is incorrect.

Specifically here, BatchSubmitter address (https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/sepolia/base.toml#L49) does not match `batcherAddress` (https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/configs/sepolia/base.toml#L37).

Recommended path: Add `batchSubmitter` to `RollupConfig`, and use `batchSubmitter`, instead of the `SystemConfig`'s `batcherAddr` for validating the posted batches.

PR #1106, opened 2025-02-25
fix(derive): Use `SystemConfig` batcher key for DAP

clabby:
## Overview

Alters the data availability provider in the derivation pipeline to use a dynamic batcher address. There is currently an issue in `kona` where chains that have changed their batch submitter key since genesis will not read batches from L1 correctly.

closes https://github.com/op-rs/kona/issues/1103

### DIFF
--- a/bin/client/src/interop/transition.rs
+++ b/bin/client/src/interop/transition.rs
@@ -6,11 +6,7 @@ use alloc::sync::Arc;
 use alloy_consensus::Sealed;
 use alloy_primitives::B256;
 use core::fmt::Debug;
-use kona_derive::{
-    errors::{PipelineError, PipelineErrorKind},
-    traits::{L2ChainProvider, SignalReceiver},
-    types::ResetSignal,
-};
+use kona_derive::errors::{PipelineError, PipelineErrorKind};
 use kona_driver::{Driver, DriverError};
 use kona_executor::{KonaHandleRegister, TrieDBProvider};
 use kona_preimage::{HintWriterClient, PreimageOracleClient};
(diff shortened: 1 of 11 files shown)


bin/client/src/interop/transition.rs
```diff
@@ -98,14 +102,31 @@ where
...
+                l2_safe_head,
+                l1_origin: cursor.read().origin(),
+                system_config: l2_provider
+                    .system_config_by_number(l2_safe_head.block_info.number, rollup_config.clone())
+                    .await
+                    .ok(),
+            }
+            .signal(),
+        )
+        .await?;
```
refcell, line 128: This looks to be duplicated. I think I'd prefer to introduce a new method on the `Pipeline` trait to populate the initial system config. Otherwise, it is easy user error to _forget_ to do this.

clabby: Yeah can do.

clabby: Updated to hide the reset in the constructor. Good?

bin/client/src/interop/transition.rs
```diff
@@ -105,7 +105,8 @@ where
         beacon,
         l1_provider.clone(),
         l2_provider.clone(),
-    );
+    )
+    .await?;
```
refcell, line 109: Yea that's ok i guess

crates/protocol/derive/src/sources/calldata.rs
```diff
@@ -21,8 +21,6 @@ where
     pub chain_provider: CP,
     /// The batch inbox address.
     pub batch_inbox_address: Address,
-    /// The L1 Signer.
-    pub signer: Address,
```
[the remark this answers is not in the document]
clabby: It's a bit of a pain in the ass - would require the DAP to hold a reference to `L1Traversal`, which is owned by `L1Retrieval`. We could re-architect in this direction, but it's a pretty invasive change.

bin/client/src/interop/transition.rs
refcell: Async instantiators are not my favorite but it's acceptable. At that point a builder pattern is preferred but can be ticketed out after.

crates/protocol/derive/src/sources/calldata.rs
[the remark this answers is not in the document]
refcell: How come? Couldn't you reference the actual sys config given to the l1 traversal and also pass it into the calldata source?

I view changing the api of every method on data sources a much more invasive change

refcell (approved):

PR #1133, opened 2025-02-27
feat(protocol): implement buffer for brotli

### LINKED ISSUE #1081: [Feature] Brotli Compression
### Component

`kona-protocol` crate

### Describe the feature you would like

Perform brotli compression using the `brotli::CompressorWriter` and a buffer instead of storing both the raw and compressed data and re-compressing the entire raw data as currently done.

### Additional context

See: https://github.com/op-rs/maili/pull/14.

Notice, tests are failing here because of a bug with the implementation.

SkandaBhat:
Fixes #1081 

Note: There are two tests I have marked as ignored, because they were testing against hardcoded compressed values. Since we are implementing a buffer, the compressed values will change, and instead we should be testing roundtrip values instead. I have added a test to check random roundtrips, in addition to the existing roundtrip test.


clabby:
Hi - This PR doesn't look to solve the issue noted in the ticket. The hard coded compressed values should still work after this change.

Going to go ahead and close this for now, glad to continue conversation on the ticket if you'd like to pick this back up.

PR #1136, opened 2025-02-27
feat(node): Engine Controller

### LINKED ISSUE #1131: feat(node/engine): Engine Controller
### Description

A core part of the rollup node is the component responsible for tracking the execution engine.

At the top level, that is an [EngineController](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/engine/engine_controller.go), which contains the engine client implemented for kona in https://github.com/op-rs/kona/pull/1120.

This ticket is to set up the `EngineController` inside `bins/node/src/engine/controller.rs`. For reference, hilo had a decently along [implementation of the `EngineController`](https://github.com/op-rs/hilo/blob/main/crates/engine/src/controller.rs).

Once the `EngineController` is brought into `kona-node`'s `engine` module, it should be used by the "Pilot" instead of the `EngineClient`.

How I envision the end-game communication for `kona-node` is through a message-based actor system. But to start, `kona-node` should use the `Pilot` to "orchestrate" different services and handle message passing. For example, when the node is started, the `Pilot` will have a sync service (might be spun into a separate thread) that will send messages to the `Pilot` to try and update the engine on sync. This is similar to how the `op-node` syncs with it's event-based-system and it's [`SyncDeriver`](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/driver/state.go#L412).

refcell:
### Description

Sets up the engine component of the consensus node.


clabby (approved):
Couple things I'd like to change about this, but can get it in to keep moving.

PR #1155, opened 2025-03-01
fix: 2021 Edition Fragment Specifier

refcell:
### Description

Locally breaking macro rules are acceptable since crates are versioned.
Further, the newly covered `const` and `_` syntax patterns are unused
when the expression fragment specifier is used currently, so there won't
be breaking backwards compatibility.

Essentially, the change from `expr_2021` -> `expr` now covers `const` and
`_` which would be a breaking change for prior usage of the macro.

The Rust edition guide [sums this up][sum] fairly well.

[sum]: https://doc.rust-lang.org/edition-guide/rust-2024/macro-fragment-specifiers.html


refcell:
## 📚 $\text{Stack Overview}$

* #1155 👈
* `main`

clabby (approved):

PR #1156, opened 2025-03-01
feat(registry): Lookup `Chain` + `RollupConfig` by identifier

clabby:
## Overview

Adds some simple convenience functions that fetch `Chain` and `RollupConfig`s by their human-readable identifiers. This is useful for the rollup node entrypoint, where we want to have a nice `--network` flag that doesn't only accept opaque numeric chain IDs for the argument.

### DIFF
--- a/crates/protocol/registry/src/chain_list.rs
+++ b/crates/protocol/registry/src/chain_list.rs
@@ -4,9 +4,40 @@ use alloc::{string::String, vec::Vec};
 
 /// List of Chains.
 #[derive(Debug, Clone, Default, Hash, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+#[serde(transparent)]
 pub struct ChainList {
     /// List of Chains.
     pub chains: Vec<Chain>,
+}
+
+impl ChainList {
+    /// Fetch a [Chain] by its identifier.
+    pub fn get_chain_by_ident(&self, identifier: &str) -> Option<&Chain> {
+        self.chains.iter().find(|c| c.identifier.eq_ignore_ascii_case(identifier))
+    }
+
+    /// Fetch a [Chain] by its chain id.
+    pub fn get_chain_by_id(&self, chain_id: u64) -> Option<&Chain> {
+        self.chains.iter().find(|c| c.chain_id == chain_id)
+    }
+
+    /// Returns the number of chains.
+    pub fn len(&self) -> usize {
+        self.chains.len()
+    }
+
+    /// Returns true if the list is empty.
+    pub fn is_empty(&self) -> bool {
+        self.chains.is_empty()
+    }
+}
+
+impl Iterator for ChainList {
+    type Item = Chain;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.chains.pop()
+    }
 }
 
 /// A Chain Definition.
--- a/crates/protocol/registry/src/chain_list.rs
+++ b/crates/protocol/registry/src/chain_list.rs
@@ -1,6 +1,7 @@
 //! List of OP Stack chains.
 
 use alloc::{string::String, vec::Vec};
+use alloy_chains::Chain as AlloyChain;
 
 /// List of Chains.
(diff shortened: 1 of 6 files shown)


crates/protocol/registry/src/chain_list.rs
```diff
@@ -4,11 +4,42 @@ use alloc::{string::String, vec::Vec};
...
+    }
+}
+
+impl Iterator for ChainList {
+    type Item = Chain;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        self.chains.pop()
+    }
+}
```
mattsse, line 41: I'd rather impl intoiterator here like vec, then this doesn't require pop

crates/protocol/registry/src/chain_list.rs
```diff
@@ -4,11 +4,42 @@ use alloc::{string::String, vec::Vec};
 
 /// List of Chains.
 #[derive(Debug, Clone, Default, Hash, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+#[serde(transparent)]
 pub struct ChainList {
     /// List of Chains.
     pub chains: Vec<Chain>,
 }
 
+impl ChainList {
```
mattsse, line 13: I think what could be nice here is `alloy_chains::Chain` support then you could look look via `Chain::optimism()` for example

mattsse (commented):
smol suggestions

mattsse (approved):

crates/protocol/registry/src/chain_list.rs
clabby: Yeah neat idea. Added.

crates/protocol/registry/src/chain_list.rs
```diff
@@ -1,14 +1,52 @@
...
+    }
+}
+
+impl IntoIterator for ChainList {
+    type Item = Chain;
+    type IntoIter = alloc::vec::IntoIter<Self::Item>;
+
+    fn into_iter(self) -> Self::IntoIter {
+        self.chains.into_iter()
+    }
```
refcell, line 47: Delete

crates/protocol/registry/src/chain_list.rs
```diff
@@ -1,14 +1,52 @@
...
 use alloc::{string::String, vec::Vec};
+use alloy_chains::Chain as AlloyChain;
 
 /// List of Chains.
 #[derive(Debug, Clone, Default, Hash, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+#[serde(transparent)]
 pub struct ChainList {
     /// List of Chains.
     pub chains: Vec<Chain>,
 }
```
refcell, line 12: Can use [`derive_more::IntoIterator`](https://docs.rs/derive_more/latest/derive_more/derive.IntoIterator.html) here instead of manual impl.

crates/protocol/registry/src/lib.rs
```diff
@@ -36,9 +37,36 @@ lazy_static::lazy_static! {
     pub static ref ROLLUP_CONFIGS: HashMap<u64, RollupConfig, DefaultHashBuilder> = _INIT.rollup_configs.clone();
 }
 
+/// Returns all available [Chain] identifiers.
+pub fn chain_idents() -> Vec<String> {
+    CHAINS.chains.iter().map(|c| c.identifier.clone()).collect()
+}
```
refcell, line 43: Is there a reason not to make this a method on `CHAINS`? (same goes for below functions as well)

refcell (changes requested):
Few comments

crates/protocol/registry/src/chain_list.rs
```diff
@@ -1,14 +1,48 @@
 //! List of OP Stack chains.
 
 use alloc::{string::String, vec::Vec};
+use alloy_chains::Chain as AlloyChain;
 
 /// List of Chains.
 #[derive(Debug, Clone, Default, Hash, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+#[serde(transparent)]
```
refcell, line 8: Nice

refcell (approved):

PR #1158, opened 2025-03-02
feat: add RPCConfig

### LINKED ISSUE #1118: feat(node): RPC Config
### Description

Introduce an RPC config that is used to setup the rpc server in the consensus node.

For reference, see the [op-node RPC Config](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/node/config.go#L84-L88)

varun-doshi:
fix #1118


bin/node/src/cli/config.rs
clabby, line 1: Can you please add docs to the module, struct, struct fields, and the `http_endpoint` function?

We've turned off doclints for now, but will require them soon.

PR #1160, opened 2025-03-03
Rename field name to be compatible with op-supervisor

SozinM:
This changes were tested, op-supervisor correctly parses the Executing Message

### DIFF
--- a/crates/protocol/interop/src/message.rs
+++ b/crates/protocol/interop/src/message.rs
@@ -28,7 +28,7 @@ sol! {
     /// @param id Encoded Identifier of the message.
     #[derive(Default, Debug, PartialEq, Eq)]
     #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-    event ExecutingMessage(bytes32 indexed msgHash, MessageIdentifier id);
+    event ExecutingMessage(bytes32 indexed payloadHash, MessageIdentifier identifier);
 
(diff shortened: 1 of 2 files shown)


crates/protocol/interop/src/message.rs
```diff
@@ -28,7 +28,7 @@ sol! {
     /// @param id Encoded Identifier of the message.
     #[derive(Default, Debug, PartialEq, Eq)]
     #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
-    event ExecutingMessage(bytes32 indexed msgHash, MessageIdentifier id);
+    event ExecutingMessage(bytes32 indexed payloadHash, MessageIdentifier identifier);
```
emhane, line 31: pls add this link to docs mentioning smthg about the choice of var names for correct json: https://github.com/ethereum-optimism/optimism/blob/4ba2eb00eafc3d7de2c8ceb6fd83913a8c0a2c0d/op-supervisor/supervisor/types/types.go#L61-L64

emhane (approved):
lgtm

emhane:
I don't think you have to worry about ci here, seems unrelated to your code, will force merge it when nitpick ab docs is fixed

crates/protocol/interop/src/message.rs
SozinM: Done!

PR #1162, opened 2025-03-03
Parse supervisor response

SozinM:


crates/external/rpc/src/api.rs
```diff
@@ -155,5 +155,5 @@ pub trait SupervisorApi {
         &self,
         messages: Vec<ExecutingMessage>,
         min_safety: SafetyLevel,
-    ) -> RpcResult<()>;
+    ) -> RpcResult<RpcResult<()>>;
```
emhane, line 158: this change is redundant since it's already using an `RpcResult`

crates/external/rpc/src/traits.rs
```diff
@@ -6,13 +6,17 @@ use alloy_primitives::Log;
...
+use jsonrpsee::{core::ClientError, types::ErrorObjectOwned};
+use jsonrpsee::core::RpcResult;
 use kona_interop::{CROSS_L2_INBOX_ADDRESS, ExecutingMessage, SafetyLevel};
 use tokio::time::error::Elapsed;
 
 /// Failures occurring during validation of [`ExecutingMessage`]s.
 #[derive(thiserror::Error, Debug)]
 pub enum ExecutingMessageValidatorError {
+    /// Safety level not sufficient
+    Unsafe,
```
emhane, line 19: nice! would be nice if variant wrapped the expected `SafetyLevel` and the safety level that we got, but maybe not possible with current op-supervisor code, so then it's out of scope for now

SozinM: I made manual parsing, but it looks ugly, maybe it's better to use regexp x)

crates/external/rpc/src/traits.rs
```diff
@@ -6,20 +6,33 @@ use alloy_primitives::Log;
...
 
+const SAFETY_START_MSG: &str = "(safety level: ";
+const UNKNOWN_CHAIN_MSG: &str = "unknown chain: ";
+const MINIMUM_SAFETY_MSG: &str = "does not meet the minimum safety";
 /// Failures occurring during validation of [`ExecutingMessage`]s.
 #[derive(thiserror::Error, Debug)]
 pub enum ExecutingMessageValidatorError {
+    /// Message does not minimum safety level
+    #[error("Message does not meet safety level. Expected: {0}, Got: {1}")]
+    MinimumSafety(SafetyLevel, SafetyLevel),
```
emhane, line 21: lets make this a struct so it's more clear which is which upon construction
```suggestion
    /// Message does not minimum safety level
    #[error("Message does not meet safety level. Expected: {0}, Got: {1}")]
    MinimumSafety { got: SafetyLevel, minimum: SafetyLevel },
```

crates/external/rpc/src/traits.rs
```diff
@@ -6,20 +6,33 @@ use alloy_primitives::Log;
...
+    /// Invalid chain
+    #[error("Unsupported chainId: {0}")]
+    UnknownChain(u64),
```
emhane, line 24:
```suggestion
    /// Invalid chain
    #[error("unsupported chain id: {0}")]
    UnknownChain(u64),
```
all error message should be lowercase
ref: https://doc.rust-lang.org/std/error/trait.Error.html

crates/external/rpc/src/traits.rs
```diff
@@ -51,16 +64,48 @@ pub trait ExecutingMessageValidator {
...
+                                        .try_into()
+                                    {
+                                        return Err(ExecutingMessageValidatorError::MinimumSafety(
+                                            safety,
+                                            got_safety_level,
+                                        ));
+                                    }
+                                }
+                            }
+                        }
```
emhane: think it's cleaner here to just match on the literals, i.e.
```rust
let got = if err.message().contains("safety level: safe") {
    SafetyLevel::Safe
} else if err.message().contains("safety level: local-safe") {
    SafetyLevel::LocalSafe
} else if ..;
```
and getting minimum with `err.message().split(' ').last().expect("non-empty string")`

emhane (commented):
nice progress

crates/external/rpc/src/traits.rs
```diff
@@ -6,20 +6,37 @@ use alloy_primitives::Log;
...
     SupervisorRpcError(#[from] ClientError),
 
     /// Message validation against the Supervisor took longer than allowed.
-    #[error("Message validation timed out: {0}")]
+    #[error("message validation timed out: {0}")]
     ValidationTimeout(#[from] Elapsed),
+
+    /// Catch-all variant for other supervisor errors.
+    #[error("unexpected error from supervisor: {0}")]
+    Other(#[from] ErrorObjectOwned),
```
emhane, line 39:
```suggestion
    /// Catch-all variant for supervisor server errors.
    #[error("unexpected error from supervisor: {0}")]
    SupervisorServerError(ErrorObjectOwned),
```

crates/external/rpc/src/traits.rs
```diff
@@ -6,20 +6,37 @@ use alloy_primitives::Log;
...
+        /// Level that was passed to supervisor
+        expected: SafetyLevel,
+    },
+    /// Invalid chain
+    #[error("unsupported chain id: {0}")]
+    UnknownChain(u64),
     /// Failure from the [`SupervisorApiClient`] when validating messages.
-    #[error("Supervisor determined messages are invalid: {0}")]
+    #[error("supervisor determined messages are invalid: {0}")]
     SupervisorRpcError(#[from] ClientError),
```
emhane, line 31: let's change this to `RpcClientError(ClientError)`. the name better reflects distinction with server errors, and the `ClientError` should not convert into this variant in case of safety level error.

crates/external/rpc/src/traits.rs
```diff
@@ -51,16 +68,60 @@ pub trait ExecutingMessageValidator {
...
+                                } else {
+                                    // Unexpected level name, return generic error
+                                    return Err(ExecutingMessageValidatorError::Other(err));
+                                };
+                            return Err(ExecutingMessageValidatorError::MinimumSafety {
+                                expected: safety,
+                                got: message_safety,
+                            });
+                        }
+                        Err(ExecutingMessageValidatorError::Other(err))
```
emhane, line 120: let's extract this code to `impl From<ErrorObjectOwned> for ExecutingMessageValidatorError`, and then add a test for the conversion

emhane (commented):
final comments

crates/external/rpc/src/traits.rs
```diff
@@ -6,20 +6,37 @@ use alloy_primitives::Log;
 use alloy_sol_types::SolEvent;
 use async_trait::async_trait;
 use core::time::Duration;
-use jsonrpsee::core::ClientError;
+use jsonrpsee::{core::ClientError, types::ErrorObjectOwned};
 use kona_interop::{CROSS_L2_INBOX_ADDRESS, ExecutingMessage, SafetyLevel};
 use tokio::time::error::Elapsed;
 
+const UNKNOWN_CHAIN_MSG: &str = "unknown chain: ";
+const MINIMUM_SAFETY_MSG: &str = "does not meet the minimum safety";
```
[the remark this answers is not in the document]
SozinM: Wasn't able to find source for `unknown message:`

SozinM:
@emhane Last fixes are here :)

crates/external/rpc/src/traits.rs
```diff
@@ -6,20 +6,41 @@ use alloy_primitives::Log;
...
+use jsonrpsee::{core::ClientError, types::ErrorObjectOwned};
 use kona_interop::{CROSS_L2_INBOX_ADDRESS, ExecutingMessage, SafetyLevel};
 use tokio::time::error::Elapsed;
 
+/// Derived from op-supervisor
+// todo: rm once resolved <https://github.com/ethereum-optimism/optimism/issues/14603>
+const UNKNOWN_CHAIN_MSG: &str = "unknown chain: ";
+/// Derived from [op-supervisor](https://github.com/ethereum-optimism/optimism/blob/4ba2eb00eafc3d7de2c8ceb6fd83913a8c0a2c0d/op-supervisor/supervisor/backend/backend.go#L479)
+// todo: rm once resolved <https://github.com/ethereum-optimism/optimism/issues/14603>
+const MINIMUM_SAFETY_MSG: &str = "does not meet the minimum safety";
```
emhane, line 18:
```suggestion
const MINIMUM_SAFETY_MSG: &str = "does not meet the minimum safety";

```
nitpick

crates/external/rpc/src/traits.rs
```diff
@@ -6,20 +6,41 @@ use alloy_primitives::Log;
...
 /// Failures occurring during validation of [`ExecutingMessage`]s.
 #[derive(thiserror::Error, Debug)]
 pub enum ExecutingMessageValidatorError {
+    /// Message does not meet minimum safety level
+    #[error("message does not meet safety level, got: {got}, expected: {expected}")]
```
emhane, line 23:
```suggestion
    #[error("message does not meet min safety level, got: {got}, expected: {expected}")]
```

crates/external/rpc/src/traits.rs
```diff
@@ -51,16 +72,117 @@ pub trait ExecutingMessageValidator {
...
+            } else if value.message().contains("safety safe") {
+                SafetyLevel::Safe
+            } else if value.message().contains("safety local-safe") {
+                SafetyLevel::LocalSafe
+            } else if value.message().contains("safety cross-unsafe") {
+                SafetyLevel::CrossUnsafe
+            } else if value.message().contains("safety unsafe") {
+                SafetyLevel::Unsafe
+            } else if value.message().contains("safety invalid") {
+                SafetyLevel::Invalid
```
emhane, line 131: this branch is redundant since all messages will be at least of invalid safety level

crates/external/rpc/src/traits.rs
```diff
@@ -51,16 +72,117 @@ pub trait ExecutingMessageValidator {
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_op_supervisor_error_parsing() {
+        let min_safety_error = r#"{"code":-32000,"message":"message {0x4200000000000000000000000000000000000023 4 1 1728507701 901} (safety level: unsafe) does not meet the minimum safety cross-unsafe"}"#;
+        let invalid_chain = r#"{"code":-32000,"message":"failed to check message: failed to check log: unknown chain: 14417"}"#;
+        let invalid_level = r#"{"code":-32000,"message":"message {0x4200000000000000000000000000000000000023 1091637521 4369 0 901} (safety level: invalid) does not meet the minimum safety unsafe"}"#;
+        let random_error = r#"{"code":-32000,"message":"gibberish error"}"#;
```
emhane, line 152: can these literals by constants?

emhane (approved):
sick! nitpicks

crates/external/rpc/src/traits.rs
SozinM: I could, bit it's tests 🤔

SozinM:
@emhane praying that all good 
image

PR #1164, opened 2025-03-03
chore(node): Move Engine into Crate

refcell:
### Description

Decouples the engine client from the `kona-node` to allow for parallelized work.

PR #1165, opened 2025-03-03
chore(node): Teeny Sync Crate

refcell:
### Description

For now, move sync types into a teeny crate to unblock dep graph.

Later, this could belong in a protocol crate :)

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -38,7 +38,7 @@ ignore:
   - "crates/utilities/serde/maili"
   # Node - Work in Progress
   - "bin/node"
-  - "crates/node/engine"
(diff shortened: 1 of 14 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1165 👈
* #1164
* `main`

clabby:
What could conflict here? idt we need to split it up into a new crate, should be able to just resolve any small conflicts as we go. Imagine this won't change much.

refcell:
> What could conflict here? idt we need to split it up into a new crate, should be able to just resolve any small conflicts as we go. Imagine this won't change much.

Sync types are used across the engine, cli, and others.

PR #1166, opened 2025-03-03
feat(node): Initial orchestration logic

clabby:
## Overview

Adds some simple abstractions + bootstraps the `node` module in the rollup node binary.

### Architecture

The `RollupNode` abstraction is intended to serve as the core orchestrator of the rollup node. Alone, it should be a programatic entry point to the rollup node binary's main mode of operation.

There are two core actor types defined in this PR:
- `Producer`s, which scan external data sources (such as RPC, ExEx events, or p2p gossip) and _create_ events for other actors to consume.
- `Actor`s, which consume events initially created by `Producer`s, and can:
    - Process incoming events
    - Mutate internal state
    - Pass off events for other `Actor`s to continue their silo of work.

For the sake of the initial iteration of the node, where sequencing isn't supported, the flow between the `Producer`s and `Actor`s will look something like:

```mermaid
flowchart LR
    DER["Derivation Actor"]
    A["L2 Sequencer Gossip Producer"] --> DER
    B["L1 Head Producer"] --> DER

   DER --> ENG["Engine Actor"]
```

### Future Work

This PR leaves a good few things unimplemented, but looking to get some feedback on the initial layout + iterate in follow up PRs. Getting the initial `RollupNodeBuilder` fully defined + hooked up to the CLI entrypoint, and then getting the `DerivationActor` implemented, are next-up.


bin/node/src/node/actors/mod.rs
```diff
@@ -0,0 +1,5 @@
+//! [NodeActor] services for the node.
+//!
+//! [NodeActor]: super::NodeActor
+
+mod derivation;
```
refcell, line 5: inaccessible

refcell (approved):

PR #1170, opened 2025-03-04
feat(protocol): Update `RollupConfig`

ratankaliani:
## Overview
Kona's `RollupConfig` is not up to date with `rollup.Config` in OP Geth. This causes compatibility issues for downstream projects such as OP Succinct which attempt to deserialize the response from `optimism_rollupConfig` into the `RollupConfig` type exposed from `op-alloy-protocol`.

This PR brings the `RollupConfig` up to date with the type from OP Geth by adding `chain_op_config` and `alt_da_config`.

## Context
op-geth: https://github.com/ethereum-optimism/optimism/blob/f9ee04cebf7663d2b3033eeb4d48fd0cbc2b674f/op-node/rollup/types.go#L68-L152

kona: https://github.com/op-rs/kona/blob/8af899597dde99ba6a5fd31eae2487c74226d3b5/crates/protocol/genesis/src/rollup.rs#L35-L99


ratankaliani:
@clabby Two questions:
1. In `op-geth`, the `chain_op_config` field is optional, should it be optional here as well?
2. Is it necessary to add more blocks to the static test harness [here](https://github.com/op-rs/kona/blob/263c8a03bf7ee1b9b6d37fce1250eac91027c009/crates/proof/executor/src/executor/mod.rs)? I needed to update this as the previous blocks had a RollupConfig incompatible with the new one.

crates/proof/executor/src/executor/mod.rs
```diff
@@ -534,13 +534,9 @@ mod test {
...
-    #[case::small_block(22884230)]
-    #[case::small_block_2(22884231)]
-    #[case::small_block_3(22880574)]
-    #[case::small_block_4(22887258)]
-    #[case::medium_block(22886464)]
-    #[case::medium_block_2(22886311)]
-    #[case::medium_block_3(22880944)]
+    #[case::small_block(10311000)] // Unichain Mainnet
+    #[case::small_block_2(10211000)] // Unichain Mainnet
+    #[case::small_block_3(10215000)] // Unichain Mainnet
```
refcell, line 539: @clabby what are your thoughts here?

refcell: Do we have a guide yet on how to produce this data? Would be super helpful.

refcell (approved):

crates/proof/executor/src/executor/mod.rs
clabby: Yeah we really should generate more cases here (and delete the old ones - I see all of these deleted ones still exist in `testdata`.) It's totally okay to cycle these out every now and then, though, since it's impossible to regenerate these old cases without a blob archiver.

There's a short guide in the commented-out test above, regenerating is pretty easy:
https://github.com/op-rs/kona/blob/263c8a03bf7ee1b9b6d37fce1250eac91027c009/crates/proof/executor/src/executor/mod.rs#L524-L534

refcell (changes requested):
GTG pending added block fixtures

crates/proof/executor/src/executor/mod.rs
```diff
@@ -534,13 +534,9 @@ mod test {
     // }
 
     #[rstest]
-    #[case::small_block(22884230)]
-    #[case::small_block_2(22884231)]
-    #[case::small_block_3(22880574)]
-    #[case::small_block_4(22887258)]
-    #[case::medium_block(22886464)]
-    #[case::medium_block_2(22886311)]
-    #[case::medium_block_3(22880944)]
```
clabby, line 543: Should delete all of these from `testdata`.

clabby (changes requested):

clabby (approved):

refcell (approved):

refcell:
Thanks @ratankaliani !

PR #1171, opened 2025-03-04
feat(interop): Dedup logic for parsing `Log` to `ExecutingMessage`

emhane:


clabby (approved):
Change looks good; Just needs lints

PR #1172, opened 2025-03-04
feat(interop): Clean up interop validator RPC component

emhane:
Cleans up interop validator rpc component logic

- Moves `reqwest` supervisor client into `kona-rpc`
- Removes non-interop related trait methods from supervisor api
- Renames supervisor api to better reflect it's a subset of supervisor api to `CheckMessages`
- Defines new error `InvalidExecutingMessage` from subset of `ExecutingMessageValidatorError` logic


clabby:
Nice! I see a couple of TODOs in here - do you want to get this reviewed + merged in the meantime, or wait for those upstream tickets to get resolved?

emhane:
@clabby usually I add todos to unblock merging, just for next dev to know that "we are aware this is an anti-pattern" or similar, try to also link todos to some issue so that any dev can see WHEN it's time to fix the todo -> when the linked issue is closed

emhane:
seems like codecov doesn't run with all features enabled, because the lines it says aren't covered, are indeed covered in `kona-rpc/interop`. could that be it @refcell @clabby ?

clabby:
> seems like codecov doesn't run with all features enabled, because the lines it says aren't covered, are indeed covered in `kona-rpc/interop`. could that be it @refcell @clabby ?

Correct; Codecov currently isn't running with all features (just `test-utils`.) Feel free to enable this in `rust_ci.yml`, in the `coverage` job.

crates/node/rpc/src/jsonrpsee.rs
```diff
@@ -0,0 +1,173 @@
...
+#[cfg(feature = "client")]
+pub use op_alloy_rpc_jsonrpsee::traits::{MinerApiExtClient, OpAdminApiClient};
+
+/// Optimism specified rpc interface.
+///
+/// https://docs.optimism.io/builders/node-operators/json-rpc
+/// https://github.com/ethereum-optimism/optimism/blob/8dd17a7b114a7c25505cd2e15ce4e3d0f7e3f7c1/op-node/node/api.go#L114
+#[cfg_attr(not(feature = "client"), rpc(server, namespace = "optimism"))]
+#[cfg_attr(feature = "client", rpc(server, client, namespace = "optimism"))]
+pub trait RollupNode {
```
clabby, line 32: Small nit, can we rename this to `RollupNodeApi` or something to the effect of that?

We've a struct named `RollupNode` for the core service over in `bin/node`

clabby (approved):
Looks great. Thank you!

crates/node/rpc/src/jsonrpsee.rs
emhane: out of scope see

PR #1176, opened 2025-03-04
chore(protocol): RPC Block -> L2BlockInfo

refcell:
### Description

Refactors conversion from an RPC block into an `L2BlockInfo`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,6 +28,9 @@ default-members = [
 # https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html
 resolver = "2"
 
+[patch.crates-io]
+op-alloy-consensus = { git = "https://github.com/alloy-rs/op-alloy", branch = "rf/as-ref-tx-env" }
+
 [workspace.lints.rust]
 missing-debug-implementations = "warn"
 missing-docs = "warn"
@@ -123,6 +126,7 @@ alloy-rpc-types = { version = "0.11.1", default-features = false }
 alloy-rpc-client = { version = "0.11.1", default-features = false }
 alloy-primitives = { version = "0.8.21", default-features = false }
 alloy-node-bindings = { version = "0.11.1", default-features = false }
+alloy-rpc-types-eth = { version = "0.11.1", default-features = false }
 alloy-transport-http = { version = "0.11.1", default-features = false }
 alloy-rpc-types-engine = { version = "0.11.1", default-features = false }
 alloy-rpc-types-beacon = { version = "0.11.1", default-features = false }
@@ -133,6 +137,7 @@ op-alloy-flz = { version = "0.10.8", default-features = false }
 op-alloy-network = { version = "0.10.8", default-features = false }
 op-alloy-provider = { version = "0.10.8", default-features = false }
 op-alloy-consensus = { version = "0.10.8", default-features = false }
+op-alloy-rpc-types = { version = "0.10.8", default-features = false }
 op-alloy-rpc-jsonrpsee = { version = "0.10.8", default-features = false }
 op-alloy-rpc-types-engine = { version = "0.10.8", default-features = false }
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -28,9 +28,6 @@ default-members = [
 # https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html
 resolver = "2"
 
-[patch.crates-io]
-op-alloy-consensus = { git = "https://github.com/alloy-rs/op-alloy", branch = "main" }
-
 [workspace.lints.rust]
 missing-debug-implementations = "warn"
 missing-docs = "warn"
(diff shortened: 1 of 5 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1176 👈
* `main`

crates/protocol/protocol/src/block.rs
```diff
@@ -143,6 +149,43 @@ impl L2BlockInfo {
         Self { block_info, l1_origin, seq_num }
     }
 
+    /// Constructs an [L2BlockInfo] from a given [alloy_rpc_types_eth::Block] and [ChainGenesis].
+    pub fn from_rpc_block_and_genesis(
```
clabby, line 153: Instead of introducing a new function specific to the rpc block, can we convert between the RPC block and the consensus block? This should work:

```rs
let consensus_block = self
    .l2_provider
    .get_block(
        BlockNumberOrTag::Safe.into(),
        BlockTransactionsKind::Full,
    )
    .await?
    .map(|b| b.into_consensus().map_transactions(|b| b.inner.inner))?;
```

crates/protocol/protocol/src/block.rs
```diff
@@ -37,19 +37,25 @@ impl BlockInfo {
     }
 }
 
-impl<T> From<Block<T>> for BlockInfo {
-    fn from(block: Block<T>) -> Self {
-        Self::from(&block)
+impl From<alloy_rpc_types_eth::Block> for BlockInfo {
```
mattsse, line 40: you likely want to relax the tx type here

crates/protocol/protocol/src/block.rs
mattsse: actually let me check if we have that conversion on the rpc type, if not I'll add

crates/protocol/protocol/src/block.rs
```diff
@@ -37,19 +37,25 @@ impl BlockInfo {
...
+    fn from(block: alloy_rpc_types_eth::Block) -> Self {
+        Self::from(&block.header.inner)
```
mattsse, line 42: I believe you could also reuse the hash in the response here

mattsse (commented):
there are a few more nicer conversions coming on the rpc types in the next alloy minor by eow

crates/protocol/protocol/src/block.rs
refcell: Thanks @mattsse - I didn't see it so this is why i just went with the dup for now. Was going to raise this as an issue if not :)

crates/protocol/protocol/src/block.rs
refcell: Definitely good callout

refcell:
Blocked until the new release is out where we can use the rpc -> consensus convenience helper

refcell:
🧊

crates/protocol/protocol/src/block.rs
```diff
@@ -144,8 +143,8 @@ impl L2BlockInfo {
     }
 
     /// Constructs an [L2BlockInfo] from a given OP [Block] and [ChainGenesis].
-    pub fn from_block_and_genesis(
-        block: &OpBlock,
+    pub fn from_block_and_genesis<T: Typed2718>(
+        block: &Block<T>,
```
refcell, line 147: So this will still error because the `T` type doesn't implement `as_deposit()`.

Since the `TransactionResponse` of the `Optimism` network trait is `op_alloy_rpc_types::Transaction` which only internally contains `OpTxEnvelope`, there's no `as_deposit` available.

I _think_ we can get around this by restricting `T` to `AsRef<OpTxEnvelope>` and then implement it on `OpTxEnvelope`.

I've opened a PR (https://github.com/alloy-rs/op-alloy/pull/464) in `op-alloy` to do this cc @mattsse.

Cargo.toml
```diff
@@ -28,6 +28,9 @@ default-members = [
 # https://doc.rust-lang.org/edition-guide/rust-2021/default-cargo-resolver.html
 resolver = "2"
 
+[patch.crates-io]
+op-alloy-consensus = { git = "https://github.com/alloy-rs/op-alloy", branch = "rf/as-ref-tx-env" }
+
```
refcell, line 33: This needs to be fixed once a release is cut with https://github.com/alloy-rs/op-alloy/pull/464

refcell: The `cargo-deny` ci job will fail until this is removed

crates/node/engine/src/client.rs
```diff
@@ -76,41 +76,9 @@ impl EngineClient {
...
-                anyhow::bail!("First transaction is not a deposit");
-            };
-
-            let l1_info = L1BlockInfoTx::decode_calldata(tx.input.as_ref())
-                .map_err(|e| anyhow::anyhow!(e))?;
-            (l1_info.id(), l1_info.sequence_number())
-        };
-
-        Ok(L2BlockInfo { block_info, l1_origin, seq_num: sequence_number })
+        let block = block.into_consensus();
```
clabby, line 79: nit: can just `ok_or_else(|| ...)?.into_consensus()`

crates/protocol/protocol/src/block.rs
```diff
@@ -161,7 +161,7 @@ impl L2BlockInfo {
             }
 
             let tx = &block.body.transactions[0];
-
+            let tx: &OpTxEnvelope = tx.as_ref();
```
clabby, line 164: nit: Can join L163+164

clabby (approved):

PR #1179, opened 2025-03-05
feat(node): Hook up `RollupNodeBuilder` to CLI

clabby:
## Overview

Hooks up the `RollupNodeBuilder` to the CLI.

PR #1180, opened 2025-03-05
feat(node): Derivation actor

clabby:
## Overview

Introduces the `Actor` trait implementation for the derivation component of the rollup node. Attempts to continue derivation every time the L2 chain's forkchoice has been updated or new data is potentially available on L1.

In the meantime, it has its own copy of the L2 safe head - This will be removed once we have the engine `Actor` + the state `Actor`.


clabby:
## 📚 $\text{Stack Overview}$

* #1180 👈
* #1179
* `main`

bin/node/src/node/actors/derivation.rs
```diff
@@ -2,35 +2,231 @@
...
+                _ = self.cancellation.cancelled() => {
+                    info!(
+                        target: "derivation",
+                        "Received shutdown signal. Exiting derivation task."
+                    );
+                    return Ok(());
+                }
+                msg = self.receiver.recv() => {
+                    self.process(msg?).await?;
+                }
```
refcell, line 177: This only makes sense to block here since the channel can fill up and tell the caller if it is ready to receive new messages?

refcell (approved):

PR #1192, opened 2025-03-05
fix(protocol): Use `Prague` blob fee calculation for L1 info tx

clabby:
## Overview

Introduces a fix for the `L1BlockInfoTx`, where we weren't using the `BlobParams::prague()` when crafting the L1 block info transaction.


refcell (approved):

Inphi:
be careful not to release this change until the optional fork config is in.

PR #1194, opened 2025-03-05
feat(engine): Insert Task Updates

refcell:
### Description

Updates the insert unsafe payload task.

I'm starting to think messaging out engine api calls from the individual engine tasks doesn't make sense.

Decoupling the state from the engine controller seems like the right abstraction, but further decoupling of the engine tasks and the engine api doesn't seem warranted.

Will follow up to give engine tasks a reference to the `EngineClient` so they can perform engine api requests internal to the task. This way, the only messages that should
be needed from engine tasks are updates to and requests for the engine state.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3845,6 +3845,7 @@ dependencies = [
  "http-body-util",
  "kona-genesis",
  "kona-protocol",
+ "kona-rpc",
  "op-alloy-network",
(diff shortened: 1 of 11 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1194 👈
* `main`

PR #1195, opened 2025-03-06
fix(protocol): Add optional pectra blob fee schedule fork

clabby:
## Overview

> [!WARNING]
> ~~Blocked by https://github.com/ethereum-optimism/superchain-registry/pull/926~~

Adds support for an optional Pectra blob fee schedule fork.

**Context**
> There was an incident on OP Stack Sepolia chains (03-05-2025) when L1 activated pectra, where the sequencer followed the incorrect chain, using the legacy Cancun blob fee schedule instead of the new Prague blob fee schedule. This portion of the chain was chosen to be canonicalized in favor of the prospect of a deep reorg imposed by the sequencers of the testnet chains. An optional hardfork was introduced for Sepolia only, where if present, activates the use of the Prague blob fee schedule. If the hardfork is not present, and L1 has activated pectra, the Prague blob fee schedule is used immediately.


refcell (approved):
Very nice

PR #1197, opened 2025-03-06
feat(engine): Insert New Payload

refcell:
### Description

- Updates forkchoice and insert tasks to perform engine api calls through client inside the task instead of messaging as described in #1194.
- Implements a bunch of the new payload insert logic - the type argument is taken as the output from derivation which doesn't line up with the inputs to the engine_newPayload methods though so that will need to be cleaned up.

### DIFF
--- a/crates/node/engine/src/tasks/insert/task.rs
+++ b/crates/node/engine/src/tasks/insert/task.rs
@@ -1,6 +1,7 @@
 //! A task to insert an unsafe payload into the execution engine.
 
 use alloy_eips::BlockNumberOrTag;
+use alloy_primitives::B256;
 use alloy_provider::ext::EngineApi;
 use alloy_rpc_types_engine::{ExecutionPayload, ExecutionPayloadInputV2, ForkchoiceState};
 use kona_protocol::L2BlockInfo;
@@ -10,11 +11,12 @@ use std::sync::Arc;
 
 use crate::{
     EngineClient, EngineState, EngineTask, InsertTaskError, InsertTaskInput, InsertTaskOut,
-    SyncStatus,
+    SyncConfig, SyncStatus,
 };
 
 /// The input type for the [InsertTask].
-type Input = (Arc<EngineClient>, OpNetworkPayloadEnvelope, L2BlockInfo);
+/// The second argument is the genesis l2 hash expected to be provided by the rollup config.
+type Input = (Arc<EngineClient>, SyncConfig, B256, OpNetworkPayloadEnvelope, L2BlockInfo);
 
 /// An external handle to communicate with a [InsertTask] spawned
 /// in a new thread.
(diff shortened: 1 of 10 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1197 👈
* #1194
* `main`

crates/node/engine/src/tasks/insert/task.rs
```diff
@@ -97,7 +119,87 @@ impl EngineTask for InsertTask {
             }
         }
 
-        // TODO: Insert Payload in Engine API
+        let (client, version, attributes, block_info) = input;
+        let response = match version {
+            EngineNewPayloadVersion::V2 => {
```
clabby, line 124: This is incorrect - `engine_newPayload` should be called after an `engine_forkchoiceUpdate`, which takes a forkchoice state + optional attributes (if we're building a new block.) We can't convert from a payload attributes to a sealed `ExecutionPayload`, since we need the EL to execute the `PayloadAttributes` on top of the parent state to generate the sealed `ExecutionPayload`.

The engine needs to handle a few situations:
1. Gossiped `ExecutionPayload` from sequencer
    - If the EL already has the block, check equivalence, and mark the block as `unsafe` without sending to the EL if the parameters match up. No FCU needed, since the chain already has the block.
    - If the EL doesn't have the block, send an `engine_newPayload` with the full signed `ExecutionPayload`, and mark the block as `unsafe` if `engine_getPayload` shows that it was imported successfully. FCU needed after the fact to update the new forkchoice and equivocate.
1. New `OpAttributesWithParent` from derivation
    - If the EL already has the block, [check for _parameter_ equivalence](https://specs.optimism.io/protocol/derivation.html#l1-consolidation-payload-attributes-matching), and mark the block as `safe` without sending to the EL if the parameters match up. FCU needed after to update the `safe` block on the EL.
    - If the EL doesn't have the block, send an `engine_forkchoiceUpdate` with the attributes set as the second parameter, and use `engine_getPayload` to check for payload validity.

clabby (changes requested):

crates/node/engine/src/tasks/insert/task.rs
refcell: I think only the argument is incorrect here.

See: https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/engine/engine_controller.go#L368

clabby: We should have the entire `ExecutionPayload` sent to us by the sequencer; We should never have to construct an `ExecutionPayloadVX` ourselves in the rollup node. Note that in the impl you linked, `InsertUnsafePayload` is called every time the sequencer sends a full signed `ExecutionPayload` (i.e., they've already executed the block.) https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/driver/state.go#L252

crates/node/engine/src/tasks/insert/task.rs
```diff
@@ -1,10 +1,29 @@
...
+/// The input type for the [InsertTask].
+/// The third argument is the genesis l2 hash expected to be provided by the rollup config.
+type Input = (
+    Arc<EngineClient>,
+    SyncConfig,
+    B256,
+    EngineForkchoiceVersion,
+    OpNetworkPayloadEnvelope,
+    L2BlockInfo,
+);
```
clabby, line 26: no need to get in this PR, but slightly prefer a struct w/ named fields

clabby (approved):
Nice - can get this in and keep iterating 💯

PR #1201, opened 2025-03-06
feat(engine): Functional Tasks

refcell:
### Description

Plays around with making engine tasks functional.

Not sure I'm a fan of this pattern since we pass around mutable references,
although it cleans up the `EngineTask` trait, removing a mutable self ref.

Open to comments + opinions.

### DIFF
--- a/crates/node/engine/src/tasks/forkchoice/task.rs
+++ b/crates/node/engine/src/tasks/forkchoice/task.rs
@@ -3,64 +3,62 @@
 use op_alloy_provider::ext::engine::OpEngineApi;
 use std::sync::Arc;
 
+use tokio::{
+    sync::broadcast::{Receiver as TokioReceiver, Sender as TokioSender},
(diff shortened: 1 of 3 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1201 👈
* #1197
* `main`

clabby (commented):
I sort of like the task having its own set of communication handles semantically, but curious if this has caused any pain in the crate so far.

In my head the handles being owned by the task, and dropped along with the task upon completion, is a nice encapsulation of concerns.

refcell:
> I sort of like the task having its own set of communication handles semantically, but curious if this has caused any pain in the crate so far.
> 
> In my head the handles being owned by the task, and dropped along with the task upon completion, is a nice encapsulation of concerns.

I agree - will close this PR out for now and can always come back to it. Mostly just wanted to test out what it would feel like having a functional interface

PR #1202, opened 2025-03-06
feat(engine): Wire up Insert Task

refcell:
### Description

Wires up the `InsertTask` into the engine actor.

Notice, since we previously removed the engine client decoupling, passing in the engine client reference directly to the task, we can now removing custom msg processing in the engine actor. Instead, we just pass along the message for the orchestrator to handle.


refcell:
## 📚 $\text{Stack Overview}$

* #1202 👈
* #1197
* `main`

clabby (approved):
Neat stuff. LGTM

PR #1203, opened 2025-03-06
feat(providers-alloy): Refactor `AlloyChainProvider`

clabby:
## Overview

Refactors the `AlloyChainProvider` to use functions from the `alloy_provider::Provider` trait to fetch blocks. Since we wrote this, APIs to convert from RPC types -> consensus types have improved.


refcell (approved):
Very nice mr allen

PR #1208, opened 2025-03-07
chore: bump alloy 0.12

mattsse:
how should I bump the reth dep?

to main?


refcell:
> how should I bump the reth dep?
> 
> to main?

If there's a tag you could use that would be great - reth doesn't have published version yet right?

mattsse:
not yet, scheduled for ~tuesday

mattsse:
actually let me try to make this work here and then simplifications after we have a tag

mattsse:
okay, last issue is here:

https://github.com/mattsse/kona/blob/cbe22e4e91f1b7ce805727fe1b0b8cef224f1a18/bin/host/src/single/handler.rs#L350-L353

because this was changed to just vec on request: https://github.com/alloy-rs/alloy/pull/2059

mattsse:
okay after reading that snippet, I think we just need to remove the preimage check

bin/host/src/interop/handler.rs
```diff
@@ -268,7 +269,8 @@ impl HintHandler for InteropHintHandler {
 
                 let Block { transactions, .. } = providers
                     .l2(&chain_id)?
-                    .get_block_by_hash(hash, BlockTransactionsKind::Full)
+                    .get_block_by_hash(hash)
+                    .full()
```
refcell, line 273: This is a nice api

refcell: Is it hashes by default?

mattsse:
the client jobs are failing with a oncecell critical error issue, idk what these images do but this error is thrown if features are not supported by the target afaik.
is this running no-std, riscv?

cc @klkvr

bin/host/src/interop/handler.rs
mattsse: yes

crates/providers/providers-local/src/chain_provider.rs
```diff
@@ -136,7 +136,7 @@ impl InMemoryChainProviderInner {
                     blob_gas_used: header.blob_gas_used,
                     excess_blob_gas: header.excess_blob_gas,
                     parent_beacon_block_root: header.parent_beacon_block_root,
-                    requests_hash: header.requests_hash(),
+                    requests_hash: header.requests_hash,
                 },
```
refcell, line 140: Should this just be an `Into::into()` too? Does reth Chain header implement `Into<alloy_consensus::Header>`

refcell (commented):
Few questions, otherwise very very nice api changes

crates/node/engine/src/client.rs
```diff
@@ -2,7 +2,6 @@
 
 use alloy_eips::eip1898::BlockNumberOrTag;
 use alloy_network::AnyNetwork;
-use alloy_network_primitives::BlockTransactionsKind;
```
refcell, line 5: Since this was removed, we need to remove the `alloy_network_primitives` from the cargo manifest so unused crate linting passes

refcell:
The protocol block type is failing in tests because it expects a recovered tx now. I've added support to do this conversion in op-alloy here: https://github.com/alloy-rs/op-alloy/pull/467

refcell:
I've fixed up the linting and testing issues - should just be the client ci jobs now

Cargo.lock
```diff
@@ -122,31 +122,52 @@ version = "0.11.1"
...
+ "alloy-rlp",
+ "alloy-serde 0.12.3",
+ "alloy-trie",
+ "arbitrary",
+ "auto_impl",
+ "c-kzg",
+ "derive_more 2.0.1",
+ "either",
  "k256",
+ "once_cell",
```
refcell, line 153: Looks like ci is failing because `once_cell` was [introduced in alloy-consensus](https://github.com/alloy-rs/alloy/blob/main/crates/consensus/Cargo.toml#L56C1-L56C10) with "critical-section" feature enabled.

mattsse: should be fixed after https://github.com/alloy-rs/alloy/pull/2165

refcell (approved):

PR #1213, opened 2025-03-07
chore(node-rpc): Delete dead code

clabby:
## Overview

Deletes dead code in the `kona-rpc` crate. `api.rs` is the exact same as `jsonrpsee.rs`, and looks to have snuck in.


refcell (approved):
weird this must have been product of a bad rebase

PR #1224, opened 2025-03-07
fix(executor): Dep on kona-host

### LINKED ISSUE #1219: bug(executor): Remove Kona Host Dependency
### Description

Kona's executor crate depends on kona-host in it's [dev dependency section](https://github.com/op-rs/kona/blob/main/crates/proof/executor/Cargo.toml#L52), disallowing the crate from being published since `kona-host` is set to not publish.

`kona-host` needs to be removed from `kona-executor`'s dev dependencies so crates like `kona-driver` that depend on `kona-executor` may also be published.

refcell:
### Description

Removes `kona-host` as a dependency of `kona-executor`.

Closes #1219


refcell:
## 📚 $\text{Stack Overview}$

* #1224 👈
* `main`

clabby (approved):

PR #1231, opened 2025-03-10
feat(node): Refactor orchestration

clabby:
## Overview

Refactors the current `kona-node` orchestrator into it's own crate, `kona-node-service`. This will serve as the programatic entry point for the node.

### Node Service Traits

Also introduces a few new traits for creating custom configurations of the rollup node: `RollupNodeService`, `ValidatorNodeService`, and `SequencerNodeService`.

The `ValidatorNodeService` is the core service implementation, enabling support for syncing the L2 chain via `unsafe` gossip and `safe` block derivation. This trait can be implemented to define:
* What DA watcher to use (ETH DA [rpc], ETH DA [reth], and alt DA)
* What derivation pipeline to use (for mods like alt DA, or to use the node with a different set of provider trait impls.)

The `SequencerNodeService` is an extension of `ValidatorNodeService`, overriding `start`. This is yet-to-be implemented, but should:
* Replace the sequencer p2p gossip in `NodeActor` with a sign + gossip-out actor.
* Start a block-building task actor
* Re-use the rest of the `ValidatorNodeService`.

Finally, the `RollupNodeService` is an extension trait of both of the above traits, choosing the entrypoint based off of the operating mode.

### Standard Config

The `kona-node-service` crate also serves a standard implementation of the `RollupNodeService` trait, configured with to-spec components for the governance approved OP Stack.

### Global Channel -> Topic-based channels

The global `broadcast` channel that sent a shared event type to all `NodeActor`s in the system has been removed in favor of the node using topic-based channels. This should prevent churn in the actors, where they have to ingest all events from all actors in their respective event loops. Instead, we will now use targeted topic-based channels to streamline specific information sharing routes.


crates/protocol/derive/src/traits/pipeline.rs
```diff
@@ -1,14 +1,13 @@
...
+use crate::{errors::PipelineErrorKind, traits::OriginProvider, types::StepResult};
 use alloc::boxed::Box;
 use async_trait::async_trait;
 use core::iter::Iterator;
 use kona_genesis::{RollupConfig, SystemConfig};
 use kona_protocol::L2BlockInfo;
 use kona_rpc::OpAttributesWithParent;
 
-use crate::{errors::PipelineErrorKind, traits::OriginProvider, types::StepResult};
-
```
refcell, line 11: Much prefer the previous formatting personally. Clear separation of external vs local types.

crates/node/service/src/actors/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+    type InboundEvent;
+    /// The error type for the actor.
+    type Error;
+
+    /// Starts the actor.
+    async fn start(self) -> Result<(), Self::Error>;
+
+    /// Processes an incoming message.
+    async fn process(&mut self, msg: Self::InboundEvent) -> Result<(), Self::Error>;
+}
```
refcell, line 31: Think I prefer trait definitions in their own trait module to make it clear to the file system enjoyer where things live

refcell: In the ideal case, mods contain nothing but re-exports

clabby: Yeah sounds good - figured it was small enough to not warrant a new file, but not super opinionated

crates/node/service/src/service/core.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// Starts the rollup node service, using the entry point respective to the [NodeMode].
+    ///
+    /// - [NodeMode::Validator]: [ValidatorNodeService::start]
+    /// - [NodeMode::Sequencer]: [SequencerNodeService::start]
+    async fn start(&self) -> Result<(), Self::Error> {
+        info!(target: "rollup_node", "Starting rollup node services");
+        info!(target: "rollup_node", "Chain ID: {}", self.config().l2_chain_id);
+        for hf in self.config().hardforks.to_string().lines() {
+            info!(target: "rollup_node", "{hf}");
+        }
```
refcell, line 34: This is just to print each display line on a new log line?

clabby: Yeah - mainly to keep everything in tracing rather than a raw write to stdout. Prints out the list of hard fork activation times configured on startup.

crates/node/service/src/service/util.rs
```diff
@@ -0,0 +1,34 @@
...
+/// [JoinSet]: tokio::task::JoinSet
+/// [NodeActor]: crate::NodeActor
+macro_rules! spawn_and_wait {
+    ($cancellation:expr, actors = [$($actor:expr$(,)?)*]) => {
+        let mut task_handles = tokio::task::JoinSet::new();
+
+        $(
+            task_handles.spawn(async move {
+                if let Err(e) = $actor.start().await {
+                    // TODO: Bubble up generic error.
```
refcell, line 16: Shouldn't this just return the error so the `join_next` call below will pick it up and gracefully shutdown?

clabby: Yes it should. Adding in a type erased error type in a follow up PR, goal here is to allow the actors to have separate error defs but exist in the same joinset.

crates/node/service/src/service/validator.rs
```diff
@@ -0,0 +1,97 @@
...
+///    transaction batches + deposits) are then derived from the DA layer.
+/// 2. The L2 sequencer, which produces unsafe L2 blocks and sends them to the network over p2p
+///    gossip.
+///
+/// From these two sources, the validator node imports `unsafe` blocks from the L2 sequencer and
+/// `safe` blocks from the L2 derivation pipeline into the L2 execution layer via the Engine API.
+///
+/// Finally, a state actor listens for new L2 block import events and updates the L2 state
+/// accordingly, sending notifications to the other actors for synchronization.
+///
```
refcell, line 26: We'll probably also want to make p2p out generic so we can sub in a noop impl if a node configs no gossip out

crates/node/service/src/service/core.rs
refcell: Very clean

refcell:
Maybe we should add `crates/node` to the ignore section in `.github/codecov.yaml` for now if it's not already.

PR #1250, opened 2025-03-12
feat(examples): Pulls Discovery and P2P Gossip into Examples

refcell:
### Description

Previously, discovery and gossip were set up as subcommands of the `kona-node` for testing.

This PR pulls both services into standalone examples so the `kona-node` can approach a more production state where there are no subcommands.
This allows the `kona-node` ux to be more similar to the `op-node`.


refcell:
## 📚 $\text{Stack Overview}$

* #1250 👈
* `main`

clabby (approved):
Nice

PR #1251, opened 2025-03-12
chore(node): Simplify Node CLI

refcell:
### Description

Since gossip and disc services are pulled into examples, there's no need to have a node subcommand.

In the future, the `kona-node` should run the node by default like the `op-node` and optionally have subcommands as well.

This PR simplifies the structure of the `kona-node` cli while also allowing it to be extended with new subcommands, splitting shared cli flags into a flags module.


refcell:
## 📚 $\text{Stack Overview}$

* #1251 👈
* #1250
* `main`

clabby (approved):
Not opposed for now - we might want some debug subcommands, but we can always add those back.

bin/node/src/commands/node.rs
```diff
@@ -84,13 +85,29 @@ impl NodeCommand {
...
+            .l1_beacon
+            .clone()
+            .ok_or(anyhow::anyhow!("Missing L1 beacon URL. Please specify with `--l1.beacon`."))?;
+        let l2_provider_rpc = self.l2_provider_rpc.clone().ok_or(anyhow::anyhow!(
+            "Missing L2 provider RPC URL. Please specify with `--l2.provider`."
+        ))?;
+        let l2_engine_rpc = self
+            .l2_engine_rpc
+            .clone()
+            .ok_or(anyhow::anyhow!("Missing L2 engine RPC URL. Please specify with `--l2`."))?;
```
clabby, line 102: We should be letting `clap` handle this IMO - non-optional flags should already have a nice "flag `x` is required" type message.

PR #1255, opened 2025-03-13
chore: Remove B256 Value Parser

refcell:
### Description

Small PR to remove the `B256` value parser that uses the default `B256` `FromStr` parsing.

Also fixes the interop cli config which required the `--l2-node-address` flag which DNE.


refcell:
## 📚 $\text{Stack Overview}$

* #1255 👈
* `main`

clabby (approved):
Oh very cool - didn't know they added this.

PR #1261, opened 2025-03-13
Additional errors for cross tx validation

SozinM:


crates/node/rpc/src/interop/error.rs
```diff
@@ -19,6 +19,14 @@ pub enum InteropTxValidatorError {
     #[error("message validation timed out, timeout: {0} secs")]
     ValidationTimeout(u64),
 
+    /// Interop hardfork is not active, all cross chain txs are invalid
+    #[error("interop hardfork has not been activated")]
+    InteropInactive,
```
emhane, line 24: should feature gate call to supervisor behind check on chainspec if interop is active, so this isn't needed

emhane (commented):
these error variants are redundant

PR #1267, opened 2025-03-14
fix(executor): Use correct empty `sha256` hash

clabby:
## Overview

Really dumb bug - corrects the empty `sha256` hash in `kona-executor` and updates the action tests to the latest revision of the monorepo.

verify with: `sha256sum /dev/null`.


crates/proof/executor/src/constants.rs
```diff
@@ -16,4 +16,4 @@ pub(crate) const HOLOCENE_EXTRA_DATA_VERSION: u8 = 0x00;
 
 /// Empty SHA-256 hash.
 pub(crate) const SHA256_EMPTY: B256 =
-    b256!("c5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470");
```
clabby, line 19: This is `keccak256("")` 🤦

refcell (approved):

crates/proof/executor/src/constants.rs
refcell: i_believe_you.jpeg

crates/proof/executor/src/constants.rs
```diff
@@ -16,4 +16,4 @@ pub(crate) const HOLOCENE_EXTRA_DATA_VERSION: u8 = 0x00;
...
+    b256!("e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855");
```
mattsse, line 19: since the holesky snafu I know this one quite well -.- 

this is indeed sha256("")

mattsse (approved):

PR #1268, opened 2025-03-14
fix(isthmus): Disallow `to` bit to be set for EIP-7702 txs

clabby:
## Overview

Disallows the `to` bit to be set in EIP-7702 transactions - this bit is only set when the transaction is a contract creation, and EIP-7702 mandates the `to` field to be present.


refcell (approved):

clabby:
@refcell actually already has us covered here @ https://github.com/op-rs/kona/blob/7c2751901e9e297e21bfb4ff46d414bee05dec37/crates/protocol/protocol/src/batch/tx_data/wrapper.rs#L164-L166 - missed when looking over. Closing this one.

PR #1235, opened 2025-03-10
chore: Cleanup and Dependency Bumps

refcell:
### Description

General cleanup and dependency bumps.

PR #1276, opened 2025-03-17
test: `rocksdb = 0.23.0`

clabby:
## Overview

Just testing if CI is playing nicely with `rocksdb` v0.23.0


refcell:
Fixed in https://github.com/op-rs/kona/pull/1235

PR #1279, opened 2025-03-19
chore(workspace): Remove `reth` dependency

clabby:
## Overview

Removes the `reth` dependency in the repository, so that we have an easier time using `kona` crates in `op-reth`. The only reason we need the `reth` dep in the repo right now is for the (in-progress) ExEx version of the rollup node.

Instead, we'll probably just ship that version in another repo. We can use the rollup node lib + reth's node builder to assemble an L1 reth binary that has the exex installed there.


refcell (approved):
We'll also want to remove `kona-providers-local` from the manifest.

https://github.com/op-rs/kona/blob/main/Cargo.toml#L92

PR #1281, opened 2025-03-19
Add FromStr support for SafetyLevel

gitToki:


crates/protocol/interop/src/safety.rs
```diff
@@ -19,6 +21,34 @@ pub enum SafetyLevel {
...
+            "unsafe" => Ok(Self::Unsafe),
+            "invalid" => Ok(Self::Invalid),
+            _ => Err(SafetyLevelParseError),
+        }
+    }
+}
+
+/// Error when parsing SafetyLevel from string.
+#[derive(Debug)]
+pub struct SafetyLevelParseError;
```
clabby, line 42: Can you derive `thiserror::Error` here, rather than manually implementing `Display` + `Error`? Also, this crate is `no_std` - should not be importing any `std` crates.

Should look like:

```rs
use thiserror::Error;

// ...

/// Error when parsing SafetyLevel from string.
#[derive(Error, Debug)]
#[error("Invalid SafetyLevel")]
pub struct SafetyLevelParseError;
```

clabby (changes requested):
Nice - couple changes req'd, but glad to add this.

.gitignore
```diff
@@ -15,3 +15,4 @@ data/
 
 # Exclude maili shadow crate Cargo.lock files
 **/**/**/maili/Cargo.lock
+.qodo
```
0xMushow, line 18: Remove this line too

0xMushow:
Also you can use `cargo +nightly fmt` to format it correctly 👍

0xMushow:
Whenever adding a new struct, you should also expose it to the `lib.rs` file within the same directory

It should look like this:

```rust
mod safety;
pub use safety::{
    SafetyLevel, SafetyLevelParseError,
};
```

crates/protocol/interop/src/safety.rs
```diff
@@ -19,6 +21,27 @@ pub enum SafetyLevel {
...
+            "invalid" => Ok(Self::Invalid),
+            _ => Err(SafetyLevelParseError),
+        }
+    }
+}
+
+/// Error when parsing SafetyLevel from string.
+#[derive(Error, Debug)]
+#[error("Invalid SafetyLevel")]
+pub struct SafetyLevelParseError;
```
mattsse, line 43: for better error feedback the error should include the unknown safety level.

for example if this is parser from incorrect user input, then it's nice to include the invalid input value

crates/protocol/interop/src/safety.rs
```diff
@@ -19,6 +21,27 @@ pub enum SafetyLevel {
...
 
+impl FromStr for SafetyLevel {
+    type Err = SafetyLevelParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "finalized" => Ok(Self::Finalized),
+            "safe" => Ok(Self::Safe),
+            "local_safe" | "localsafe" => Ok(Self::LocalSafe),
+            "cross_unsafe" | "crossunsafe" => Ok(Self::CrossUnsafe),
```
mattsse, line 32: don't think these underscore `_` values are accurate, since this uses kebab-case we should also support those or replace the `_` with `-`

mattsse (changes requested):
nits

gitToki:
I think this is good now 👍. Please let me know if anything is missing

crates/protocol/interop/src/safety.rs
```diff
@@ -41,7 +41,7 @@ impl FromStr for SafetyLevel {
 /// Error when parsing SafetyLevel from string.
 #[derive(Error, Debug)]
 #[error("Invalid SafetyLevel, error: {0}")]
-pub struct SafetyLevelParseError(pub String);
+pub struct SafetyLevelParseError(pub ToString);
```
mattsse: this can be just 
```suggestion
pub struct SafetyLevelParseError(pub String);
```

mattsse (approved):
lgtm

crates/protocol/interop/src/safety.rs
gitToki: Yes I made a mistake but the latest commit already implement this change

crates/protocol/interop/src/safety.rs
```diff
@@ -41,3 +65,23 @@ mod tests {
...
+    assert_eq!(SafetyLevel::from_str("invalid").unwrap(), SafetyLevel::Invalid);
+}
+
+#[test]
+fn test_safety_level_from_str_invalid() {
+    assert!(SafetyLevel::from_str("unknown").is_err());
+    assert!(SafetyLevel::from_str("123").is_err());
+    assert!(SafetyLevel::from_str("").is_err());
+    assert!(SafetyLevel::from_str("safe ").is_err());
+}
```
clabby, line 87: tiny nit, but please move this into `mod tests` and `cfg` gate the other two tests on `serde`.

clabby: After this, lgtm!

gitToki: It should be good now

crates/protocol/interop/src/safety.rs
```diff
@@ -19,6 +22,27 @@ pub enum SafetyLevel {
...
+    }
+}
+
+/// Error when parsing SafetyLevel from string.
+#[derive(Error, Debug)]
+#[error("Invalid SafetyLevel, error: {0}")]
+pub struct SafetyLevelParseError(pub String);
+
 #[cfg(test)]
 #[cfg(feature = "serde")]
```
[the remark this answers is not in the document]
clabby: ^ this is so the other tests don't need the `serde` feature to run, since the `FromStr` impl you've added doesn't require that feature to be enabled.

clabby (approved):
Thanks @gitToki - lgtm

gitToki:
Thanks to you for the review and for providing details!

crates/protocol/interop/src/safety.rs
```diff
@@ -19,12 +22,33 @@ pub enum SafetyLevel {
...
 
+impl FromStr for SafetyLevel {
+    type Err = SafetyLevelParseError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "finalized" => Ok(Self::Finalized),
+            "safe" => Ok(Self::Safe),
+            "local-safe" | "localsafe" => Ok(Self::LocalSafe),
+            "cross-unsafe" | "crossunsafe" => Ok(Self::CrossUnsafe),
```
SozinM, line 33: crossunsafe and localsafe are not valid safety level IMO
i think @emhane wanted to use serde_json::from_str in here, to keep things consistent

PR #1285, opened 2025-03-20
Refactor: Clap attribute macros from #[clap(...)] to #[arg(...)] and #[command(...)] in v4.x

### LINKED ISSUE #1284: feat: migrate #[clap(...)] to #[arg(...)] / #[command(...)] in clap 4.x
In **clap** version 4.x and above, the use of `#[clap(...)]` is no longer valid.  
Many instances of `#[clap(...)]` are still present in the project.  
These should be replaced with `#[arg(...)]` or `#[command(...)]`.

I ran the following command to check for deprecated clap attributes: 

 `cargo check --features clap/deprecated ` 

The following warnings were generated:

Image

Himess:
This PR updates deprecated #[clap(...)] attributes to their modern equivalents in clap 4.x.
The current codebase still uses outdated syntax that has been deprecated since version 4.0.
By making this update, we ensure compatibility with future versions and maintain code quality.

I ran cargo check --features clap/deprecated after making the changes, and the previous clap deprecation warnings are no longer present.

image

Fixes #1284

### DIFF
--- a/bin/host/src/bin/host.rs
+++ b/bin/host/src/bin/host.rs
@@ -27,7 +27,7 @@ pub struct HostCli {
     #[arg(long, short, action = ArgAction::Count)]
     pub v: u8,
     /// Host mode
-    #[arg(subcommand)]
+    #[command(subcommand)]
     pub mode: HostMode,
 }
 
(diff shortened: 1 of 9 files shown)


mattsse (approved):
ty

refcell (approved):

PR #1287, opened 2025-03-20
fix(isthmus): Upgrade transactions + system calls

clabby:
## Overview

Fixes the upgrade transactions and system calls for Isthmus.


crates/proof/executor/src/syscalls/tx_env.rs
```diff
@@ -41,7 +41,7 @@ pub(crate) fn fill_tx_env_for_contract_call(
         blob_hashes: Vec::new(),
         max_fee_per_blob_gas: None,
         optimism: OptimismFields {
-            source_hash: None,
+            source_hash: Some(B256::ZERO),
```
clabby, line 44: Syscalls are free; The operator fee is always charged to non-deposits, and can have a non-zero constant floor.

crates/protocol/protocol/src/fee.rs
```diff
@@ -77,7 +77,7 @@ pub fn calculate_tx_l1_cost_bedrock_empty_scalars(
     base_fee: U256,
     l1_fee_scalar: U256,
 ) -> U256 {
-    if input.is_empty() || input.first() == Some(&0x7F) {
+    if input.is_empty() || input.first() == Some(&0x7E) {
```
mattsse, line 80: this exists as a constant here

https://github.com/alloy-rs/op-alloy/blob/88c6c2efc063a7baf14b5a84829157dd9a5d1d75/crates/consensus/src/transaction/tx_type.rs#L10-L10

useful to avoid these things :)

refcell (approved):

crates/proof/executor/src/syscalls/tx_env.rs
refcell: We really should put this in a comment for posterity but can be a follow-up pr

refcell:
Thanks for fixing these 🙏

PR #1292, opened 2025-03-24
chore: Remove Magic 0x7E Deposit Identifier Bytes

refcell:
### Description

Removes magic deposit tx identifier bytes (`0x7E`) in favor of the strongly typed `op_alloy_consensus::OpTxType::Deposit` variant.


clabby (approved):
bless

emhane (approved):

PR #1314, opened 2025-03-27
feat(node): RPC CLI Args

### LINKED ISSUE #1312: feat(node/cli): RPC CLI Args
### Description

Add rpc flags to the `kona-node` as a flattened set of args similar to how the [`p2p` args are flattened](https://github.com/op-rs/kona/blob/main/bin/node/src/commands/node.rs#L55-L57) in the `kona-node` `node` subcommand.

refcell:
### Description

RPC CLI arguments.

Closes #1312.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3747,7 +3747,7 @@ dependencies = [
  "revm",
  "serde",
  "serde_json",
- "spin",
+ "spin 0.10.0",
  "thiserror 2.0.12",
  "tracing",
 ]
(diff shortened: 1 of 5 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1314 👈
* `main`

refcell:
> LGTM
> 
> For the `long` variant of the args - do we want to standardize on the Go-style `--namespace.option` style? Not against it, but there's a bit of disparity in some of the base CLI args where they use `--non-namespaced-flags`.

Let's standardize on the go style so it can be as compatible with the `op-node` from a UX perspective as possible. Having a sort-of minimal diff (where perf doesn't matter) will make it much nicer experience for operators.

PR #1318, opened 2025-03-27
feat(node): Rpc Actor

### LINKED ISSUE #1317: feat(node/service): RpcActor
### Description

This ticket is to setup an `RpcActor` in the `kona-node-service` that contains a [`ServerHandle`](https://docs.rs/jsonrpsee-server/0.24.9/jsonrpsee_server/struct.ServerHandle.html) and a `CancellationToken`.

The actor should simply call cancel on the `CancellationToken` if the [`ServerHandle`](https://docs.rs/jsonrpsee-server/0.24.9/jsonrpsee_server/struct.ServerHandle.html) stops and call `stop()` on the [`ServerHandle`](https://docs.rs/jsonrpsee-server/0.24.9/jsonrpsee_server/struct.ServerHandle.html) if the `CancellationToken` is cancelled. Handling these futures should be done through the [`tokio::select!`](https://docs.rs/tokio/latest/tokio/macro.select.html) macro.

A future ticket will handle wiring up this `RpcActor` into the `ValidatorNodeService` trait in `kona-node-service`.

refcell:
### Description

Implements an actor in the `kona-node-service` that handles the RPC server.

Closes #1317

### DIFF
--- a/crates/node/service/src/actors/rpc.rs
+++ b/crates/node/service/src/actors/rpc.rs
@@ -25,12 +25,6 @@ pub struct RpcActor {
     cancellation: CancellationToken,
 }
 
-// impl std::fmt::Debug for RpcActor {
-//     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-//         write!(f, "RPCServerActor")
-//     }
-// }
-
 impl RpcActor {
     /// Constructs a new [`RpcActor`] given the [`ServerHandle`] and [`CancellationToken`].
     pub const fn new(server: ServerHandle, cancellation: CancellationToken) -> Self {
--- a/crates/node/service/src/actors/rpc.rs
+++ b/crates/node/service/src/actors/rpc.rs
@@ -1,0 +1,65 @@
+//! RPC Server Actor
+
+use crate::NodeActor;
+use async_trait::async_trait;
+use jsonrpsee::server::ServerHandle;
+use tokio_util::sync::CancellationToken;
+
+/// An error returned by the [`RpcActor`].
+#[derive(Debug, thiserror::Error)]
+pub enum RpcActorError {
+    /// The [`RpcActor`]'s RPC server stopped unexpectedly.
+    #[error("RPC server stopped unexpectedly")]
+    ServerStopped,
+    /// Failed to stop the RPC server.
+    #[error("Failed to stop the RPC server")]
+    StopFailed,
+}
+
+/// An actor that handles the RPC server for the rollup node.
+#[derive(Debug)]
+pub struct RpcActor {
+    /// The handle to the RPC server.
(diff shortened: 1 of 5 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1318 👈
* `main`

crates/node/service/src/actors/rpc.rs
```diff
@@ -0,0 +1,65 @@
...
+impl NodeActor for RpcActor {
+    type InboundEvent = ();
+    type Error = RpcActorError;
+
+    async fn start(mut self) -> Result<(), Self::Error> {
+        tokio::select! {
+            _ = self.server.clone().stopped() => {
+                // The server has stopped, so we should stop as well.
+                self.cancellation.cancel();
+                return Err(RpcActorError::ServerStopped);
```
clabby, line 51: Returning an error signals to the orchestration code to cancel all other tasks. Not a problem to be explicit, but the cancellation call isn't required.

crates/node/service/src/actors/rpc.rs
```diff
@@ -0,0 +1,65 @@
...
+    /// The cancellation token, shared between all tasks.
+    cancellation: CancellationToken,
+}
+
+// impl std::fmt::Debug for RpcActor {
+//     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+//         write!(f, "RPCServerActor")
+//     }
+// }
+
```
clabby, line 33: If the derive compiles, can delete this
```suggestion
```

clabby (approved):
Neat 🗡️

crates/node/service/src/actors/rpc.rs
refcell: Oops haha

crates/node/service/src/actors/rpc.rs
refcell: What should the policy here be you think? Restart the server?

clabby: For now it's fine to just halt the whole process if the server fails like you're doing now imo. Let's circle back once the node is running - I'd wager that if the server crashes, something has gone really wrong.

clabby: But if we do see it intermittently stopping the node, totally - should just eagerly attempt to restart it such that this actor can never fail, only be cancelled.

refcell: Ok will just add a small ticket to track for now.

I think it may make sense for periphery services like this to have some sort of restart policy by default and have a "fail fast" mode via toggle.

refcell: https://github.com/op-rs/kona/issues/1320

PR #1322, opened 2025-03-27
fix(node/service): Unsafe Block Signer

refcell:
### Description

Fixes the rollup node to use the correct unsafe block signer.

### DIFF
--- a/bin/node/src/flags/globals.rs
+++ b/bin/node/src/flags/globals.rs
@@ -1,6 +1,8 @@
 //! Global arguments for the CLI.
 
+use alloy_primitives::Address;
 use clap::{ArgAction, Parser};
+use kona_registry::ROLLUP_CONFIGS;
 
 /// Global arguments for the CLI.
 #[derive(Parser, Clone, Debug)]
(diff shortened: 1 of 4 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1322 👈
* `main`

bin/node/src/flags/globals.rs
```diff
@@ -20,3 +22,18 @@ pub struct GlobalArgs {
...
+    pub fn signer(&self) -> anyhow::Result<Address> {
+        let id = self.l2_chain_id;
+        ROLLUP_CONFIGS
+            .get(&id)
+            .ok_or(anyhow::anyhow!("No rollup config found for chain ID: {id}"))?
+            .genesis
+            .system_config
+            .as_ref()
+            .map(|c| c.batcher_address)
+            .ok_or(anyhow::anyhow!("No system config found for chain ID: {id}"))
```
clabby, line 37: Can we do this? The unsafe block signer can be changed in the `SystemConfig`, so it's not guaranteed to be the same as the one in the genesis state of the `SystemConfig`. Base for example changed theirs recently, and that bit us over in kona-proof.

bin/node/src/flags/globals.rs
```diff
@@ -20,3 +22,18 @@ pub struct GlobalArgs {
     )]
     pub metrics_port: u16,
 }
+
+impl GlobalArgs {
+    /// Returns the signer [`Address`] from the rollup config for the given l2 chain id.
+    pub fn signer(&self) -> anyhow::Result<Address> {
```
clabby, line 28: Let's rename this to `genesis_signer` so that it's a bit more clear.

clabby (approved):

bin/node/src/flags/globals.rs
refcell: #1326

PR #1323, opened 2025-03-27
feat(node): Network Config

refcell:
### Description

Pulls sparse networking parameters in the `RollupNode` into an encapsulated `NetConfig` similar to the `RpcConfig`.


refcell:
## 📚 $\text{Stack Overview}$

* #1323 👈
* #1322
* `main`

clabby (approved):

crates/node/service/src/service/standard/mod.rs
```diff
@@ -101,27 +93,17 @@ impl ValidatorNodeService for RollupNode {
...
-            RollupNodeError::Network(NetworkBuilderError::DiscoveryDriverBuilder(
-                Discv5BuilderError::ListenConfigNotSet,
-            ))
-        })?;
-        let keypair = self.keypair.clone().unwrap_or_else(Keypair::generate_secp256k1);
-
+        if self.net_config.is_none() {
+            warn!(
+                target: "rollup_node",
+                "No network configuration provided, skipping network initialization"
```
clabby, line 99: I think we should call this p2p. Network configuration is a bit opaque, means quite a few things. (Questions a user might ask themselves: Is network configuration for L2 network config? L1 network config? What network are we talking about?)

Issue #1127, opened 2025-02-26
feat(node/rpc): P2P Namespace Support


refcell:
### Description

Add support for the `p2p` namespace following the [`op-node` implementation](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/node/server.go#L56-L63).

PR #1327, opened 2025-03-27
feat(node/p2p): P2P RPC Server

refcell:
### Description

P2P RPC Server. Notice it only implements a single method for now.

Makes progress on #1127.


refcell:
## 📚 $\text{Stack Overview}$

* #1327 👈
* `main`

crates/node/p2p/src/rpc/server.rs
```diff
@@ -0,0 +1,123 @@
...
+impl OpP2PApiServer for NetworkRpc {
+    async fn opp2p_self(&self) -> RpcResult<PeerInfo> {
+        tracing::trace!("opp2p_self");
+        let (tx, rx) = tokio::sync::oneshot::channel();
+        self.sender
+            .send(NetRpcRequest::PeerInfo(tx))
+            .await
+            .map_err(|_| ErrorObject::from(ErrorCode::InternalError))?;
+
+        rx.await.map_err(|_| ErrorObject::from(ErrorCode::InternalError))
```
clabby, line 46: Good comms pattern.

Not necessary for this PR, but we're going to have _a lot_ of this. Maybe it's worth making a `Request` type that encapsulates this handshake into an easy-to-use abstraction.

clabby (approved):
Good start

crates/node/p2p/src/rpc/server.rs
refcell: Good callout, will do

PR #1330, opened 2025-03-28
chore(node/p2p): Network RPC Request Handling

refcell:
### Description

Handles rpc requests sent to the network stack.

### DIFF
--- a/crates/node/p2p/src/net/driver.rs
+++ b/crates/node/p2p/src/net/driver.rs
@@ -59,6 +59,17 @@ impl Network {
         tokio::spawn(async move {
             loop {
                 select! {
+                    event = self.gossip.select_next_some() => {
+                        trace!(target: "p2p::driver", "Received event: {:?}", event);
+                        self.gossip.handle_event(event);
+                    },
+                    enr = handler.enr_receiver.recv() => {
+                        let Some(ref enr) = enr else {
+                            trace!(target: "p2p::driver", "Receiver `None` peer enr");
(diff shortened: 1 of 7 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1330 👈
* #1327
* `main`

crates/node/p2p/src/net/driver.rs
```diff
@@ -47,12 +52,47 @@ impl Network {
...
     /// and continually listens for new peers and messages to handle
     pub fn start(mut self) -> Result<(), TransportError<std::io::Error>> {
+        let mut rpc = self.rpc.unwrap_or_else(|| tokio::sync::mpsc::channel(1).1);
         let mut handler = self.discovery.start();
         self.gossip.listen()?;
         let mut interval = tokio::time::interval(tokio::time::Duration::from_secs(5));
         tokio::spawn(async move {
             loop {
                 select! {
+                    req = rpc.recv() => {
```
clabby, line 62: Small nit: We should probably `bias` this `select!`.

By default, this event loop will poll all of the futures in pseudo-random order, but there's a hierarchy here in terms of which futures we want to poll first (i.e., we never want to block unsafe block gossip because of RPC requests.)

not blocking on this PR, this is a problem we'll need to sort out everywhere.

```rs
select! {
    biased;

    // ... futures in order of poll priority ...
}
```

clabby (approved):
very nice paul allen

PR #1334, opened 2025-03-28
chore(binaries): Argument Cleanup

refcell:
### Description

Argument cleanup that follows #1323.


refcell:
Replaced by https://github.com/op-rs/kona/pull/1338

PR #1337, opened 2025-03-28
fix(node/p2p): Forward ENRs to the Swarm

refcell:
### Description

Previously, we only sent ENRs to the gossip service _iff_ they were discovered via a random search.

But we already have a bootstore and bootnodes that we can try to peer with... DUHHHH

So now the discovery service will just forward all ENRs along from the bootstore + bootnodes to the gossip service on startup.
(errors are handled gracefully but we may have to make the channel size more configurable since the bootstore could get  h u g e)


refcell:
## 📚 $\text{Stack Overview}$

* #1337 👈
* `main`

PR #1339, opened 2025-03-29
fix(bin): Correct Unsafe Block Signer

refcell:
### Description

Use the correct unsafe block signer.


refcell:
## 📚 $\text{Stack Overview}$

* #1339 👈
* `main`

PR #1348, opened 2025-03-31
chore(node/p2p): Log Cleanup

refcell:
### Description

Small log cleanup.

### DIFF
--- a/crates/node/p2p/src/gossip/driver.rs
+++ b/crates/node/p2p/src/gossip/driver.rs
@@ -92,11 +92,11 @@ impl GossipDriver {
     pub fn dial(&mut self, enr: Enr) {
         let key = OpStackEnr::OP_CL_KEY.as_bytes();
         if enr.get_raw_rlp(key).is_none() {
-            debug!(target: "p2p::gossip::driver", "Ignoring peer without opstack CL key...");
+            debug!("Ignoring peer without opstack CL key...");
             return;
         }
         let Some(multiaddr) = enr_to_multiaddr(&enr) else {
-            debug!(target: "p2p::gossip::driver", "Failed to extract tcp socket from enr: {:?}", enr);
+            debug!("Failed to extract tcp socket from enr: {:?}", enr);
             return;
         };
         self.dial_multiaddr(multiaddr);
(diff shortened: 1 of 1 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1348 👈
* `main`

crates/node/p2p/src/gossip/driver.rs
```diff
@@ -105,17 +105,17 @@ impl GossipDriver {
     /// Dials the given [`Multiaddr`].
     pub fn dial_multiaddr(&mut self, addr: Multiaddr) {
         if self.has_dialed(&addr) {
-            debug!(target: "p2p::gossip::driver", "Already connected to peer: {:?}", addr.clone());
+            trace!("Already connected to peer: {:?}", addr.clone());
```
emhane, line 108: why remove the target filter? 
```suggestion
            trace!(peer=%addr
                "Already connected to peer"
            );
```

refcell: It's already registered and printed by module path

crates/node/p2p/src/gossip/driver.rs
```diff
@@ -128,34 +130,40 @@ impl GossipDriver {
...
-                    debug!(target: "p2p::gossip::driver", "Reporting message validation result: {:?}", status);
                     _ = self
                         .swarm
                         .behaviour_mut()
                         .gossipsub
                         .report_message_validation_result(&id, &src, status);
                 }
             }
+            libp2p::gossipsub::Event::Subscribed { peer_id, topic } => {
+                debug!(target: "p2p::gossip::driver", "Peer: {:?} subscribed to topic: {:?}", peer_id, topic);
```
clabby, line 144: Bit of a mixture of logs with manual `target` settings, prefer to go one way or the other. Whole codebase has favored manually setting the `target` to date, but I don't really mind going either direction for net-new components like this so long as it's consistent.

refcell: Right - i'm trying to move away from this because it's already handled by tracing and most of the time we're just making the code less legible. I also think it's fine to have a mix of both in the case where we do want to specify the target manually.

clabby: That's fine. I'm not too attached, definitely don't need to go either way in this PR.

Eventually I'd like to follow up on a global prefix, which probably means removing the `target` filters across the codebase. Makes using the `EnvFilter` a _lot_ easier.

clabby (approved):
Log cleanup lgtm, approving to unblock - can merge after CI is fixed.

PR #1349, opened 2025-03-31
chore(node/p2p): Cleanup Network Driver

refcell:
### Description

Cleans up the network driver to remove the logs on an interval.

Allows the downstream user to request stats through the RPC channel instead of always logging on an interval, hurting perf.

Since this is an exposed mpsc, we should really move away from calling it an RPC channel since it's open to any external request and is used as such.


refcell:
## 📚 $\text{Stack Overview}$

* #1349 👈
* `main`

clabby (approved):

PR #1353, opened 2025-03-31
fix(node/p2p): OP Stack ENRs

refcell:
### Overview

Fixes `kona-p2p` to bootstrap the discv5 table with ENRs that contain the OP Stack CL ENR key.

### Description

Previously, we were populating the table with any enr found and recorded in the bootstore.
Quickly, as the bootstore filled up, we hit a limit on the number of ENRs that could be added to the discv5 table.
We essentially had no priority in which ENRs to place in the discovery table.

This PR filters out ENRs in the bootstore that do not contain op stack keys so the discv5 table is only initially populated with:
- Bootnodes, hardcoded into `kona-p2p`
- ENRs in the bootstore that have the op stack CL key

Bootstrapping now effectively ignores ENRs in the bootstore that do not have the op stack CL key.

### DIFF
--- a/crates/node/p2p/src/discv5/driver.rs
+++ b/crates/node/p2p/src/discv5/driver.rs
@@ -93,9 +93,6 @@ impl Discv5Driver {
     async fn bootstrap(&mut self) {
         let nodes = BootNodes::from_chain_id(self.chain_id);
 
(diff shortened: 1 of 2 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1353 👈
* `main`

clabby (approved):

PR #1364, opened 2025-04-02
chore: Bump op-alloy Patch

refcell:
### Description

Bumps `op-alloy` to `0.11.4`, a backport of changes made after `0.12.0`.
This release contains the payload hash fix and ssz encoding of the `OpNetworkPayloadEnvelope` type.
With these changes, kona's P2P stack successfully decodes unsafe blocks over p2p gossip
and is in a spot to support gossiping out unsafe blocks using the ssz encoding.

### DIFF
--- a/crates/node/p2p/src/gossip/handler.rs
+++ b/crates/node/p2p/src/gossip/handler.rs
@@ -44,22 +44,17 @@ pub struct BlockHandler {
 }
 
 impl Handler for BlockHandler {
-    /// Checks validity of a block received via p2p gossip, and sends to the block update channel if
-    /// valid.
+    /// Checks validity of a [`OpNetworkPayloadEnvelope`] received over P2P gossip.
+    /// If valid, sends the [`OpNetworkPayloadEnvelope`] to the block update channel.
     fn handle(&self, msg: Message) -> MessageAcceptance {
         let decoded = if msg.topic == self.blocks_v1_topic.hash() {
-            info!("received v1 block");
             OpNetworkPayloadEnvelope::decode_v1(&msg.data)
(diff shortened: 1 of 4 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1364 👈
* `main`

crates/node/p2p/src/gossip/handler.rs
```diff
@@ -68,10 +63,11 @@ impl Handler for BlockHandler {
         match decoded {
             Ok(envelope) => {
                 if self.block_valid(&envelope) {
-                    _ = self.block_sender.send(envelope);
```
refcell, line 71: Big rookie mistake - we ignored the future here instead of attempting to send. Unfortunately this can fail, and unsafe blocks could be ignored.

Maybe what we want here is to hold the blocks in a cache instead and allow the gossip driver to pull any unsafe blocks periodically.

Alternatively, this handler is invoked directly from the gossip driver. Instead of having the block handler send the `OpNetworkPayloadEnvelope` through a channel, we can hoist this logic up. The block handler can actually just return the `OpNetworkPayloadEnvelope` itself, which could be bubbled up all the way to the network driver. Since the network driver runs the blocking loop, it can continuously attempt to send any unsafe blocks in its cache through the channel.

clabby: Yeah throwing them into a queue makes sense for sure. We could have it push/pull, but ensuring that the unsafe payloads stick around until we can guarantee the engine actor has received them sounds ideal. The implementation detail I'm not super opinionated on - good callout 💯

refcell: https://github.com/op-rs/kona/issues/1365

PR #1366, opened 2025-04-02
fix(node/p2p): Unsafe Payload Sending

### LINKED ISSUE #1365: fix(node/p2p): Bubble Up Unsafe Payload
### Description

Currently, kona's `BlockHandler` handles both unsafe block validation _and_ [sending the `OpNetworkPayloadEnvelope` through a channel](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/gossip/handler.rs#L71) to whoever is receiving the unsafe block. The linked code actually fails to even send the payload through the channel since the future is dropped immediately. This is fixed in #1364. Still, the issue remains to decouple payload forwarding and block validation.

Since sending messages over a channel is a fallible operation, it makes sense to split this logic out of the infallible block validation method. The `BlockHandler` should be updated to return the payload with the `MessageAcceptance` type. It can be returned as `(MessageAcceptance, Option<OpNetworkPayloadEnvelope>)`. Alternatively, a new type can be introduced to contain both `MessageAcceptance` and the `Option<OpNetworkPayloadEnvelope>`, allowing for convenience methods like `is_valid()` and such.

Since the block handler now no longer uses the unsafe block `Sender`, it can be removed and handled in the `Network` driver loop at a higher level. The gossip driver should also bubble up the unsafe block [here](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/gossip/driver.rs#L133) since the `Network` driver calls `handle_event` directly on the gossip driver.

refcell:
### Description

Fixes unsafe payload sending in kona's P2P stack by using a queue at the top-level.

This splits the fallible block sending logic out of the `BlockHandler`, letting the block handling solely perform validation.

Closes #1365.

### DIFF
--- a/bin/node/src/commands/net.rs
+++ b/bin/node/src/commands/net.rs
@@ -46,8 +46,7 @@ impl NetCommand {
             .with_chain_id(args.l2_chain_id)
             .with_rpc_receiver(rx)
             .build()?;
-        let mut recv =
-            network.take_unsafe_block_recv().ok_or(anyhow::anyhow!("No unsafe block receiver"))?;
+        let mut recv = network.unsafe_block_recv();
         network.start()?;
         info!("Network started, receiving blocks.");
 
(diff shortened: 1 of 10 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1366 👈
* `main`

clabby (approved):

PR #1367, opened 2025-04-03
feat(node/p2p): Redial Peers

refcell:
### Description

A more aggressive policy, but the peers keep disconnecting when running kona's p2p stack.

The logs on my machine read that the address is in use (for the selected tcp address?)

I'm thinking redialing might allow for a new tcp addr to be used, but i'm not sure if kona p2p just doesn't have enough listeners or what's going on.

Regardless, this PR adds a peerstore to the gossip driver that tracks the mapping from peer id to multiaddress.


refcell:
## 📚 $\text{Stack Overview}$

* #1367 👈
* `main`

crates/node/p2p/src/gossip/driver.rs
```diff
@@ -62,7 +71,7 @@ impl GossipDriver {
 
     /// Returns if the swarm has already successfully dialed the given [`Multiaddr`].
     pub fn has_dialed(&mut self, addr: &Multiaddr) -> bool {
-        self.dialed_peers.contains(addr)
+        self.dialed_peers.get(addr).cloned().unwrap_or(false)
```
clabby, line 74: This doesn't make much sense at a glance - why is this better than `.contains` when `.get` doesn't take ownership?

crates/node/p2p/src/gossip/config.rs
```diff
@@ -60,6 +62,38 @@ lazy_static! {
...
+/// Returns the default score thresholds for the gossipsub protocol `Behaviour`.
+pub fn default_score_thresholds() -> PeerScoreThresholds {
+    PeerScoreThresholds {
+        gossip_threshold: -10.0,
+        publish_threshold: -50.0,
+        graylist_threshold: -80.0,
+        accept_px_threshold: 10.0,
+        opportunistic_graft_threshold: 20.0,
+    }
+}
```
clabby, line 95: (non-review comment)

Any way I can get more familiar with this tuning?

clabby (approved):
Redial policy + p2p CLI changes look good.

I don't understand the peer scoring configuration well enough to give a good review there, but fine to get this in to keep iterating.

crates/node/p2p/src/gossip/config.rs
refcell: These are actually mostly default, but I was just checking if they were not set for some reason. Will remove :)

crates/node/p2p/src/gossip/driver.rs
refcell: Because it's a hashmap and we want to get the value not if the key is contained.

But you call out a good point here that the `.cloned()` can be fixed to be `.copied()`.

PR #1369, opened 2025-04-04
chore(node/p2p): Small Log Cleanup

refcell:
### Description

Small log cleanup.

### DIFF
--- a/crates/node/p2p/src/net/driver.rs
+++ b/crates/node/p2p/src/net/driver.rs
@@ -78,21 +78,21 @@ impl Network {
                 }
                 select! {
                     event = self.gossip.select_next_some() => {
-                        trace!(target: "p2p::driver", "Received event: {:?}", event);
+                        trace!("Received event: {:?}", event);
                         if let Some(payload) = self.gossip.handle_event(event) {
                             unsafe_blocks.push_back(payload);
                         }
                     },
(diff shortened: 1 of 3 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1369 👈
* `main`

crates/node/p2p/src/net/driver.rs
```diff
@@ -78,21 +78,21 @@ impl Network {
                 }
                 select! {
                     event = self.gossip.select_next_some() => {
-                        trace!(target: "p2p::driver", "Received event: {:?}", event);
+                        trace!("Received event: {:?}", event);
```
emhane, line 81:
```suggestion
                        trace!(target: "p2p::driver",
                            %event,
                            "Received event"
                        );
```
I would do it like this. i would highly recommend to keep the targets, that way you can zoom in on certain components when debugging, e.g. `RUST_LOG=p2p::driver,p2p::discv5=trace`

refcell: You can still do this natively using the module as a target
Screenshot 2025-04-07 at 8 07 57 PM

clabby (approved):
LGTM

re: `target` env filter, fine for now - let's please standardize on using native module derived targets or manual targets at some point.

PR #1384, opened 2025-04-08
fix(bin/host): Fix typo

palango:
Fix a small typo.


clabby (commented):
Thanks - this error doesn't even look to be used, though. Could you delete it instead?

clabby (approved):

PR #1385, opened 2025-04-08
chore(ci): Bump monorepo commit

clabby:
## Overview

Bumps the monorepo commit to use the latest version of action tests.

Also adjusts the semantics of the pectra blob schedule fork activation. Previously, we treated it as a fork with hierarchy. The go implementation has just implemented it as an independent feature toggle. This PR corrects this to match the behavior.

ref: https://github.com/ethereum-optimism/optimism/blob/f76ef1f2a010a1860423f703eb776d5ae9506049/op-node/rollup/derive/l1_block_info.go#L419-L431


refcell (approved):
Nice thanks for updating

PR #1386, opened 2025-04-08
feat(node/service): Unsafe Block Signer Updates

### LINKED ISSUE #1326: feat(node/net): Update Unsafe Block Signer
### Description

https://github.com/op-rs/kona/pull/1322 initializes the `RollupNode` - specifically the `Network` - with the unsafe block signer from the rollup config. That is, the batcher address from the chain's rollup config grabbed from the `kona-registry`.

The issue is the unsafe block signer can change when the `SystemConfig` is updated. 

This ticket is to send a message from the l1 chain watcher actor to the `Network` actor if the `SystemConfig` unsafe block signer is updated.

refcell:
### Description

Properly wires up unsafe block signer updates using the `L1WatcherRpc` and the `SystemConfigUpdate` log type.

Closes #1326.

### DIFF
--- a/crates/node/service/src/actors/l1_watcher_rpc.rs
+++ b/crates/node/service/src/actors/l1_watcher_rpc.rs
@@ -1,11 +1,14 @@
 //! [NodeActor] implementation for an L1 chain watcher that checks for L1 head updates over RPC.
 
 use crate::NodeActor;
-use alloy_primitives::B256;
+use alloy_primitives::{Address, B256};
 use alloy_provider::{Provider, RootProvider};
+use alloy_rpc_types_eth::Log;
 use async_trait::async_trait;
 use futures::StreamExt;
+use kona_genesis::{RollupConfig, SystemConfigLog, SystemConfigUpdate, UnsafeBlockSignerUpdate};
 use kona_protocol::BlockInfo;
+use std::sync::Arc;
 use thiserror::Error;
 use tokio::{
     select,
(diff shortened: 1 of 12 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1386 👈
* `main`

crates/node/service/src/actors/l1_watcher_rpc.rs
```diff
@@ -94,7 +120,30 @@ impl NodeActor for L1WatcherRpc {
...
+                                );
+                                if let Err(e) = self.block_signer_sender.send(unsafe_block_signer) {
+                                    error!(
+                                        target: "l1_watcher",
+                                        "Error sending unsafe block signer update: {e}"
+                                    );
+                                }
+                            }
+                        }
+
```
clabby, line 146:
```suggestion
```

clabby (approved):
Looks fine for now. I think we can de-duplicate this processing by lifting the signal of an unsafe block signer update out of the derivation pipeline, which should already process this event.

This is definitely easier, though. Not a problem to start off with this.

crates/node/service/src/actors/l1_watcher_rpc.rs
refcell: The event might be registered by the derivation pipeline, but the log update was never actually parsed from the `Log` instance.

Issue #1110, opened 2025-02-25
feat(pilot): Engine Startup Signal


refcell:
### Description

- Implement the [`op_alloy_provider::ext::engine::OpEngineApi`](https://docs.rs/op-alloy-provider/latest/op_alloy_provider/ext/engine/trait.OpEngineApi.html) trait.
- Instantiate the provider in the node binary.
- On startup, query the [forkchoice update](https://docs.rs/op-alloy-provider/latest/op_alloy_provider/ext/engine/trait.OpEngineApi.html#tymethod.fork_choice_updated_v3) with empty payload attributes to get the status of the node.
  - See: https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/driver/state.go#L437

PR #1390, opened 2025-04-08
feat(node/service): Wire in the Engine Actor

refcell:
### Description

Wires in the `EngineActor`. Not refactored nicely yet. Avoiding premature abstractions for something that works.

Makes progress on #1110.

PR #1383, opened 2025-04-08
feat(node/service): Wire in Engine Arguments

refcell:
### Description

Wires in the engine arguments.

PR #1391, opened 2025-04-08
feat(node/service): Engine Consolidation Task

refcell:
### Description

Sets up the consolidation task for the `EngineActor`.

Progress on #1110.

### DIFF
--- a/crates/node/engine/src/attributes.rs
+++ b/crates/node/engine/src/attributes.rs
@@ -1,0 +1,324 @@
+//! Contains a utility method to check if attributes match a block.
+
+use alloy_primitives::{Address, B256};
+use alloy_rpc_types_eth::{Block, Withdrawals};
+use kona_genesis::RollupConfig;
+use kona_rpc::OpAttributesWithParent;
+use op_alloy_rpc_types::Transaction;
+
+/// Represents whether the attributes match the block or not.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum AttributesMatch {
+    /// The attributes match the block.
+    Match,
+    /// The attributes do not match the block.
+    Mismatch(AttributesMismatch),
+}
+
+impl AttributesMatch {
+    /// Returns true if the attributes match the block.
+    pub const fn is_match(&self) -> bool {
+        matches!(self, Self::Match)
+    }
+
+    /// Returns true if the attributes do not match the block.
+    pub const fn is_mismatch(&self) -> bool {
+        matches!(self, Self::Mismatch(_))
+    }
+
+    /// Checks that withdrawals for a block and attributes match.
+    pub fn check_withdrawals(
+        config: &RollupConfig,
+        attributes: &OpAttributesWithParent,
+        block: &Block<Transaction>,
+    ) -> Self {
+        let attr_withdrawals = attributes.attributes.payload_attributes.withdrawals.as_ref();
+        let attr_withdrawals = attr_withdrawals.map(|w| Withdrawals::new(w.to_vec()));
+        let block_withdrawals = block.withdrawals.as_ref();
+
+        if config.is_canyon_active(block.header.timestamp) {
+            // In canyon, the withdrawals list should be some and empty
+            if attr_withdrawals.is_none_or(|w| !w.is_empty()) {
+                return Self::Mismatch(AttributesMismatch::CanyonWithdrawalsNotEmpty);
+            }
+            if block_withdrawals.is_none_or(|w| !w.is_empty()) {
+                return Self::Mismatch(AttributesMismatch::CanyonWithdrawalsNotEmpty);
+            }
+            if !config.is_isthmus_active(block.header.timestamp) {
+                // In canyon, the withdrawals root should be set to the empty value
+                let empty_hash = alloy_consensus::EMPTY_ROOT_HASH;
+                if block.header.inner.withdrawals_root != Some(empty_hash) {
+                    return Self::Mismatch(AttributesMismatch::CanyonNotEmptyHash);
+                }
+            }
+        } else {
+            // In bedrock, the withdrawals list should be None
+            if attr_withdrawals.is_some() {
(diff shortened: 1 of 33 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1391 👈
* #1390
* #1387
* #1383
* `main`

crates/node/service/src/service/validator.rs
```diff
@@ -111,9 +111,17 @@ pub trait ValidatorNodeService {
         let derivation = Some(derivation);
 
         let launcher = self.engine();
+        let client = launcher.client();
         let sync = launcher.sync.clone();
         let engine = launcher.launch().await?;
-        let engine = EngineActor::new(sync, engine, cancellation.clone());
+        let engine = EngineActor::new(
+            std::sync::Arc::new(self.config().clone()),
```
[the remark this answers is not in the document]
clabby: all good.

crates/node/service/src/actors/engine.rs
```diff
@@ -104,6 +119,20 @@ impl NodeActor for EngineActor {
...
+                        debug!(target: "engine", "Received `None` attributes from receiver");
+                        continue;
+                    };
+                    let task = BuildTask::new(
+                        Arc::clone(&self.client),
+                        Arc::clone(&self.config),
+                        attributes,
+                        true,
+                    );
+                    let task = EngineTask::BuildBlock(task);
```
clabby, line 133: We'll need to add another check here that does the L1 consolidation. We only want to build a block if `safe` + `unsafe` are equal.

If you'd like to get this in and iterate tho all for it.

refcell: Yea the whole point of this stack is just to get the basic channels set up between actors. Then the correct rules can be applied.

refcell:
Not ready to merge. Working on consolidation.

crates/node/engine/src/attributes.rs
```diff
@@ -0,0 +1,324 @@
...
+        } else {
+            // In bedrock, the withdrawals list should be None
+            if attr_withdrawals.is_some() {
+                return Self::Mismatch(AttributesMismatch::BedrockWithdrawals);
+            }
+        }
+
+        if config.is_isthmus_active(block.header.timestamp) {
+            // In isthmus, the withdrawals root must be set
+            if block.header.inner.withdrawals_root.is_none() {
```
clabby, line 63: I think this is fine; Properly checking the WD root against the state root would take a while and this is a particularly hot path.. Does the `op-node` also not actually check this?

crates/node/engine/src/task_queue/core.rs
```diff
@@ -31,10 +34,40 @@ impl Engine {
...
+    ) -> Result<(), ConsolidationError> {
+        if self.needs_consolidation() {
+            debug!(target: "engine", "Performing consolidation");
+            // TODO: consolidate
+            // see: https://specs.optimism.io/protocol/derivation.html#l1-consolidation-payload-attributes-matching
+        } else {
+            debug!(target: "engine", "Skipping consolidation. Safe head [{}] == Unsafe Head [{}]", self.state.safe_head().block_info.number, self.state.unsafe_head().block_info.number);
+        }
+        Ok(())
+    }
```
refcell, line 69: Oops forgot to delete this.

crates/node/engine/src/attributes.rs
```diff
@@ -0,0 +1,324 @@
...
+
+        if attributes.attributes.payload_attributes.timestamp != block.header.inner.timestamp {
+            return AttributesMismatch::Timestamp(
+                attributes.attributes.payload_attributes.timestamp,
+                block.header.inner.timestamp,
+            )
+            .into();
+        }
+
+        let difficulty: B256 = block.header.inner.difficulty.into();
```
clabby, line 95: This should be `mix_hash`; `difficulty` is always `0`'d out after the merge. Can validate w/ a `cast block --json | jq`.

crates/node/engine/src/attributes.rs
```diff
@@ -0,0 +1,324 @@
...
+        if attributes.attributes.payload_attributes.prev_randao != difficulty {
+            return AttributesMismatch::PrevRandao(
+                attributes.attributes.payload_attributes.prev_randao,
+                difficulty,
+            )
+            .into();
+        }
```
clabby, line 102:
```suggestion
        let mix_hash = block.header.inner.mix_hash;
        if attributes.attributes.payload_attributes.prev_randao != mix_hash {
            return AttributesMismatch::PrevRandao(
                attributes.attributes.payload_attributes.prev_randao,
                mix_hash,
            )
            .into();
        }
```

crates/node/service/src/actors/engine.rs
```diff
@@ -104,12 +119,38 @@ impl NodeActor for EngineActor {
...
+                    } else {
+                        let task = ConsolidateTask::new(
+                            Arc::clone(&self.client),
+                            Arc::clone(&self.config),
+                            attributes,
+                            true,
+                        );
+                        let task = EngineTask::Consolidate(task);
+                        self.engine.enqueue(task);
+                    }
```
clabby, line 146: Your consolidate task looks to have this branch contained within it - it returns early with just a FCU if the match checks out. Maybe we can remove `needs_consolidation` and just leave that up to the `ConsolidateTask` to handle?

clabby (commented):
Great start

crates/node/engine/src/attributes.rs
refcell: I mirrored what the `op-node` did here yes.

crates/node/service/src/actors/engine.rs
refcell: Yep that's a perfectly fine route to go - probably nicer to head the direction of keeping logic encapsulated within the tasks rather than leaking out into the actor

clabby (approved):

PR #1395, opened 2025-04-09
chore(node/engine): Use thiserror instead of anyhow

palango:
Part of #916 

Open for feedback, especially around naming.


refcell (approved):
Nice work!

PR #1400, opened 2025-04-09
feat: Use `alloy-evm` for stateless block building

clabby:
## Overview

Upgrades the `kona-executor` crate, and its consumers, to share block building logic with `op-reth` through the new `alloy-evm` + `alloy-op-evm` crates.

Also unblocks the upgrade of all `alloy-*` + `op-alloy-*` + `revm` crates.


clabby:
Failing rust tests are related to https://alloy.rs/migrating-to-core-1.0/encoding-decoding-changes/removing-validate-bool. Will need to go forwards with those under the assumption that `alloy-sol-types` no longer actually validates the data..

clabby:
very close

action tests are failing; probably some issue in the spec configuration of the custom EVM builder.

.github/workflows/client_host.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         target: ["native", "asterisc"]
-        name: ["OP Sepolia (Holocene) - Block #22012816"]
+        name: ["OP Sepolia (Holocene) - Block #26215604"]
```
emhane, line 15: unrelated bug fix? would be beneficial if this can be committed in a separate pr to reduce the diff in this pr as much as possible. are there any other parts of the code that can be scoped out to smaller prs and merged independently?

clabby: Not unrelatred. `alloy-evm` accesses more state than our legacy block executor, which changed the contents of the trie witness. This update just pulls in a new test block that was crafted with the updated setup for a complete witness.

bin/client/src/fpvm_evm/precompiles/bls12_g1_add.rs
refcell, line 1: Would be nice to reuse these precompiles.

bin/client/src/fpvm_evm/precompiles/provider.rs
```diff
@@ -0,0 +1,227 @@
...
+    context::{Cfg, ContextTr},
+    handler::{EthPrecompiles, PrecompileProvider},
+    interpreter::{Gas, InputsImpl, InstructionResult, InterpreterResult},
+    precompile::{PrecompileError, PrecompileResult, Precompiles, bls12_381_const, bn128},
+    primitives::{hardfork::SpecId, hash_map::HashMap},
+};
+
+/// The FPVM-accelerated precompiles.
+#[derive(Debug)]
+pub struct OpFpvmPrecompiles<C: Channel> {
```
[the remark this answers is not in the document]
clabby: It is `no_std`, but we have a bit of a special need for a custom `PrecompileProvider` implementation. `OpPrecompiles` is unable to be extended, and our custom `PrecompileProvider` implementation allows us to pass in refs to the preimage oracle comms that it owns which is a nice improvement.

See [this issue](https://github.com/bluealloy/revm/issues/1867) for a bit more context - our current route of [having to use the static oracle reader / hint writer](https://github.com/op-rs/kona/blob/main/bin/client/src/precompiles/utils.rs#L36) isn't really desirable since the host program doesn't have those static pipes, and isn't currently able to run accelerated precompiles.

.github/workflows/client_host.yaml
emhane: hm I see, so it's not possible to do this part first or shrink scope in any other way?

clabby: I can extend the stack upwards, but ultimately yes, without this CI will fail on the PR closest to trunk.

bin/host/src/interop/handler.rs
```diff
@@ -104,7 +105,7 @@ impl HintHandler for InteropHintHandler {
                 // Fetch the blob sidecar from the blob provider.
                 let mut sidecars = providers
                     .blobs
-                    .fetch_filtered_sidecars(&partial_block_ref, &[indexed_hash])
+                    .fetch_filtered_sidecars(&partial_block_ref, [indexed_hash].as_slice())
```
refcell, line 108: Isn't this the same?

crates/proof/executor/src/db/mod.rs
```diff
@@ -415,26 +420,21 @@ mod tests {
...
-        )
+        TrieDB::new(Header::default().seal_slow(), NoopTrieDBProvider, NoopTrieHinter)
     }
 
     #[test]
     fn test_trie_db_take_root_node() {
         let db = new_test_db();
         let root_node = db.take_root_node();
-        assert_eq!(root_node.blind(), B256::default());
+        assert_eq!(root_node.blind(), EMPTY_ROOT_HASH);
```
refcell, line 430: Oh

crates/proof/executor/src/builder/assemble.rs
```diff
@@ -0,0 +1,195 @@
...
+        block_env: &BlockEnv,
+        ex_result: &BlockExecutionResult<OpReceiptEnvelope>,
+        bundle: BundleState,
+    ) -> ExecutorResult<Sealed<Header>> {
+        let timestamp = block_env.timestamp;
+
+        // Compute the roots for the block header.
+        let state_root = self.trie_db.state_root(&bundle)?;
+        let transactions_root = ordered_trie_with_encoder(
+            attrs.transactions.as_ref().expect("Transactions must be non-empty"),
```
refcell, line 42: Maybe worthwhile to leave a `// SAFETY: ` comment why it's ok to panic for empty txs

refcell (commented):
Left a few comments. Nice work - so happy we have test coverage because this is beefy

crates/proof/executor/src/db/mod.rs
clabby: indeed. old api was wonky!

crates/protocol/genesis/src/updates/eip1559.rs
```diff
@@ -31,20 +31,30 @@ impl TryFrom<&SystemConfigLog> for Eip1559Update {
             return Err(EIP1559UpdateError::InvalidDataLen(log.data.data.len()));
         }
 
-        let Ok(pointer) = <sol!(uint64)>::abi_decode(&log.data.data[0..32], true) else {
+        let word: [u8; 32] = log.data.data[0..32].try_into().unwrap();
```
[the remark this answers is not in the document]
clabby: I'll add safety comments, though. These unwraps can never fail.

crates/proof/executor/src/builder/assemble.rs
clabby: Added

bin/host/src/interop/handler.rs
clabby: Removed this + the other occurence, for some reason rust-analyzer yells at me about it but it compiles just fine.

refcell (approved):

Issue #1371, opened 2025-04-07
feat(bin/node): P2P Scoring CLI Flags


refcell:
### Description

The `kona-node` binary already has [a number of P2P cli flags configured](https://github.com/op-rs/kona/blob/main/bin/node/src/flags/p2p.rs#L22) in a separate P2P module, which nicely encapsulates p2p cli options.

This ticket is to extend those options to include peer scoring cli flags like the `op-node` has [here](https://github.com/ethereum-optimism/optimism/blob/b4070c5244c8c7ba6d37b9e31fa7d43ada25070d/op-node/flags/p2p_flags.go#L20C2-L26).

#### Task List

- [x] Peer Scoring Option (`--p2p.scoring`) with `PeerScoreLevel` enum
- [ ] Option to ban peers (`--p2p.ban.peers`)
- [ ] Banning threshold (`--p2p.ban.threshold`)
- [ ] The duration that peers are banned (`--p2p.ban.duration`)

refcell:
The first cli flag is handled by https://github.com/op-rs/kona/pull/1376

PR #1405, opened 2025-04-10
feat(bin/node): implement peer banning

theochap:
## Description
Allows nodes to ban poorly performing peers, similarly to https://github.com/ethereum-optimism/optimism/blob/b4070c5244c8c7ba6d37b9e31fa7d43ada25070d/op-node/flags/p2p_flags.go#L20C2-L26.

This PR:
- Adds CLI flags to allow banning nodes, specify a banning threshold and a banning duration
- Incorporates the P2P flag configuration to the gossip config/driver
- Adds banning logic to the network thread. Ban checks happen every 1 sec by default.

To ban peers on the network thread we:
- Scan the gossip layer for poorly performing peers
- Remove the peers from the gossip layer
- Scan the gossip layer and detect peers to ban. Ban the peers for a given duration.

If merged, this PR completes #1371 

Small note: I have ran clippy so there might be some semi-colons added to unrelated parts of the code. That is minor, can rebase if that's an issue

### DIFF
--- a/crates/node/p2p/src/net/driver.rs
+++ b/crates/node/p2p/src/net/driver.rs
@@ -4,16 +4,17 @@ use alloy_primitives::Address;
 use kona_genesis::RollupConfig;
 use libp2p::TransportError;
 use op_alloy_rpc_types_engine::OpNetworkPayloadEnvelope;
-use std::collections::VecDeque;
+use std::collections::{HashSet, VecDeque};
 use tokio::{
     select,
     sync::{
         broadcast::{Receiver as BroadcastReceiver, Sender as BroadcastSender},
         watch::Sender,
     },
+    time::Duration,
 };
 
-use crate::{Discv5Driver, GossipDriver, NetRpcRequest, NetworkBuilder};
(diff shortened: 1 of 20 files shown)


crates/node/p2p/src/net/driver.rs
```diff
@@ -110,6 +121,63 @@ impl Network {
...
+                            if let Some(addr) = self.gossip.peerstore.remove(&peer_to_remove){
+                                self.gossip.dialed_peers.remove(&addr);
+
+                                return Some(addr);
+                            }
+
+                            None
+                        }).collect::<HashSet<_>>();
+
+                        let ban_futures = handler.table_enrs().await.into_iter().filter_map(
```
refcell, line 164: I think we'll want to spin this into a thread by sending a message through the `handler`. Otherwise this will block the main network driver here which process gossip messages.

refcell (changes requested):
Nice work!

I think we need to avoid blocking the main `Network` driver loop and instead send a non-blocking message to the handler. The handler should spin out a thread to handle banning peers and keep a hold of that handle so the thread isn't dropped.

refcell (approved):
This should be in a good state.

PR #1410, opened 2025-04-10
fix(node/service): Error Bubbling and Shutdown

### LINKED ISSUE #1394: fix(node/service): Error Bubbling from spawn_and_wait
### Description

The `kona-node-service` contains the top-level [`RollupNode` struct](https://github.com/op-rs/kona/blob/main/crates/node/service/src/service/standard/node.rs#L32), orchestrating core functionality for kona's rollup node (aka consensus client). Running the rollup node is trait abstracted such that the node can be run in a variety of different "modes" - validator mode, and sequencer mode. As of right now, only validator mode is supported.

The validator mode contains the core node logic to [spawn and wait on the set of actors](https://github.com/op-rs/kona/blob/main/crates/node/service/src/service/validator.rs#L130) that are part of the node. But this is broken. The `spawn_and_wait!` macro [isn't able to bubble up errors originating from actors](https://github.com/op-rs/kona/blob/main/crates/node/service/src/service/util.rs#L20-L21).

This ticket is to support bubbling up those errors instead of just logging them with tracing.

#### Provenance

https://github.com/op-rs/kona/pull/1363 takes a pass at this but it's quite ugly because the associated `Error` type of the Validator trait will require a number of trait bounds to allow the errors to be coerced.

refcell:
### Description

Previously, the `spawn_and_wait` macro was just ignoring if the actor's start method returned an error since the return object was wrapped in a `JoinTask` result.

This PR fixes the macro to correctly shutdown the node on actor startup error as well as bubbling up the error via its `std::fmt::Debug` impl (not the strongest typing but it allows the `NodeActor` and `ValidatorNodeService` error types to be generic with the only restriction of implementing `std::fmt::Debug`).

Closes #1394.

### DIFF
--- a/crates/node/service/src/service/util.rs
+++ b/crates/node/service/src/service/util.rs
@@ -17,19 +17,28 @@ macro_rules! spawn_and_wait {
             if let Some(actor) = $actor {
                 task_handles.spawn(async move {
                     if let Err(e) = actor.start().await {
-                        // TODO: Bubble up generic error.
-                        tracing::error!(target: "rollup_node", "{e}");
+                        return Err(format!("{e:?}"));
                     }
+                    Ok(())
                 });
             }
         )*
 
         while let Some(result) = task_handles.join_next().await {
-            if let Err(e) = result {
-                tracing::error!(target: "rollup_node", "Critical error in sub-routine: {e}");
-
-                // Cancel all tasks and gracefully shutdown.
-                $cancellation.cancel();
+            match result {
+                Ok(Ok(())) => {
+                    tracing::info!(target: "rollup_node", "Actor finished.");
+                }
+                Ok(Err(e)) => {
+                    tracing::error!(target: "rollup_node", "Critical error in sub-routine: {e}");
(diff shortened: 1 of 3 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1410 👈
* `main`

crates/node/service/src/service/util.rs
```diff
@@ -17,19 +17,28 @@ macro_rules! spawn_and_wait {
...
 
         while let Some(result) = task_handles.join_next().await {
-            if let Err(e) = result {
-                tracing::error!(target: "rollup_node", "Critical error in sub-routine: {e}");
-
-                // Cancel all tasks and gracefully shutdown.
-                $cancellation.cancel();
+            match result {
+                Ok(Ok(())) => {
+                    tracing::info!(target: "rollup_node", "Actor finished.");
```
clabby, line 30: This log is probably redundant; The actors each should log when they get cancelled.

clabby (approved):

PR #1411, opened 2025-04-10
chore(bin/node): In-use Port Erroring

refcell:
### Description

Networking crates do a really bad job at providing a useful error message for simple errors like ports already in use.

This PR adds a small check to the node binary to make it really user friendly if a port is used that's already in use.


refcell:
## 📚 $\text{Stack Overview}$

* #1411 👈
* `main`

emhane (approved):
right so the tcp and udp ports are then dropped at the end of `check_ports` scope and that way application can bind to them again, makes sense

PR #1412, opened 2025-04-11
feat(node/engine): Add transaction checks to the consolidation step

theochap:
## Description
This PR adds transaction checks to the `ConsolidateTask` inside the `EngineActor`. This loosely reproduces the behavior of https://github.com/ethereum-optimism/optimism/blob/c0b1bbcb9802114fb3982e307613197134504c0c/op-node/rollup/attributes/engine_consolidate.go#L42-L62.

In particular this PR:
- Ensures that the safe transactions (obtained from the `DerivationActor` and contained in the `attributes` payload) match the transactions stemming from the oldest unsafe head (contained inside the `block`). We start by comparing the number of transactions, then their contents. If the contents mismatch we return the hashes of the mismatched pair `(safe_tx_hash, unsafe_tx_hash)`.
- Adds unit tests to ensure edge cases are covered. In particular, we want to make sure that we panic if either the `safe` or the `unsafe` transactions are not well formatted.
     - Not well formatted for `safe` transactions mean that the transaction bytes cannot be `serde_json` deserialized to an `OpTxEnvelope`. See, for instance, https://github.com/op-rs/kona/blob/main/crates/proof/executor/src/executor/mod.rs#L205-L213 for an example of transaction payload decoding in the proof executor - we use the same assumption
     - Not well formatted for `unsafe` transactions mean that the block doesn't contain the full transaction content

Merging this PR will close https://github.com/op-rs/kona/issues/1397


crates/node/engine/src/attributes.rs
```diff
@@ -101,7 +103,54 @@ impl AttributesMatch {
...
+
+        // We start by checking that there are the same number of transactions
+        if let Some(attributes_txs) = &attributes.attributes.transactions {
+            // Before checking the number of transactions, we have to make sure that the block has
+            // the right transactions format. We need to have access to the full
+            // transactions to be able to compare their contents.
+            let BlockTransactions::Full(ref block_txs) = block.transactions else {
+                panic!(
+                    "Critical error: the transactions contained in the block don't have the correct format. This is a bug."
+                );
```
refcell, line 116: We should return a mismatch here that says that full transactions weren't provided. Let's add an error log as well so it's easy to debug if this branch is ever hit. In reality we will have the full block txs so it's unlikely, but better than panicking.

crates/node/engine/src/attributes.rs
```diff
@@ -101,7 +103,54 @@ impl AttributesMatch {
...
+            if attributes_txs_len != block_txs_len {
+                return AttributesMismatch::TransactionLen(attributes_txs_len, block_txs_len).into();
+            }
+
+            // Then we need to check that the content of the encoded transactions match
+            // Note that it is safe to zip both iterators because we checked their length
+            // beforehand.
+            for (attr_tx_bytes, block_tx) in attributes_txs.iter().zip(block_txs) {
+                // Let's try to deserialize the attributes transaction
+                let attr_tx = serde_json::from_slice::<OpTxEnvelope>(attr_tx_bytes).expect("Impossible to deserialize transaction from attributes. If we have stored these attributes it means the transactions where well formatted. This is a bug");
```
refcell, line 131: We'll also want to just return a mismatch here + error log. In case something is serialized incorrectly, it should result in a mismatch and not a panic.

crates/node/engine/src/attributes.rs
```diff
@@ -101,7 +103,54 @@ impl AttributesMatch {
             .into();
         }
 
-        // TODO: check transactions
+        // Check transactions
+
+        // We start by checking that there are the same number of transactions
+        if let Some(attributes_txs) = &attributes.attributes.transactions {
```
refcell, line 109: Bit of styling preference, but can we "flatten" these branches? So rather then having the main logic nested in the first if branch, we can grab the `attributes_txs` from the attributes first, otherwise returning.

e.g.
```rust
// First pull out the attributes_txs, otherwise do sad path processing.
let attributes_txs = match &attributes.attributes.transactions {
    Some(attributes_txs) = attributes_txs,
    None => {
        if !block.transactions.is_empty() {
            // Transaction length mismatch! Attributes are empty but block isn't.
            return AttributesMismatch::TransactionLen(0, block.transactions.len()).into();
        } else {
            // The else edge case happens when both transactions from attributes and block are
            // empty. We just have to check that the block transactions is the right format
            if BlockTransactions::Uncle == block.transactions {
                panic!(
                    "Invalid format for the block transactions. The `Uncle` transaction format is not relevant in that context and should not get used here. This is a bug"
                );
            }

            vec![] // Since the block txs will be empty too, this will pass happy path processing.
        }
    }
};

// Now we continue to do the happy path `attributes_txs` processing here.
// ...
```

crates/node/engine/src/attributes.rs
```diff
@@ -101,7 +103,54 @@ impl AttributesMatch {
...
+        } else if !block.transactions.is_empty() {
+            // Transaction length mismatch! Attributes are empty but block isn't.
+            return AttributesMismatch::TransactionLen(0, block.transactions.len()).into();
+        } else {
+            // The else edge case happens when both transactions from attributes and block are
+            // empty. We just have to check that the block transactions is the right format
+            if BlockTransactions::Uncle == block.transactions {
+                panic!(
+                    "Invalid format for the block transactions. The `Uncle` transaction format is not relevant in that context and should not get used here. This is a bug"
+                );
```
refcell, line 151: Let's also return a mismatch here rather than erroring.

The pattern we should follow is generally not to panic in low level processing functions, but instead bubble up the error. The error can be converted into a new wrapper that contains `Temporary`, `Reset`, `Fatal` variants. Then higher up logic can decide what to do with fatal errors.

refcell (changes requested):
Really nice unit tests. Would like to see less panics, but good work so far.

crates/node/engine/src/attributes.rs
theochap: Changed that to an error log + custom error variant

crates/node/engine/src/attributes.rs
theochap: Changed that to an error log + custom error variant

crates/node/engine/src/attributes.rs
theochap: I have tried to reproduce something in that spirit and pulled the attributes_txs outside of the happy path. I extracted the transaction check in a separate method and clearly outlined the format assertions for the block transactions (which were a bit tricky to get). Hope that makes the code more readable!

crates/node/engine/src/attributes.rs
theochap: Same thing, returned a mismatch and emitted an error log

refcell (approved):
Very nice work @theochap

crates/node/engine/src/attributes.rs
refcell: Awesome stuff. That's exactly what I was looking for just to make it easier to read and have less nested branching

PR #1414, opened 2025-04-11
chore: make nightly clippy happy

jackwener:


refcell:
Closing for now - some folks have asked for a happy stable clippy as opposed to nightly. We just need to configure this via a ci check tbh. Since CI isn't failing without this I'm inclined to close this PR until we have a CI job.

PR #1419, opened 2025-04-11
feat(node/engine): check eip1559 parameters inside consolidation

### LINKED ISSUE #1399: feat(node/engine): EIP-1559 Parameter Checks
### Background

> [!NOTE]
>
> This is Part 2 of implementing checks for the `ConsolidateTask` from https://github.com/op-rs/kona/pull/1391

https://github.com/op-rs/kona/pull/1391 adds support to the `EngineActor` to properly handle `OpAttributesWithParent` received through it's unbounded receiver channel. In the `kona-node-service`, the `EngineActor` is wired up such that it's receiver is complimentary to the sender channel in the derivation actor. That is to say, the `DerivationActor` produces payload attributes (in the form of `OpAttributesWithParent` and sends them to the `EngineActor`.

When the `EngineActor` receives these payload attributes, it enqueues a new `ConsolidateTask`. The `ConsolidateTask` was built in https://github.com/op-rs/kona/pull/1391. It handles consolidation of the attributes if needed or proceeds with payload attribute processing. For the purposes of this ticket, what's important here is the consolidation branch. Consolidation occurs when the unsafe head is ahead of the safe head. The oldest unsafe head (the head directly after the safe head) is compared to the attributes from the `DerivationActor`.

 [The OP Stack specs](https://specs.optimism.io/protocol/derivation.html#l1-consolidation-payload-attributes-matching) outline which fields of the attributes and oldest unsafe head need to be checked during consolidation.

In the `kona-node` this occurs in the [`attributes`](https://github.com/op-rs/kona/blob/rf/build-task/crates/node/engine/src/attributes.rs) module of the `crates/node/engine`. The `AttributesMatch` type encapsulates consolidation checks in its [`AttributesMatch::check`](https://github.com/op-rs/kona/blob/rf/build-task/crates/node/engine/src/attributes.rs#L74) method.

Importantly, while https://github.com/op-rs/kona/pull/1391 implemented the `ConsolidateTask`, it did not implement all the checks required for consolidation. The `AttributesMatch` type is missing the following checks.
- [transactions checks](https://github.com/op-rs/kona/blob/rf/build-task/crates/node/engine/src/attributes.rs#L104).
- [eip1559 parameter checks](https://github.com/op-rs/kona/blob/rf/build-task/crates/node/engine/src/attributes.rs#L138)

#### Task

This task is to implement the EIP-1559 parameter checks.

The `op-node` may be used as a reference. It checks eip-1559 parameters [here](https://github.com/ethereum-optimism/optimism/blob/c0b1bbcb9802114fb3982e307613197134504c0c/op-node/rollup/attributes/engine_consolidate.go#L100-L141).

theochap:
## Description
This PR checks the eip1559 inside the consolidation step. I tried to stick as much as possible to the behavior of the op-node and reproduce their edge cases (in particular, *only substitute default params inside the attributes and not the block*).

I have added extensive test coverage to try and reproduce all the different edge-case configurations.

## Note
Merging this PR will close #1399

### DIFF
--- a/crates/node/engine/src/attributes.rs
+++ b/crates/node/engine/src/attributes.rs
@@ -1,11 +1,12 @@
 //! Contains a utility method to check if attributes match a block.
 
+use alloy_eips::eip1559::BaseFeeParams;
 use alloy_network::TransactionResponse;
 use alloy_primitives::{Address, B256, Bytes};
 use alloy_rpc_types_eth::{Block, BlockTransactions, Withdrawals};
 use kona_genesis::RollupConfig;
 use kona_rpc::OpAttributesWithParent;
-use op_alloy_consensus::OpTxEnvelope;
+use op_alloy_consensus::{EIP1559ParamError, OpTxEnvelope, decode_holocene_extra_data};
 use op_alloy_rpc_types::Transaction;
 
 /// Represents whether the attributes match the block or not.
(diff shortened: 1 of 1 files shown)


refcell (approved):
Nice work, this looks correct to me!

Issue #1416, opened 2025-04-11
chore(bin/node): Exchange Engine Capabilities at Startup


refcell:
### Description

When `kona-node` is run against an engine api with a mismatched JWT secret, right now it'll wind up stuck in a loop. It'll keep trying to perform forkchoice updates which fail. But it's not obvious and is actually quite obscure to the user. All they see is "signature invalid" from an rpc transport error, making it seem like a networking issue rather than a misconfigured JWT secret.

One solution I think could be nice is to validate the JWT secret explicitly at startup by using the [`exchange_capabilities`](https://docs.rs/op-alloy-provider/latest/op_alloy_provider/ext/engine/trait.OpEngineApi.html#tymethod.exchange_capabilities) engine api method. Kona already has this available via the `kona-engine` [`EngineClient`](https://github.com/op-rs/kona/blob/main/crates/node/engine/src/client.rs#L221-L229).

I'm thinking in the `bin/node` Node command, add a new method that constructs a new `EngineClient` using the cli arguments. Then validates the JWT secret passed in through the `--l2.jwt-secret` cli argument using the constructed `EngineClient`. Call this new method from inside the node's `run` method prior to building and starting the `RollupNode`.

Something like this:
```rust
impl NodeCommand {
    /// Validate the jwt secret if specified by exchanging capabilities with the engine.
    pub async fn validate_jwt(&self, args: &GlobalArgs) -> anyhow::Result<()> {
        let jwt_secret = self.jwt_secret().ok_or(anyhow::anyhow!("Invalid JWT secret"))?;
        let engine_client = kona_engine::EngineClient::new_http(
            self.l2_engine_rpc.clone(),
            self.l2_provider_rpc.clone(),
            args.rollup_config().map(std::sync::Arc::new).ok_or(anyhow::anyhow!(
                "Failed to get rollup config for chain id: {}",
                args.l2_chain_id
            ))?,
            jwt_secret,
        );
        match engine_client.exchange_capabilities(vec![]).await {
            Ok(_) => {
                debug!("Successfully exchanged capabilities with engine");
                Ok(())
            }
            Err(e) => {
                if e.to_string().contains("signature invalid") {
                    error!(
                        "Engine API JWT secret differs from the one specified by --l2.jwt-secret"
                    );
                    error!(
                        "Ensure that the JWT secret file specified is correct (by default it is `jwt.hex` in the current directory)"
                    );
                }
                bail!("Failed to exchange capabilities with engine: {}", e);
            }
[...]

PR #1421, opened 2025-04-11
chore(bin/node): add jwt startup check

theochap:
## Description
Tiny PR that solves #1416. It adds a check for the validity of the l2 engine jwt *before starting the node*. This PR is merely a copy-paste of the code snippet from the issue with the associated invocation in the node startup.

### DIFF
--- a/bin/node/src/commands/node.rs
+++ b/bin/node/src/commands/node.rs
@@ -6,9 +6,10 @@ use clap::Parser;
 use kona_engine::{EngineKind, SyncConfig, SyncMode};
 use kona_genesis::RollupConfig;
 use kona_node_service::{RollupNode, RollupNodeService};
+use op_alloy_provider::ext::engine::OpEngineApi;
 use serde_json::from_reader;
 use std::{fs::File, path::PathBuf};
-use tracing::debug;
+use tracing::{debug, error};
 use url::Url;
 
 use crate::flags::{GlobalArgs, P2PArgs, RpcArgs};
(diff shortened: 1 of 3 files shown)


bin/node/src/commands/node.rs
```diff
@@ -65,10 +66,43 @@ impl NodeCommand {
...
+            Err(e) => {
+                if e.to_string().contains("signature invalid") {
+                    error!(
+                        "Engine API JWT secret differs from the one specified by --l2.jwt-secret"
+                    );
+                    error!(
+                        "Ensure that the JWT secret file specified is correct (by default it is `jwt.hex` in the current directory)"
+                    );
+                }
+                bail!("Failed to exchange capabilities with engine: {}", e);
```
emhane, line 97: is there another way to do this than match on a specific substring? at least leave a comment about which impl of `OpEngineApi` this is assuming this uses and open an issue to clean up debt later.

emhane (changes requested):

bin/node/src/commands/node.rs
refcell: Let's make this a follow-up task. For now it's only in the happy path that these errors are printed, so it shouldn't be misleading at the least.

refcell (approved):

bin/node/src/commands/node.rs
theochap: @emhane I tried to figure out the exact error code returned by the alloy json rpc when there is an invalid jwt signature but I haven't managed to get around any sort of documentation so far. I will open an issue and get back to it once I have full access to the node for reproducibility.

theochap: https://github.com/op-rs/kona/issues/1429

PR #1422, opened 2025-04-11
feat(bin/node): add more metrics flags

### LINKED ISSUE #1402: feat(bin/node): Metrics Flags
### Description

Consolidate and extend metrics flags like `RpcArgs` does.

Right now, the `kona-node` only allows for configuring the metrics port via a global cli flag [here](https://github.com/op-rs/kona/blob/main/bin/node/src/flags/globals.rs#L19-L26).

But the `op-node` provides a few more options to configure metrics.
- `--metrics.enabled`
- `--metrics.addr`
- `--metrics.port`

### Output

This ticket is to introduce a new `MetricsArgs` struct that contains the metrics cli flags. This should be in a new `metrics` module similar to how the [`rpc`](https://github.com/op-rs/kona/blob/main/bin/node/src/flags/rpc.rs#L11) module contains all RPC flags.

Move the [`metrics_port`](https://github.com/op-rs/kona/blob/main/bin/node/src/flags/globals.rs#L19-L26) cli flag from `GlobalArgs` into the new `MetricsArgs` struct. Add the `--metrics.enabled` and `--metrics.addr` flags to the `MetricsArgs` struct.

Update the `kona-node` to only call [`init_prometheus`](https://github.com/op-rs/kona/blob/main/crates/utilities/cli/src/prometheus.rs#L7-L20) if the metrics are enabled. Update the [`init_prometheus`](https://github.com/op-rs/kona/blob/main/crates/utilities/cli/src/prometheus.rs#L7-L20) method to pass in the `--metrics.addr` cli argument as the socket address instead of the hardcoded `0.0.0.0`.

theochap:
## Description
This PR extends configuration options for prometheus metrics inside the kona-node. In particular:
- Adds an `metrics.enabled` configuration to allow enabling/disabling prometheus metrics
- Adds an `metrics.address` configuration flag to specify a custom ip address to instantiate prometheus
- Moves the `metrics.port` configuration option outside of the global configuration flags.

I left the `MetricsArgs` struct in the main command (ie, I did not move it to the subcommands) to avoid copy-pasting the metrics configuration for each subcommand.

## Note
Merging this PR will fix #1402


refcell (approved):
Very nice :D

Issue #1340, opened 2025-03-29
fix(node/p2p): Block Validation Rules


refcell:
### Description

The OP Stack requires that unsafe blocks are validated according to [the rules described in the specs](https://specs.optimism.io/protocol/rollup-node-p2p.html#block-validation).

Kona currently incorrectly validates blocks in the [`BlockHandler`s `block_valid` method](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/gossip/handler.rs#L117-L138).

emhane:
what are the action items here?

refcell:
> what are the action items here?

> Kona currently incorrectly validates blocks in the [BlockHandlers block_valid method](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/gossip/handler.rs#L117-L138).

PR #1424, opened 2025-04-13
feat(p2p):  enhance block validation

jackwener:
part of #1340

- Introduced a new `PayloadValidation` trait for validating execution payloads.
- Implemented the trait for `OpExecutionPayload` with various validation methods.
- Updated block validation logic in `BlockHandler` to incorporate payload validation rules based on topic versions.
- Refactored the `block_valid` method to include checks for withdrawals and blob gas usage according to the OP Stack specification.

### DIFF
--- a/crates/node/p2p/src/payload.rs
+++ b/crates/node/p2p/src/payload.rs
@@ -1,0 +1,227 @@
+use op_alloy_rpc_types_engine::OpExecutionPayload;
+
+/// Trait for validating execution payloads
+pub trait PayloadValidation {
+    /// Returns true if the payload has withdrawals
+    fn has_withdrawals(&self) -> bool;
+
+    /// Returns true if the payload has a withdrawals list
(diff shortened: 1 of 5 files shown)


crates/node/p2p/src/payload.rs
```diff
@@ -0,0 +1,226 @@
...
+
+    fn has_non_zero_excess_blob_gas(&self) -> bool {
+        match self {
+            OpExecutionPayload::V1(_) | OpExecutionPayload::V2(_) => false,
+            OpExecutionPayload::V3(payload) => payload.excess_blob_gas != 0,
+            OpExecutionPayload::V4(payload) => payload.payload_inner.excess_blob_gas != 0,
+        }
+    }
+
+    fn is_block_hash_valid(&self) -> bool {
```
jackwener, line 95: I'm not sure how to verify block_hash

PR #1434, opened 2025-04-15
chore: Use `OpPayloadAttributes::recovered_transactions`

clabby:
## Overview

Upgrades `op-alloy` and uses the new `OpPayloadAttributes::recovered_transactions`, rather than `decode_and_recover_tx`.


mattsse (approved):
progress

refcell (approved):

PR #1435, opened 2025-04-15
bug(node/p2p): fix bootstrapping for enode addresses

### LINKED ISSUE #1413: fix(node/p2p): Bootstrapping
### Description

Currently, `kona-p2p` is unable to bootstrap using Enode (aka Multiaddr) node identifiers. Since we had bootnode ENRs for op mainnet, that was fine - only those ENRs were [used by the discovery service during bootstrapping](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/discv5/driver.rs#L96-L103). Enodes are ignored.

This ticket is to find a way to bootstrap with Enodes. For testnet, `kona-p2p` only has a [set of enode identifiers](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/peers/nodes.rs#L106-L119). Basically what happens when we run the kona-p2p networking stack for op-sepolia or some other testnet chain in the superchain is it is unable to add any peers in the discovery table - leaving the network stuck.

theochap:
## Description
This PR provides a fix for the bootstrapping of nodes given an enode address. 
In particular:
- it fixes the conversion from "local" `PeerId` to "p2plib" `PeerId`. These two types represent the same underlying concept (secp256k1 public key) but using different encodings (the local `PeerId` is the uncompressed representation of the public key, while the "p2plib" `PeerId` is a more complex representation, involving protobuf encoding and bitcoin encoding, defined here: https://github.com/libp2p/specs/blob/master/peer-ids/peer-ids.md).
- It slightly changes the logic of the `bootstrap` method to ensure that both enode and enr encoded bootstrap nodes are added.
- It adds util methods for the conversion from local representation to "p2plib" representation of `PeerId`. 
- It adds unit testing to highlight the difference of format and ensure the conversion works.

As a follow-up, we should probably try and rename the "local" `PeerId`, or add type-safety/newtypes to ensure that we limit the likelihood of doing wrongful conversions in the future.

## Associated issue
Merging this PR will close #1413

### DIFF
--- a/crates/node/p2p/src/discv5/driver.rs
+++ b/crates/node/p2p/src/discv5/driver.rs
@@ -2,6 +2,7 @@
 
 use derive_more::Debug;
 use discv5::{Discv5, Enr, Event, enr::NodeId};
+use libp2p::Multiaddr;
 use tokio::{
     sync::mpsc::channel,
     time::{Duration, sleep},
@@ -89,6 +90,34 @@ impl Discv5Driver {
         }
     }
 
+    /// Adds a bootnode to the discv5 service given an enode address.
+    async fn add_enode(&mut self, enode: Multiaddr) -> Option<Enr> {
+        let enr = match self.disc.request_enr(enode.clone()).await {
+            Ok(enr) => enr,
+            Err(err) => {
+                error!(
+                    ?enode,
+                    %err,
+                    "failed to add boot node"
+                );
+
+                return None;
+            }
+        };
+
+        if let Err(err) = self.disc.add_enr(enr.clone()) {
+            warn!(
+                ?enode,
+                %err,
+                "failed to add boot node"
+            );
+
+            return None;
+        }
+
+        Some(enr)
+    }
+
     /// Bootstraps the [`Discv5`] service with the bootnodes.
     async fn bootstrap(&mut self) {
         let nodes = BootNodes::from_chain_id(self.chain_id);
@@ -125,25 +154,37 @@ impl Discv5Driver {
                 Err(e) => debug!("Failed to add ENR to discv5 table: {:?}", e),
             }
         }
-        info!("Added {} ENRs to discv5 | {} in bootstore", count, self.store.len());
 
         // Merge the bootnodes into the bootstore.
         self.store.merge(boot_enrs);
+        debug!(target: "p2p::discv5",
+            new=%count,
+            total=%self.store.len(),
+            "Added new ENRs to discv5 bootstore"
+        );
+
+        let mut boot_enodes_enrs = Vec::new();
 
         for node in nodes.0 {
             match node {
                 BootNode::Enode(enode) => {
-                    if let Err(err) = self.disc.request_enr(enode.to_string()).await {
-                        debug!(
-                            ?enode,
-                            %err,
-                            "failed adding boot node"
-                        );
-                    }
+                    let Some(enr) = self.add_enode(enode).await else {
+                        continue;
+                    };
+
+                    boot_enodes_enrs.push(enr);
                 }
-                _ => { /* ignore: bootnode enrs already added */ }
+                BootNode::Enr(_) => { /* ignore: bootnode enrs already added */ }
             }
         }
+
+        // Merge the bootnodes into the bootstore.
+        self.store.merge(boot_enodes_enrs);
+        debug!(target: "p2p::discv5",
+            new=%count,
+            total=%self.store.len(),
+            "Added new ENRs to discv5 bootstore"
+        );
     }
 
     /// Sends ENRs from the boot store to the enr receiver.
@@ -261,8 +302,16 @@ impl Discv5Driver {
 
 #[cfg(test)]
 mod tests {
+    use discv5::{enr::CombinedPublicKey, handler::NodeContact};
+    use kona_genesis::OP_SEPOLIA_CHAIN_ID;
+
+    use crate::{NodeRecord, OP_RAW_TESTNET_BOOTNODES};
+
     use super::*;
-    use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+    use std::{
+        net::{IpAddr, Ipv4Addr, SocketAddr},
+        str::FromStr,
+    };
 
     #[tokio::test]
     async fn test_discv5_driver() {
@@ -275,5 +324,46 @@ mod tests {
         let _ = discovery.start();
         // The service starts.
         // TODO: verify with a heartbeat.
+    }
+
+    #[tokio::test]
+    async fn test_discv5_driver_bootstrap_testnet() {
+        let socket = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 9099);
+        let mut discovery = Discv5Driver::builder()
+            .with_address(socket)
+            .with_chain_id(OP_SEPOLIA_CHAIN_ID)
+            .build()
+            .expect("Failed to build discovery service");
+
+        discovery.init().await;
+
+        discovery.bootstrap().await;
+
+        // Verify bootstore has entries.
+        assert!(!discovery.store.is_empty(), "No nodes were added to the bootstore");
+        // Verify discv5 has entries in its enr table.
+        assert!(!discovery.disc.table_entries_enr().is_empty());
+
+        // It should have the same number of entries as the testnet table.
+        let testnet = OP_RAW_TESTNET_BOOTNODES;
+
+        // Verifies that the public keys are the same.
+        let testnet_pkeys: Vec<CombinedPublicKey> = testnet
+            .iter()
+            .map(|node| {
+                let node_record = NodeRecord::from_str(node).unwrap();
+                let bootnode = BootNode::from_unsigned(node_record).unwrap();
+                let node_contact =
+                    NodeContact::try_from_multiaddr(bootnode.to_multiaddr().unwrap()).unwrap();
+
+                // The public key contained in the NodeContact is used to connect to the
+                // bootnode.
+                node_contact.public_key()
+            })
+            .collect();
+
+        for enr in discovery.disc.table_entries_enr() {
+            assert!(testnet_pkeys.contains(&enr.public_key()));
+        }
     }
 }
--- a/crates/node/p2p/src/peers/any.rs
+++ b/crates/node/p2p/src/peers/any.rs
@@ -2,8 +2,10 @@
 
 use crate::{NodeRecord, PeerId};
 use derive_more::From;
-use discv5::{Enr, enr::EnrPublicKey, libp2p_identity::ParseError};
+use discv5::{Enr, enr::EnrPublicKey};
 use libp2p::swarm::dial_opts::DialOpts;
+
+use super::utils::peer_id_to_secp256k1_pubkey;
 
 /// A peer that can come in [`Enr`] or [`NodeRecord`] form.
 #[derive(Debug, Clone, From, Eq, PartialEq, Hash)]
@@ -16,8 +18,16 @@ pub enum AnyNode {
     PeerId(PeerId),
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum DialOptsError {
+    #[error("Invalid peer id. Error: {0}")]
+    InvalidPeerId(secp256k1::Error),
+    #[error("Invalid public key. Error: {0}")]
+    InvalidPublicKey(#[from] discv5::libp2p_identity::DecodingError),
+}
+
 impl AnyNode {
-    /// Returns the peer id of the node.
+    /// Returns the local peer id of the node.
     pub fn peer_id(&self) -> PeerId {
         match self {
             Self::NodeRecord(record) => record.id,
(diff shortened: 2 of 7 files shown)


crates/node/p2p/src/discv5/driver.rs
```diff
@@ -125,25 +155,29 @@ impl Discv5Driver {
                 Err(e) => debug!("Failed to add ENR to discv5 table: {:?}", e),
             }
         }
-        info!("Added {} ENRs to discv5 | {} in bootstore", count, self.store.len());
 
         // Merge the bootnodes into the bootstore.
         self.store.merge(boot_enrs);
+        info!("Added {} ENRs to discv5 | {} in bootstore", count, self.store.len());
```
emhane, line 161: can be downgraded to debug log level. and pls add so we can filter when debugging. we also tend to use the k=v format for values. see lighthouse or reth for reference.
```suggestion
        debug!(target: "p2p::discv5",
            new=%count,
            total=%self.store.len(),
            "Added new ENRs to discv5 bootstore"
        );
```
then we debug like,
`$ RUST_LOG=p2p::discv5=trace <kona-cmd>`

crates/node/p2p/src/discv5/driver.rs
```diff
@@ -89,6 +90,35 @@ impl Discv5Driver {
         }
     }
 
+    /// Adds a bootnode to the discv5 service given an enode address.
+    async fn add_enode(&mut self, enode: Multiaddr) -> Option<Enr> {
+        let enr = match self.disc.request_enr(enode.clone()).await {
+            Ok(enr) => enr,
+            Err(err) => {
+                println!("Failed to add enr: {:?}", err);
```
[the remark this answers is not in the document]
emhane: can keep the print statement for now but should be removed, ref https://github.com/op-rs/kona/issues/1437

crates/node/p2p/src/discv5/driver.rs
```diff
@@ -89,6 +90,35 @@ impl Discv5Driver {
...
+                    %err,
+                    "failed to add boot node"
+                );
+
+                return None;
+            }
+        };
+
+        if let Err(err) = self.disc.add_enr(enr.clone()) {
+            error!(
```
emhane, line 109: can be downgraded to `warn`. `error` is reserved for events which will make the node unable to progress. like for example zero boot nodes successfully added from boot node list.

crates/node/p2p/src/discv5/driver.rs
```diff
@@ -125,25 +155,29 @@ impl Discv5Driver {
...
+                    boot_enodes_enrs.push(enr);
                 }
-                _ => { /* ignore: bootnode enrs already added */ }
+                BootNode::Enr(_) => { /* ignore: bootnode enrs already added */ }
             }
         }
+
+        // Merge the bootnodes into the bootstore.
+        self.store.merge(boot_enodes_enrs);
+        info!("Added {} ENRs to discv5 | {} in bootstore", count, self.store.len());
```
emhane: can also be downgraded to debug, see above comment

crates/node/p2p/src/discv5/driver.rs
```diff
@@ -276,4 +317,47 @@ mod tests {
         // The service starts.
         // TODO: verify with a heartbeat.
     }
+
+    #[tokio::test]
+    async fn test_discv5_driver_bootstrap_testnet() {
+        let socket = SocketAddr::new(IpAddr::V4(Ipv4Addr::new(0, 0, 0, 0)), 9099);
+        let mut discovery = Discv5Driver::builder()
+            .with_address(socket)
+            .with_chain_id(11155420)
```
emhane, line 333: pls replace magic number with constant

crates/node/p2p/src/discv5/driver.rs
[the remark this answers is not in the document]
theochap: Good catch, definitely going to remove that

crates/node/p2p/src/discv5/driver.rs
```diff
@@ -276,4 +317,47 @@ mod tests {
...
+            .expect("Failed to build discovery service");
+
+        discovery.init().await;
+
+        discovery.bootstrap().await;
+
+        // Verify bootstore has entries.
+        assert!(!discovery.store.is_empty(), "No nodes were added to the bootstore");
+        // Verify discv5 has entries in its enr table.
+        assert!(!discovery.disc.table_entries_enr().is_empty());
```
emhane, line 344: this test runs on WAN it appears as opposed to the test being rigged on localhost spinning up boot nodes manually, this means this test is likely to be flaky and some point in future and fail CI. can we therefore limit this test to run on a nightly cronjob instead of on each pr?

emhane (changes requested):

crates/node/p2p/src/discv5/driver.rs
refcell: Not a huge deal since it's in a test, but `kona-genesis` [exports the `OP_SEPOLIA_CHAIN_ID` constant](https://github.com/op-rs/kona/blob/main/crates/protocol/genesis/src/chain/mod.rs#L7).

crates/node/p2p/src/peers/utils.rs
```diff
@@ -17,3 +19,76 @@ pub fn enr_to_multiaddr(enr: &Enr) -> Option<Multiaddr> {
...
+    secp256k1::PublicKey::from_slice(&s)
+}
+
+/// Converts an uncoded [`PeerId`] to a [`libp2p::PeerId`].
+pub fn local_id_to_p2p_id(peer_id: PeerId) -> Result<libp2p::PeerId, secp256k1::Error> {
+    // The libp2p library works with compressed public keys.
+    let encoded_pk_bytes = id2pk(peer_id)?.serialize();
+    let pk: discv5::libp2p_identity::PublicKey =
+        discv5::libp2p_identity::secp256k1::PublicKey::try_from_bytes(&encoded_pk_bytes)
+            .expect("secp256k1 serialization should be valid")
```
emhane, line 49: this is a place where it is tolerable to panic

crates/node/p2p/src/peers/utils.rs
```diff
@@ -17,3 +19,76 @@ pub fn enr_to_multiaddr(enr: &Enr) -> Option<Multiaddr> {
...
+            .into();
+
+    Ok(pk.to_peer_id())
+}
```
emhane, line 53: make sure these functions are full test covered, i.e. all the error branches, seems like code cov is happy with them already :)

crates/node/p2p/src/peers/utils.rs
```diff
@@ -17,3 +19,76 @@ pub fn enr_to_multiaddr(enr: &Enr) -> Option<Multiaddr> {
...
+
+        let p2p_public_key: discv5::libp2p_identity::PublicKey =
+            p2p_keypair.public().clone().into();
+
+        assert_eq!(peer_id, p2p_public_key.to_peer_id());
+    }
+
+    #[test]
+    fn test_hardcoded_peer_id() {
+        let pub_key_str = "548f715f3fc388a7c917ba644a2f16270f1ede48a5d88a4d14ea287cc916068363f3092e39936f1a3e7885198bef0e5af951f1d7b1041ce8ba4010917777e71f";
```
emhane, line 78: can be a const

crates/node/p2p/src/peers/utils.rs
```diff
@@ -17,3 +19,76 @@ pub fn enr_to_multiaddr(enr: &Enr) -> Option<Multiaddr> {
...
+    use crate::{PeerId, peers::utils::id2pk};
+
+    use super::*;
+
+    #[test]
+    fn test_convert_local_peer_id_to_multi_peer_id() {
+        let p2p_keypair = discv5::libp2p_identity::secp256k1::Keypair::generate();
+        let uncompressed = p2p_keypair.public().to_bytes_uncompressed();
+        let local_peer_id = PeerId::from_slice(&uncompressed[1..]);
+        let peer_id = local_id_to_p2p_id(local_peer_id).unwrap();
```
emhane, line 68: agreed that the word local to p2p is unhelpful here, could you pls leave a smol code comment what local vs p2p refers to here

crates/node/p2p/src/discv5/driver.rs
theochap: Done!

crates/node/p2p/src/discv5/driver.rs
theochap: Done

crates/node/p2p/src/discv5/driver.rs
theochap: Done

emhane (changes requested):
the pr has a really nice description, can we get some of that text better into docs?

crates/node/p2p/src/peers/utils.rs
theochap: Done

crates/node/p2p/src/peers/utils.rs
theochap: Changed that to graceful error

theochap:
> the pr has a really nice description, can we get some of that text better into docs?

I put some part of the description inside the method that converts the local peer id to the libp2p peer id (`local_id_to_p2p_id`)

crates/node/p2p/src/peers/utils.rs
theochap: Added a doc comment

crates/node/p2p/src/discv5/driver.rs
emhane: sure, this test isn't controlled fully in this scope because it depends on the liveness of boot nodes in the testnet. the testnet is not controlled by this test. if those nodes go offline, then ci will fail and we can't merge prs.

emhane: it doesn't make sense to make ci passing dependent on if those boot nodes are online or not.

does that make sense?

crates/node/p2p/src/peers/any.rs
```diff
@@ -27,17 +37,21 @@ impl AnyNode {
...
+        let pub_key: discv5::libp2p_identity::PublicKey =
+            discv5::libp2p_identity::secp256k1::PublicKey::try_from_bytes(
+                &id2pk(self.peer_id()).map_err(DialOptsError::InvalidPeerId)?.serialize(),
+            )?
+            .into();
+
+        let libp2p_id = libp2p::PeerId::from_public_key(&pub_key);
         Ok(libp2p_id.into())
     }
 }
```
emhane, line 50: here we want full test coverage too or are we 100% it is infallible because it's converting the local peer id which will always be the correct number of bytes, so we can use expect here too?

emhane (changes requested):

crates/node/p2p/src/discv5/driver.rs
theochap: Just pushed a change https://github.com/op-rs/kona/pull/1435/commits/e27c1490338377020aa1c1ef1b9979fb516a9569#diff-e3c2d30d9bcd63009fc8fd07a05b5a190843f0c2b78f8c58f5d63493288c64ecR345

crates/node/p2p/src/peers/any.rs
theochap: Exactly: in theory we should never raise the outer err because we know that `PeerId::serialize` will always yield the right byte sequence

emhane: apparently you can use
```rust
// coverage:ignore-start
fn not_covered() {
    println!("This is ignored for coverage");
}
// coverage:ignore-end
```
for codecov to ignore these lines, then you don't need to replace the error propagation for the panics here

emhane (approved):

crates/node/p2p/src/peers/utils.rs
```diff
@@ -17,3 +19,93 @@ pub fn enr_to_multiaddr(enr: &Enr) -> Option<Multiaddr> {
...
+/// bytes with the `SECP256K1_TAG_PUBKEY_UNCOMPRESSED` tag.
+#[inline]
+pub fn id2pk(id: PeerId) -> Result<secp256k1::PublicKey, secp256k1::Error> {
+    // NOTE: B512 is used as a PeerId because 512 bits is enough to represent an uncompressed
+    // public key.
+    let mut s = [0u8; secp256k1::constants::UNCOMPRESSED_PUBLIC_KEY_SIZE];
+    s[0] = SECP256K1_TAG_PUBKEY_UNCOMPRESSED;
+    s[1..].copy_from_slice(id.as_slice());
+    secp256k1::PublicKey::from_slice(&s)
+}
```
emhane, line 41: hm, this is identical to the reth code could be good to switch it up so as to not violate copyright or add their licenses to this file. up to you, each is on there own wrt copyrights as by contract I believe.

theochap: I thought this was ok because we have the following disclaimer here: https://github.com/op-rs/kona/blob/e7cadd2600183b52ffb2eed75102ddf981c91fa4/crates/node/p2p/src/peers/mod.rs#L1-L4. Isn't that enough?

theochap: Changed the code to not match reth's

PR #1438, opened 2025-04-15
feat(executor): Add example + docs for generating new test fixtures

clabby:
## Overview

Adds some documentation and a new example binary for generating new static test fixtures for the `StatelessL2Builder` in `kona-executor`. Also improves the test to automatically pick up on the test fixtures in `kona-executor`'s `testdata` directory, rather than having to manually specify which ones to run.

### DIFF
--- a/.github/workflows/rust_ci.yaml
+++ b/.github/workflows/rust_ci.yaml
@@ -141,6 +141,8 @@ jobs:
         with:
           submodules: true
       - uses: ./.github/actions/setup
+      - name: Install clang
+        run: sudo apt install clang libclang-dev
       - uses: taiki-e/install-action@cargo-hack
       - name: cargo hack
         run: just hack
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2681,6 +2681,20 @@ dependencies = [
 ]
 
 [[package]]
+name = "execution-fixture"
+version = "0.0.0"
+dependencies = [
+ "anyhow",
+ "clap",
+ "kona-cli",
+ "kona-executor",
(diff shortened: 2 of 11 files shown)


crates/proof/executor/TEST_FIXTURES.md
clabby, line 1: Added in a new file so that these docs don't show up in the rustdoc.

theochap (approved):

crates/proof/executor/TEST_FIXTURES.md
```diff
@@ -0,0 +1,33 @@
...
+#[tokio::test(flavor = "multi_thread")]
+async fn create_fixture() {
+    let fixture_creator = crate::test_utils::ExecutorTestFixtureCreator::new(
+        "<rpc_url>",
+        <block_number>,
+        PathBuf::from(env!("CARGO_MANIFEST_DIR")).join("testdata"),
+    );
+    fixture_creator.create_static_fixture().await;
+}
+```
```
refcell, line 25: One suggestion is we could hoist this into an example. Kona already has a few [workspace-level examples](https://github.com/op-rs/kona/tree/main/examples) that can be run with `cargo run --example <example>`. This way we don't have commented out code

refcell (commented):
Small suggestion. Might be nice to place this in the kona book instead of a random markdown file

crates/proof/executor/TEST_FIXTURES.md
clabby: this sounds better - will do in the am

PR #1439, opened 2025-04-15
chore(ci): Add action to free up disk space in action test job

clabby:
## Overview

We're still seeing issues with the action tests job running out of disk space, even after #1428. This is another step to mitigate the issue.

[`free-disk-space-ubuntu`](https://github.com/marketplace/actions/free-disk-space-ubuntu) removes a lot of the pre-installed tools that aren't necessary for the job to run.

### DIFF
--- a/.github/workflows/action_tests.yaml
+++ b/.github/workflows/action_tests.yaml
@@ -14,6 +14,10 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          large-packages: false
       - uses: taiki-e/install-action@just
       - uses: dtolnay/rust-toolchain@stable
         with:


clabby:
Screenshot 2025-04-15 at 6 15 27 PM

The free runner rampage lives on!

excited-fuego

.github/workflows/action_tests.yaml
```diff
@@ -14,6 +14,12 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v4
+      - name: Show disk usage (pre-cleanup)
+        run: df . -h
```
theochap, line 18: nit: do we still want to keep the disk usage pre-cleanup in CI? I don't think that is super important but unless we want to keep it around for a few PRs to see the effect of the cleanup in practice, we probably can get rid of it!

theochap (approved):

.github/workflows/action_tests.yaml
clabby: Yeah we can take it or leave it. Still playing with this, will move into draft for now :) https://github.com/op-rs/kona/pull/1439#issuecomment-2807823283

PR #1442, opened 2025-04-16
chore(comp): Enable `test-utils` feature with `cfg(test)`

clabby:
## Overview

Makes `test-utils` enabled in `cfg(test)` always in `kona-comp`, removing the need to include the module if either `cfg(test)` or `feature = "test-utils` is enabled.


refcell (approved):
Nice pattern

PR #1443, opened 2025-04-16
chore(derive): Enable `test-utils` feature with `cfg(test)`

clabby:
## Overview

Makes `test-utils` enabled in `cfg(test)` always in `kona-derive`, removing the need to include the module if either `cfg(test)` or `feature = "test-utils` is enabled.

Also removes the conditional `no_std` status of the crate; All of the `test-utils` here happen to be `no_std` compatible 😄


refcell (approved):
Very nice

PR #1449, opened 2025-04-16
fix(node): Configurable Bootstore Path

refcell:
### Description

Updates the `P2PArgs` CLI flags to contain a new `--p2p.bootstore` flag that accepts an optional path.

This CLI flag allows the user to set the path at which `kona-node` should store and read the "bootstore".

Previously, this was set to `None` manually in `kona-p2p` which effectively used the home directory.


refcell:
## 📚 $\text{Stack Overview}$

* #1449 👈
* `main`

clabby (approved):

PR #1453, opened 2025-04-17
feat(node/p2p): add manual block hash checks

theochap:
## Description
As a follow-up of #1451, this PR adds manual block hash checks based on the block validation rules from the [spec](https://specs.optimism.io/protocol/rollup-node-p2p.html#block-validation). 

In particular, this PR:
- Manually recomputes the block header from the network payload envelope. 
- Computes the hash of the block header and compares it with the existing one.
- Adds unit testing.

Small note: I have not been able to use the `try_into_block` method from `alloy` on the `ExecutionPayload` types because I don't have access to the generic transaction type contained in the `Block` struct (in `alloy-consensus`). I had to manually port some of the methods from alloy. We could probably add a method in `alloy`/`op-alloy` to compute the block header from without having to provide the generic transaction type.

## Note
This PR should be reviewed after #1451. Merging this PR will make progress towards #1340

### DIFF
--- a/crates/node/p2p/src/gossip/block_validity.rs
+++ b/crates/node/p2p/src/gossip/block_validity.rs
@@ -23,10 +23,10 @@ pub enum BlockInvalidError {
     Signature,
     #[error("Invalid signer, expected: {expected}, received: {received}")]
     Signer { expected: Address, received: Address },
-    // TODO: add the rest of the errors variants in follow-up PRs
     /// Invalid block.
     #[error(transparent)]
     InvalidBlock(#[from] OpPayloadError),
+    // TODO: add the rest of the errors variants in follow-up PRs
 }
 
 impl From<BlockInvalidError> for MessageAcceptance {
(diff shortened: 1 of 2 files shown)


crates/node/p2p/src/gossip/block_validity.rs
```diff
@@ -0,0 +1,469 @@
...
+            return Err(BlockInvalidError::Signer { expected: msg_signer, received: block_signer });
+        }
+
+        Ok(())
+    }
+}
+
+/// Checks that the block hash is valid. For that we need to recreate the header, recompute the
+/// associated block hash and compare it with the one in the block.
+pub fn check_block_hash(
```
refcell, line 90: I think we should be able to use `try_into_block` here since this is guaranteed to be an L2 block gossiped over p2p. That means, the block type has to be `Block<OpTxEnvelope>`. We can just strongly type this. I've put up a draft pr to demonstrate using `try_into_block` [here](https://github.com/op-rs/kona/pull/1457). A few tests are failing which likely tells me the block construction in the tests should be updated. Most likely, we can construct a block without unstructured data. Maybe we have a mock test block in `op-alloy-rpc-types-engine`.

theochap: Great suggestion! I have incorporated your suggestion and fixed the tests!

refcell (approved):

PR #1454, opened 2025-04-17
feat(node/p2p): add version specific block checks

theochap:
## Description
This PR adds version specific block checks to the block validation method. It mostly ensures that the topic specific rules from the [spec](https://specs.optimism.io/protocol/rollup-node-p2p.html#block-validation) are respected.

Note that most of the rules are verified through type-checking and using the `ExecutionPayload::try_into_block` methods. So there is only a couple more rules left to check. I tried to detail as much as possible the remaining rules to check and why we can skip explicit checks for some of them.

## Note
It should be reviewed after #1453. Merging it will make progress towards #1340.

### DIFF
--- a/crates/node/p2p/src/gossip/block_validity.rs
+++ b/crates/node/p2p/src/gossip/block_validity.rs
@@ -2,7 +2,7 @@ use std::time::SystemTime;
 
 use alloy_consensus::Block;
 use alloy_primitives::{Address, B256};
-use alloy_rpc_types_engine::{ExecutionPayloadV2, ExecutionPayloadV3, PayloadError};
+use alloy_rpc_types_engine::{ExecutionPayloadV3, PayloadError};
 use libp2p::gossipsub::MessageAcceptance;
 use op_alloy_consensus::OpTxEnvelope;
 use op_alloy_rpc_types_engine::{
@@ -26,8 +26,6 @@ pub enum BlockInvalidError {
     /// Invalid block.
     #[error(transparent)]
     InvalidBlock(#[from] OpPayloadError),
-    #[error("Payload is on v2+ topic, but has non-empty withdrawals")]
-    Withdrawals,
     #[error("Payload is on v3+ topic, but has empty parent beacon root")]
     ParentBeaconRoot,
     #[error("Payload is on v3+ topic, but has non-zero blob gas used")]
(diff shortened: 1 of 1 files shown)


crates/node/p2p/src/gossip/block_validity.rs
```diff
@@ -90,13 +101,71 @@ impl BlockHandler {
...
+
+    // Same as v1, except:
+    // 1. The block should have an empty withdrawals list. This is checked during the call to
+    //    [`OpExecutionPayload::try_into_block`].
+
+    // Same as v2, except:
+    // 1. The block should have a zero blob gas used
+    // 2. The block should have a zero excess blob gas
+    // 3. The block should have a non empty parent beacon block root
+    fn validate_v3(
```
refcell, line 126: This is quite functional, in the style of golang.

I almost wish the [`OpExecutionPayload`](https://docs.rs/op-alloy-rpc-types-engine/latest/op_alloy_rpc_types_engine/payload/enum.OpExecutionPayload.html) (inside the [`OpNetworkPayloadEnvelope`](https://docs.rs/op-alloy-rpc-types-engine/latest/op_alloy_rpc_types_engine/struct.OpNetworkPayloadEnvelope.html)) had op-specific types.

That way we could have `validate()` methods on all the inner payload types.

I wonder if @mattsse has any preference on this.

So:

```
pub enum OpExecutionPayload {
    V1(ExecutionPayloadV1),
    V2(ExecutionPayloadV2),
    V3(ExecutionPayloadV3),
    V4(OpExecutionPayloadV4),
}
```

Would become this, where the inner types are just wrappers. Though, might be more hassle than it's worth having to update everywhere this is used.

```
pub enum OpExecutionPayload {
    V1(OpExecutionPayloadV1),
    V2(OpExecutionPayloadV2),
    V3(OpExecutionPayloadV3),
    V4(OpExecutionPayloadV4),
}
```

refcell: For now, I think we can merge this PR and if this is a direction we want to head, we can follow up since it'll require a dep update.

theochap: Totally agree, I think we can definitely update op alloy to have the checks there instead of here. We could even add them to the `try_into_block` method there

PR #1459, opened 2025-04-17
fix(protocol): Incorrect Genesis Hash Consensus Block

refcell:
### Description

Converting the alloy rpc types eth Block into a consensus block, and then hashing the resulting header produces a mismatch for the genesis hash.

Previously we tried to de-duplicate the rpc and consensus block conversions to the `L2BlockInfo` type but it's produced a bug, so we'll live with this small duplication for now.

### DIFF
--- a/crates/protocol/protocol/src/block.rs
+++ b/crates/protocol/protocol/src/block.rs
@@ -144,7 +144,7 @@ impl L2BlockInfo {
     /// Constructs an [`L2BlockInfo`] from a given [`alloy_rpc_types_eth::Block`] and
     /// [`ChainGenesis`].
     pub fn from_rpc_block_and_genesis(
-        block: &alloy_rpc_types_eth::Block<op_alloy_rpc_types::Transaction>,
+        block: alloy_rpc_types_eth::Block<op_alloy_rpc_types::Transaction>,
         genesis: &ChainGenesis,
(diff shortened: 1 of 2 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1459 👈
* `main`

crates/protocol/protocol/src/block.rs
```diff
@@ -141,6 +141,27 @@ impl L2BlockInfo {
         Self { block_info, l1_origin, seq_num }
     }
 
+    /// Constructs an [`L2BlockInfo`] from a given [`alloy_rpc_types_eth::Block`] and
+    /// [`ChainGenesis`].
+    pub fn from_rpc_block_and_genesis(
+        block: &alloy_rpc_types_eth::Block<op_alloy_rpc_types::Transaction>,
```
theochap, line 147: nit: would it be possible to avoid the clone below by passing an owned `Block` instead of a reference here? It seems that we have ownership of the block in the calling function and we don't reuse it after. 
It sounds like cloning a whole block would be expensive so maybe we could try and avoid that!

theochap (approved):
Looks good, one nit

crates/protocol/protocol/src/block.rs
refcell: Good nit thx

PR #1465, opened 2025-04-17
feat(docker): Use generic dockerfile for all binary apps

clabby:
## Overview

Adds in a generic dockerfile for all binary applications from this repository. This dockerfile replaces the old `kona_host` dockerfile, and makes some improvements on it such as setting the default entrypoint to the binary as well as placing it into the `PATH` by default.

Also includes some docs on how to include new binary app targets in the `docker-bake.hcl` file.


docker/apps/README.md
```diff
@@ -0,0 +1,23 @@
...
+  inherits = ["docker-metadata-action"]
+  context = "."
+  dockerfile = "docker/apps/kona_app_generic.dockerfile"
+  args = {
+    TAG = "${GIT_REF_NAME}"
+    BIN_TARGET = "<bin-name>"
+  }
+  platforms = split(",", PLATFORMS)
+}
+```
```
refcell, line 20: Do we actually need to do this? Or can we make the target "kona-generic-apps" and then use an input arg in the tag as the `BIN_TARGET`? Then we don't need to change any code at all for new apps, we can just pass the bin name in as a variable in the tag.

refcell (commented):
One comment/suggestion

theochap (approved):

docker/apps/README.md
refcell: Will be a future enhancement

refcell (approved):

PR #1467, opened 2025-04-18
feat: function to init tracing in testing

### LINKED ISSUE #1437: Test utils tracing
Make function for init tracing in testing, under the hood calling
https://github.com/op-rs/kona/blob/d8dbf5156edc0acf9646c1b002eda62be7228bae/crates/utilities/cli/src/tracing.rs#L6-L32
and passing `EnvFilter::from_default_env` as param to `env_filter`

This eradicates the need for print statements which we don't want in production code.

See how it's done in reth tests
https://github.com/paradigmxyz/reth/blob/c79b63876e80a633818be882e7d3625785353c38/crates/tracing/src/lib.rs#L220-L229

PoulavBhowmick03:
Fixes #1437

### DIFF
--- a/crates/utilities/cli/src/lib.rs
+++ b/crates/utilities/cli/src/lib.rs
@@ -11,8 +11,7 @@ pub use clap::cli_styles;
 pub mod backtrace;
 
 mod tracing;
-pub use tracing::init_tracing_subscriber;
-
+pub use tracing::{init_test_tracing, init_tracing_subscriber};
 mod prometheus;
 pub use prometheus::init_prometheus_server;
 
--- a/crates/utilities/cli/src/lib.rs
+++ b/crates/utilities/cli/src/lib.rs
@@ -12,6 +12,7 @@ pub mod backtrace;
 
 mod tracing;
 pub use tracing::{init_test_tracing, init_tracing_subscriber};
+
 mod prometheus;
 pub use prometheus::init_prometheus_server;
 
--- a/crates/utilities/cli/src/tracing.rs
+++ b/crates/utilities/cli/src/tracing.rs
@@ -25,3 +25,13 @@ pub fn init_tracing_subscriber(
     let subscriber = tracing_subscriber::fmt().with_max_level(level);
     tracing::subscriber::set_global_default(subscriber.with_env_filter(filter).finish())
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn init_test_tracing(){
+        let _ = init_tracing_subscriber(2, Some(EnvFilter::from_default_env()));
+    }
+}
--- a/crates/utilities/cli/src/tracing.rs
+++ b/crates/utilities/cli/src/tracing.rs
@@ -31,7 +31,7 @@ mod tests {
     use super::*;
 
     #[test]
-    fn init_test_tracing(){
+    fn init_test_tracing() {
         let _ = init_tracing_subscriber(2, Some(EnvFilter::from_default_env()));
     }
 }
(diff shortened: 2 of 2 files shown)


crates/utilities/cli/src/tracing.rs
```diff
@@ -25,3 +25,13 @@ pub fn init_tracing_subscriber(
...
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn init_test_tracing() {
+        let _ = init_tracing_subscriber(2, Some(EnvFilter::from_default_env()));
+    }
+}
```
refcell, line 37: I _think_ what we actually want here is a method that can be called from tests to init tracing, not to invoke the `init_tracing_subscriber` call from in a test.

For example, in reth the `init_test_tracing` method is [called at the top of a test](https://github.com/paradigmxyz/reth/blob/main/crates/net/downloaders/src/receipt_file_client.rs#L543-L545) to enable tracing.

We'll want to provide a similar function that can be called. So we'll need something like:

```suggestion
#[cfg(test)]
pub fn init_test_tracing() {
        let _ = init_tracing_subscriber(2, None);
}
```

Sidenode: we also don't need the env filter since this is by default setup in the `init_tracing_subscriber` method.

refcell (changes requested):

crates/utilities/cli/src/tracing.rs
PoulavBhowmick03: seems to give me this error
```sh
consider specifying the generic argument: `::<T>`rustc[E0283](https://doc.rust-lang.org/error-index.html#E0283)
tracing.rs(31, 44): original diagnostic
type annotations needed
cannot satisfy `_: Into<EnvFilter>`rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[0]?0#file:///home/odinson/OpenSource/kona/crates/utilities/cli/src/tracing.rs)
tracing.rs(31, 17): required by a bound introduced by this call
tracing.rs(16, 29): required by a bound in `init_tracing_subscriber`
tracing.rs(31, 48): consider specifying the generic argument: `::<T>`
core::option::Option
None
No value.
```

and the pub fn is unreachable currently, after adding EnvFilter as an argument. What do you suggest?

PoulavBhowmick03:
```sh
unreachable `pub` item
or consider exporting it for use by other crates
requested on the command line with `-W unreachable-pub`rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[1]?1#file:///home/odinson/OpenSource/kona/crates/utilities/cli/src/tracing.rs)
tracing.rs(30, 1): consider restricting its visibility: `pub(crate)`
function `init_test_tracing` is never used
`#[warn(dead_code)]` on by defaultrustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20[0]?0#file:///home/odinson/OpenSource/kona/crates/utilities/cli/src/tracing.rs)
```

PoulavBhowmick03:
@refcell can you ptal?

crates/utilities/cli/src/lib.rs
```diff
@@ -11,8 +11,7 @@ pub use clap::cli_styles;
 pub mod backtrace;
 
 mod tracing;
-pub use tracing::init_tracing_subscriber;
-
+pub use tracing::{init_test_tracing, init_tracing_subscriber};
```
refcell, line 14:
```suggestion
pub use tracing::{init_test_tracing, init_tracing_subscriber};

```

refcell: Add back the removed line pls

refcell (approved):

PR #1470, opened 2025-04-18
feat(ci): Generic Binary

### LINKED ISSUE #1466: feat(docker): Generic Binary Targets
### Description

> [!NOTE]
>
> Follows: https://github.com/op-rs/kona/pull/1465

Kona lets you dockerize applications via a tag. This is supported in the [`docker.yaml`](https://github.com/op-rs/kona/blob/main/.github/workflows/docker.yaml#L15-L18) github action workflow. Currently, only tags in the format `*/v*` are supported. That is, in the representation of `service/version`. For example, the `docker.yaml` workflow will dockerize and publish a new docker image for the `kona-host` binary when a tag like `kona-host/v1.0.0` is cut and pushed to the `kona` repository.

Previously, adding a docker support for a binary in kona required a new target in the [`docker-bake.hcl`](https://github.com/op-rs/kona/blob/main/docker/docker-bake.hcl) as well as adding a new `<binary-name>.dockerfile` file. https://github.com/op-rs/kona/pull/1465 updated kona to make the dockerfile generic. So the only step required to support dockerizing a new binary in kona, is to add a target to the [`docker-bake.hcl`](https://github.com/op-rs/kona/blob/main/docker/docker-bake.hcl).

But we can make it even simpler.

As part of this ticket, make the `kona-host` [target in `docker-bake.hcl`](https://github.com/op-rs/kona/blob/main/docker/docker-bake.hcl#L50-L58) generic so the target is called "generic". Then, add a new github action similar to [`docker.yaml`](https://github.com/op-rs/kona/blob/main/.github/workflows/docker.yaml#L15-L18) that parses tags in the format of `*/*/v*`. For example, a `kona-host` tag would look like: `generic/kona-host/v1.1.0`.

Update the docs in the [app](https://github.com/op-rs/kona/tree/main/docker/apps) directory and the [docker README](https://github.com/op-rs/kona/blob/main/docker/README.md) to document cutting releases for generic apps using the new format.

refcell:
### Description

Use a generic binary target in the docker bake.

Closes #1466

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -69,11 +69,7 @@ jobs:
       - name: Create matrix
         id: platforms
         run: |
-          if [[ "${{ steps.target-spec.outputs.bake_target }}" == "generic" ]]; then
-            echo 'matrix=["linux/amd64", "linux/arm64"]' >> $GITHUB_OUTPUT
-          else
-            echo "matrix=$(docker buildx bake -f docker/docker-bake.hcl ${{ steps.target-spec.outputs.bake_target }} --print | jq -cr '.target."${{ steps.target-spec.outputs.bake_target }}".platforms')" >> ${GITHUB_OUTPUT}
-          fi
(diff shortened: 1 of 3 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1470 👈
* `main`

refcell:
Runs successfully: https://github.com/op-rs/kona/actions/runs/14541415982

theochap (approved):

PR #1473, opened 2025-04-19
fix(proof): Blob preimage keys

clabby:
## Overview

Implements a fix for a bug recently found in `op-program` by @imtei that `kona-proof` shares. Our old blob preimage key seeded the field element index, rather than the corresponding bit-reversed root of unity.

The fix for this issue is to alter the blob field element preimage key construction to consist of `kzg_commitment ++ brp_root_of_unity[i]`, rather than `kzg_commitment ++ i`.

For users such as `op-succinct` + `kailua`, which use the full KZG proof to verify the entire blob at once, the preimage key @ `i = 4096` containing the full KZG proof is retained under the legacy key format. No action should be needed from these users. This issue should only affect users that run `kona-proof` in a fault proof environment, of which there currently are none to my knowledge.

A more in-depth overview of this issue can be found [in the `op-program` fix PR within the OP monorepo](https://github.com/ethereum-optimism/optimism/pull/15354), as well as a disclosure of the bug [here](https://gov.optimism.io/t/vulnerability-disclosure-incorrect-blob-preimages/9833).

### DIFF
--- a/crates/proof/proof/src/l1/blob_provider.rs
+++ b/crates/proof/proof/src/l1/blob_provider.rs
@@ -5,10 +5,14 @@ use alloc::{boxed::Box, sync::Arc, vec::Vec};
 use alloy_consensus::Blob;
 use alloy_eips::eip4844::{FIELD_ELEMENTS_PER_BLOB, IndexedBlobHash};
 use alloy_primitives::keccak256;
+use ark_bls12_381::Fr;
+use ark_ff::{AdditiveGroup, BigInteger, BigInteger256, Field, PrimeField};
 use async_trait::async_trait;
+use core::str::FromStr;
 use kona_derive::traits::BlobProvider;
 use kona_preimage::{CommsClient, PreimageKey, PreimageKeyType};
 use kona_protocol::BlockInfo;
+use spin::Lazy;
 
 /// An oracle-backed blob provider.
 #[derive(Debug, Clone)]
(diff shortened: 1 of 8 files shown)


crates/proof/proof/src/l1/blob_provider.rs
```diff
@@ -91,3 +96,113 @@ impl<T: CommsClient + Sync + Send> BlobProvider for OracleBlobProvider<T> {
...
+#[cfg(test)]
+mod test {
+    use super::ROOTS_OF_UNITY;
+    use alloy_eips::eip4844::{FIELD_ELEMENTS_PER_BLOB, env_settings::EnvKzgSettings};
+    use ark_ff::{BigInteger, PrimeField};
+    use c_kzg::{BYTES_PER_BLOB, Blob, Bytes32, Bytes48};
+    use rand::Rng;
+
+    #[test]
+    fn test_slow_roots_of_unity() {
```
clabby, line 162: Note to reader - this test takes a long time (~1m30s) to run. I've excluded it from the default `test` directive, as the functionality here is not subject to frequent change.

Inphi: is there a faster smoke test we could add alongside this one?

clabby: I could maybe pull in KZG-rs, which is parallel to do this. Didn't want to add a dep that heavy just for testing, though.

Perhaps could use a fully 0'd out blob? That would likely be quite a bit faster.

clabby: er actually, I'll just try to parallelize this rq. 1 sec.

clabby: Parallelizing it brought it down to 9-10s on my machine. Probably good enough to include in CI.

Inphi: Nice. i was thinking of checking a single field element as a fast, but non-comprehensive test. Rayon power is even better.

Inphi (approved):

PR #1476, opened 2025-04-21
chore(workspace): Improve proof tracing

clabby:
## Overview

Makes some tweaks to the common traces in the proof program. Favors native k/v trace attributes and moves some `snake_case` targets to `kebab-case`.

### DIFF
--- a/bin/client/src/single.rs
+++ b/bin/client/src/single.rs
@@ -78,9 +78,9 @@ where
     if boot.claimed_l2_block_number < safe_head.number {
         error!(
             target: "client",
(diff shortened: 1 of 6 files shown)


refcell (approved):
I'm not a fan of the kebab casing because it's harder to select. With snake casing you can select the whole target as a word

clabby:
> I'm not a fan of the kebab casing because it's harder to select. With snake casing you can select the whole target as a word

I don't mind either way. I'll follow up to standardize all of the kebab case targets to snake case, just don't like having both.

PR #1477, opened 2025-04-21
feat(node): Redial Peers

### LINKED ISSUE #1468: feat(bin/node): Tune Peer Redialing
### Description

Kona has a p2p networking stack comprised of two key components.
- `discv5` discovery to discover peers
- `libp2p` Swarm to receiving gossip

When peers disconnect, kona currently has a strategy to redial peers aggressively in order to keep a robust mesh of peers. Redialing is performed in kona's `libp2p` "gossip" service [here](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/gossip/driver.rs#L187-L194).

Instead of hardcoding the logic to always redial peers, kona should extend the option to redial as a CLI arg. To make it configurable, provide a cli arg that accepts a number. That is, the number of times to redial a peer before abandoning the peer (should probably evict it from the bootstore as well but that's another topic). Kona's gossip service can just keep a map of peer addresses to the number of times it has attempted to redial a peer. This way, once the configured threshold is reached, kona will never redial the peer.

Add this flag to the [p2p cli args](https://github.com/op-rs/kona/blob/main/bin/node/src/flags/p2p.rs#L20-L22). It could be `--p2p.redial`: the threshold number of times to redial a peer before abandoning the peer.

The cli flag should be `Option<u32>` so when `None`, the gossip service will always redial. By default, it'll probably make sense to set the limit at around `3` redials or so. A future ticket could address a backoff strategy to backoff redialing.

refcell:
### Description

Redial peers with a configurable CLI argument.

Closes #1468

### DIFF
--- a/bin/node/src/flags/p2p.rs
+++ b/bin/node/src/flags/p2p.rs
@@ -125,6 +125,11 @@ pub struct P2PArgs {
     /// The directory to store the bootstore.
     #[arg(long = "p2p.bootstore", env = "KONA_NODE_P2P_BOOTSTORE")]
     pub bootstore: Option<PathBuf>,
+    /// Peer Redialing threshold is the maximum amount of times to attempt to redial a peer that
+    /// disconnects. By default, peers are *not* redialed. If set to 0, the peer will be
+    /// redialed indefinitely.
+    #[arg(long = "p2p.redial", env = "KONA_NODE_P2P_REDIAL")]
+    pub peer_redial: Option<u64>,
 }
 
 impl Default for P2PArgs {
@@ -151,6 +156,7 @@ impl Default for P2PArgs {
             ban_duration: 30,
             discovery_interval: 5,
             bootstore: None,
+            peer_redial: None,
         }
     }
 }
@@ -240,6 +246,7 @@ impl P2PArgs {
             block_time,
             monitor_peers,
             bootstore: self.bootstore.clone(),
+            redial: self.peer_redial,
         })
     }
 
--- a/crates/node/p2p/src/gossip/builder.rs
+++ b/crates/node/p2p/src/gossip/builder.rs
@@ -35,6 +35,10 @@ pub struct GossipDriverBuilder {
     /// If set, the gossip layer will monitor peer scores and ban peers that are below a given
     /// threshold.
     peer_monitoring: Option<PeerMonitoring>,
+    /// The number of times to redial a peer.
+    /// If unset, peers will not be redialed.
+    /// If set to `0`, peers will be redialed indefinitely.
+    peer_redial: Option<u64>,
 }
 
 impl GossipDriverBuilder {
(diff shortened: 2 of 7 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1477 👈
* `main`

bin/node/src/flags/p2p.rs
```diff
@@ -125,6 +125,10 @@ pub struct P2PArgs {
     /// The directory to store the bootstore.
     #[arg(long = "p2p.bootstore", env = "KONA_NODE_P2P_BOOTSTORE")]
     pub bootstore: Option<PathBuf>,
+    /// Peer Redialing threshold is the maximum amount of times to attempt to redial a peer that
+    /// disconnects. By default, peers are *not* redialed.
+    #[arg(long = "p2p.redial", env = "KONA_NODE_P2P_REDIAL")]
```
theochap, line 130: nit: Maybe we can mention that if the flag is set to 0 peers are redialed infinitely

crates/node/p2p/src/gossip/driver.rs
```diff
@@ -30,12 +30,14 @@ pub struct GossipDriver {
     /// The [`BlockHandler`].
     pub handler: BlockHandler,
     /// A list of peers we have successfully dialed.
-    pub dialed_peers: HashMap<Multiaddr, bool>,
+    pub dialed_peers: HashMap<Multiaddr, u64>,
```
theochap, line 33: What is the `u64` for here? Maybe we could add a small comment to explain

crates/node/p2p/src/gossip/driver.rs
```diff
@@ -105,9 +113,24 @@ impl GossipDriver {
...
+        let Some(dialed) = self.dialed_peers.get(addr) else {
+            return false;
+        };
+        // If the peer has been dialed and the threshold is not set, the threshold is reached.
+        let Some(redialing) = self.peer_redialing else {
+            return true;
+        };
+        // If the threshold is set to `0`, redial indefinitely.
+        if redialing == 0 {
+            return true;
```
theochap, line 128: Should we return `true` (meaning the dial threshold is reached) if we redial indefinitely?

crates/node/p2p/src/gossip/builder.rs
```diff
@@ -35,6 +35,10 @@ pub struct GossipDriverBuilder {
     /// If set, the gossip layer will monitor peer scores and ban peers that are below a given
     /// threshold.
     peer_monitoring: Option<PeerMonitoring>,
+    /// The number of times to redial a peer.
+    /// If unset, peers will not be redialed.
+    /// If set to `0`, peers will be redialed indefinitely.
+    peer_redial: Option<u64>,
```
emhane, line 41: this is typically referred to as "retries" within network engineering

theochap (approved):
Looks great to me!

PR #1478, opened 2025-04-21
feat(node): Bootstore Debugging Tool

refcell:
### Description

Upstreams a simple bootstore debugging cli tool that I've been using locally.

Prints bootstore file info with `kona-node boostore --all` (for all available bootstores).

Info for a single bootstore can be output by specifying the chain id like so.

```
kona-node --l2-chain-id 11155420 bootstore
```

### DIFF
--- a/bin/node/src/cli.rs
+++ b/bin/node/src/cli.rs
@@ -1,7 +1,7 @@
 //! Contains the node CLI.
 
 use crate::{
-    commands::{NetCommand, NodeCommand, RegistryCommand},
+    commands::{BootstoreCommand, NetCommand, NodeCommand, RegistryCommand},
     flags::{GlobalArgs, MetricsArgs},
 };
 use anyhow::Result;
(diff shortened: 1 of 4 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1478 👈
* `main`

bin/node/src/cli.rs
```diff
@@ -18,6 +18,8 @@ pub enum Commands {
     Net(NetCommand),
     /// Lists the OP Stack chains available in the superchain-registry.
     Registry(RegistryCommand),
+    /// Utility tool to interact with local bootstores.
+    Bootstore(BootstoreCommand),
```
theochap, line 22: nit: Would it be better to feature-gate this command to avoid adding test-code to the production binary? That sounds like a useful feature so I don't want to block on it, but maybe we could think of a `test-utils` feature that we could activate for this kind of tooling. Not super partial though so feel free to disregard

theochap (approved):
One nit. Can be merged and the nit be tracked with an issue as well

bin/node/src/cli.rs
refcell: I think it's less of a testing tool rather than a utility tool. I think it'll be useful for folks to quickly check where bootstores are located, and if they're full of peers

PR #1482, opened 2025-04-21
fix(bin/node): Global Argument Positioning

refcell:
### Description

Previously, global and metrics arguments were not global. Meaning, when running `kona-node`, the global and metrics args had to be specified _prior_ to the subcommand.

For example, when running the `node` subcommand the global `--l2-chain-id` argument had to be specified prior to the `node` subcommand.

This made for a weird UX as follows.

```sh
$ kona-node --l2-chain-id 11155420 node ...
```

Now, the global args are marked as `global=true` so Clap is able to parse them both before *and* _after_ the subcommand like so.

```sh
$ kona-node node --l2-chain-id 11155420 ...
```

### DIFF
--- a/bin/node/src/flags/globals.rs
+++ b/bin/node/src/flags/globals.rs
@@ -11,10 +11,10 @@ use tracing_subscriber::EnvFilter;
 #[derive(Parser, Default, Clone, Debug)]
 pub struct GlobalArgs {
     /// Verbosity level (0-2)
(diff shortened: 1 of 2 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1482 👈
* `main`

theochap (approved):
Amazing!

PR #1483, opened 2025-04-21
chore(bin/node): Subcommand Aliases

refcell:
### Description

Adds a few intuitive aliases for various `kona-node` subcommands.

### DIFF
--- a/bin/node/src/cli.rs
+++ b/bin/node/src/cli.rs
@@ -13,10 +13,13 @@ use kona_cli::cli_styles;
 #[allow(clippy::large_enum_variant)]
 pub enum Commands {
     /// Runs the consensus node.
+    #[command(alias="n")]
     Node(NodeCommand),
     /// Runs the networking stack for the node.
+    #[command(alias="p2p", alias="network")]
     Net(NetCommand),
     /// Lists the OP Stack chains available in the superchain-registry.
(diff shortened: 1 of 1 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1483 👈
* `main`

bin/node/src/cli.rs
```diff
@@ -13,10 +13,13 @@ use kona_cli::cli_styles;
 #[allow(clippy::large_enum_variant)]
 pub enum Commands {
     /// Runs the consensus node.
+    #[command(alias="n")]
     Node(NodeCommand),
     /// Runs the networking stack for the node.
+    #[command(alias="p2p", alias="network")]
     Net(NetCommand),
     /// Lists the OP Stack chains available in the superchain-registry.
+    #[command(alias="r", alias="scr")]
```
theochap, line 22: nit: What does `scr` stand for? Maybe just "r" and "reg" would be enough?

theochap (approved):

bin/node/src/cli.rs
refcell: Meant to be shorthand for "superchain-registry" which a lot of folks refer to as the "scr". Let's remove it down the road if it's annoying

PR #1486, opened 2025-04-21
bug(bin/node): fix metrics address flag

theochap:
## Description

This metrics fixes the name of the metrics address flag to make it compliant with [op-node](https://github.com/ethereum-optimism/optimism/blob/836d50be5d5f4ae14ffb2ea6106720a2b080cdae/op-node/flags/flags.go#L287)

### DIFF
--- a/bin/node/src/flags/metrics.rs
+++ b/bin/node/src/flags/metrics.rs
@@ -18,7 +18,7 @@ pub struct MetricsArgs {
     #[arg(long = "metrics.port", default_value = "9090", env = "KONA_NODE_METRICS_PORT")]
     pub port: u16,
     /// The ip address to use to emit prometheus metrics.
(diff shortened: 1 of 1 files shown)


refcell (approved):

bin/node/src/flags/metrics.rs
```diff
@@ -18,7 +18,7 @@ pub struct MetricsArgs {
     #[arg(long = "metrics.port", default_value = "9090", env = "KONA_NODE_METRICS_PORT")]
     pub port: u16,
     /// The ip address to use to emit prometheus metrics.
-    #[arg(long = "metrics.address", default_value = "0.0.0.0", env = "KONA_NODE_METRICS_ADDR")]
+    #[arg(long = "metrics.addr", default_value = "0.0.0.0", env = "KONA_NODE_METRICS_ADDR")]
```
refcell, line 21: Looks like this change breaks the `test_metrics_args_listen_ip` test below. After fixing that up, should be good to merge.

PR #1489, opened 2025-04-22
chore(node/p2p): Broadcast Wrapper

refcell:
### Description

Wraps broadcasting to encapsulate the unsafe block buffer.

Improves `kona-p2p` so the unsafe block buffer is better abstracted.

### DIFF
--- a/crates/node/p2p/src/net/broadcast.rs
+++ b/crates/node/p2p/src/net/broadcast.rs
@@ -1,0 +1,50 @@
+//! Broadcast handles broadcasting unsafe blocks in-order.
+
+use op_alloy_rpc_types_engine::OpNetworkPayloadEnvelope;
+use std::collections::VecDeque;
+use tokio::sync::broadcast::{Receiver as BroadcastReceiver, Sender as BroadcastSender};
+
+/// The `Broadcast` struct is responsible for broadcasting unsafe blocks in order.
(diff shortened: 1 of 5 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1489 👈
* `main`

crates/node/p2p/src/net/broadcast.rs
```diff
@@ -0,0 +1,50 @@
...
+                break;
+            }
+            if let Some(block) = self.buffer.front() {
+                if let Err(e) = self.channel.send(block.clone()) {
+                    trace!("Failed to send unsafe block through channel: {:?}", e);
+                    break;
+                }
+                self.buffer.pop_front();
+            }
+        }
```
emhane, line 48: how about having a set number of blocks to flush per call to this function? the number can be configurable on start up of the node. having this loop infinitely over and unbounded queue is a dos vector.

emhane (changes requested):

crates/node/p2p/src/net/broadcast.rs
refcell: Good idea. Write up a ticket and it's a good follow-up for this PR.

clabby (approved):

PR #1493, opened 2025-04-23
fix(node/p2p): Peer Score Level Off

refcell:
### Description

Currently, setting the `PeerScoreLevel` to `None` isn't possible via clap since it seems to be interpreted as an option variant.

In other words, if you pass `--p2p.scoring none` to the `kona-node` it will error on parsing `none` as a variant of `PeerScoreLevel`.

This PR changes the `r#None` variant to `Off` so peer scoring can be disabled with `--p2p.scoring off`.

> [!NOTE]
>
> Notice, this creates a diff in the api of the `--p2p.scoring` cli flag between the `kona-node` and `op-node`.


refcell:
## 📚 $\text{Stack Overview}$

* #1493 👈
* `main`

emhane (approved):

clabby (approved):

PR #1495, opened 2025-04-23
chore(node/p2p): Forward Discovery Events

refcell:
### Description

Fixes the `kona-node` discovery driver to properly forward valid ENRs discovered through `discv5` to the gossip driver

### DIFF
--- a/crates/node/p2p/src/discv5/driver.rs
+++ b/crates/node/p2p/src/discv5/driver.rs
@@ -386,6 +386,20 @@ impl Discv5Driver {
                                     });
                                 }
                             }
+                            discv5::Event::UnverifiableEnr { enr, .. } => {
+                                debug!(target: "discovery", "Unverifiable ENR discovered: {:?}", enr);
+                                if EnrValidation::validate(&enr, chain_id).is_valid() {
+                                    debug!(target: "discovery", "Valid ENR discovered, forwarding to swarm: {:?}", enr);
+                                    self.store.add_enr(enr.clone());
+                                    let sender = enr_sender.clone();
+                                    tokio::spawn(async move {
+                                        if let Err(e) = sender.send(enr).await {
+                                            debug!(target: "discovery", "Failed to send enr: {:?}", e);
+                                        }
+                                    });
+                                }
+
+                            }
                             _ => {}
                         }
                     }
--- a/crates/node/p2p/src/discv5/driver.rs
+++ b/crates/node/p2p/src/discv5/driver.rs
@@ -279,6 +279,28 @@ impl Discv5Driver {
             // Step 3: Forward ENRs in the bootstore to the enr receiver.
             self.forward(enr_sender.clone()).await;
 
+            // Continuously attempt to start the event stream with a retry limit and shutdown
+            // signal.
+            let mut retries = 0;
+            let max_retries = 10; // Maximum number of retries before giving up.
+            let mut event_stream = loop {
+                if retries >= max_retries {
+                    error!(target: "discovery", "Exceeded maximum retries for event stream startup. Aborting...");
+                    return; // Exit the task if the retry limit is reached.
+                }
+                match self.disc.event_stream().await {
+                    Ok(event_stream) => {
+                        break event_stream;
+                    }
+                    Err(e) => {
+                        warn!(target: "discovery", "Failed to start event stream: {:?}", e);
+                        retries += 1;
+                        sleep(Duration::from_secs(2)).await;
+                        info!(target: "discovery", "Retrying event stream startup... (Attempt {}/{})", retries, max_retries);
+                    }
+                }
+            };
+
             // Step 4: Run the core driver loop.
             loop {
                 tokio::select! {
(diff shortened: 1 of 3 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1495 👈
* `main`

crates/node/p2p/src/discv5/driver.rs
```diff
@@ -314,6 +328,39 @@ impl Discv5Driver {
...
+                                    tokio::spawn(async move {
+                                        if let Err(e) = sender.send(enr).await {
+                                            debug!(target: "discovery", "Failed to send enr: {:?}", e);
+                                        }
+                                    });
+                                }
+                            }
+                            discv5::Event::SessionEstablished(enr, addr) => {
+                                if EnrValidation::validate(&enr, chain_id).is_valid() {
+                                    debug!(target: "discovery", "Session established with valid ENR, forwarding to swarm. Address: {:?}, ENR: {:?}", addr, enr);
```
emhane, line 351:
```suggestion
                                    debug!(target: "discovery",
                                        %addr,
                                        ?enr,
                                        "Session established with valid ENR, forwarding to swarm"
                                    );
```
let's use the logs the way they were designed to be used with kv pairs. the above log will output
`.. addr=<the socket addr value> enr=<the enr in human readable format> ..`
NOTE! use the debug operator for enrs, ie `?enr`, otherwise they serialise in base 64 which is super unhelpful for debugging

emhane (commented):
check out reth's discv5 workaround for geth enrs. sigp/discv5 is much stricter about correct enr configuration for its peers than geth, so we had to ask sigp to add this special new event for being geth compatible when using sigp/discv5 for EL
https://github.com/paradigmxyz/reth/blob/dd4aa1a85b5babbcedf2a5978cd88b631bc486e8/crates/net/discv5/src/lib.rs#L255-L283

crates/node/p2p/src/discv5/driver.rs
```diff
@@ -279,6 +279,28 @@ impl Discv5Driver {
...
+                        break event_stream;
+                    }
+                    Err(e) => {
+                        warn!(target: "discovery", "Failed to start event stream: {:?}", e);
+                        retries += 1;
+                        sleep(Duration::from_secs(2)).await;
+                        info!(target: "discovery", "Retrying event stream startup... (Attempt {}/{})", retries, max_retries);
+                    }
+                }
+            };
```
theochap, line 302: question: did we have issues starting the discovery event stream? That's interesting, never noticed it on my end

crates/node/p2p/src/discv5/driver.rs
```diff
@@ -334,6 +356,53 @@ impl Discv5Driver {
...
+                                        if let Err(e) = sender.send(enr).await {
+                                            debug!(target: "discovery", "Failed to send enr: {:?}", e);
+                                        }
+                                    });
+                                }
+
+                            }
+                            _ => {}
+                        }
+                    }
```
theochap, line 405: Just checking: do we unit test the `EnrValidation::validate` method? If not, maybe we should add a couple of unit tests just to be on the safe side. Can be done in a separate PR

crates/node/p2p/src/discv5/driver.rs
```diff
@@ -334,6 +356,53 @@ impl Discv5Driver {
...
+                    event = event_stream.recv() => {
+                        let Some(event) = event else {
+                            trace!(target: "discovery", "Received `None` event");
+                            continue;
+                        };
+                        match event {
+                            discv5::Event::Discovered(enr) => {
+                                if EnrValidation::validate(&enr, chain_id).is_valid() {
+                                    debug!(target: "discovery", "Valid ENR discovered, forwarding to swarm: {:?}", enr);
+                                    self.store.add_enr(enr.clone());
```
theochap, line 368: Just checking: below ([here](https://github.com/op-rs/kona/pull/1495/files#diff-e3c2d30d9bcd63009fc8fd07a05b5a190843f0c2b78f8c58f5d63493288c64ecR426)), don't we already add the ENRs to the store? Maybe we should try to remove this merge below so that we only have one place where we mutate the store (that is provided we are confident that all the enr update events are handled here)

refcell: If it is successfully added, then this becomes a no-op since the bootstore will check if it contains the peer before adding it again inside [`add_enr`](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/peers/store.rs#L52-L54). So we don't have to worry about calling `add_enr` with redundant ENRs by design. Better to call again in case the discovery service discovers an ENR outside of the find node query.

crates/node/p2p/src/discv5/driver.rs
refcell: Good callout. I have an open pr [here](https://github.com/op-rs/kona/pull/1552) that could make sense to add those tests.

theochap (approved):

theochap:
Looks great thanks for taking care of it!

PR #1496, opened 2025-04-23
feat(bin/node): add custom bootnode list CLI arg

theochap:
## Description
This PR adds a new CLI flag to allow manually defining a custom list of initial bootnodes to be added to the p2p discovery layer at startup. This change will allow us to add local network addresses inside Kurtosis.

### DIFF
--- a/bin/node/src/flags/p2p.rs
+++ b/bin/node/src/flags/p2p.rs
@@ -8,6 +8,7 @@ use crate::flags::GlobalArgs;
 use alloy_primitives::B256;
 use anyhow::Result;
 use clap::Parser;
+use discv5::Enr;
 use kona_p2p::{Config, PeerMonitoring, PeerScoreLevel};
 use libp2p::identity::Keypair;
 use std::{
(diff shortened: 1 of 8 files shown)


bin/node/src/flags/p2p.rs
```diff
@@ -247,6 +258,10 @@ impl P2PArgs {
             monitor_peers,
             bootstore: self.bootstore.clone(),
             redial: self.peer_redial,
+            // It is ok to clone here since the config only happens at startup
+            // and that we assume the number of bootnodes explicitly specified
+            // through the CLI is small.
+            bootnodes: self.bootnodes.clone(),
```
refcell, line 264: I think we should actually change this method to be `into_config`, consuming the `P2PArgs`.

Can split it out into another PR though, not super important here since we're already cloning the bootstore.

crates/node/p2p/src/peers/store.rs
```diff
@@ -21,7 +21,7 @@ pub struct BootStore {
     pub peers: Vec<Enr>,
 }
 
-// This custom implementation of `Deserialize` allows us to avignore
```
refcell, line 24: `avignore` lol, no idea what I was trying to say there

refcell (approved):

bin/node/src/flags/p2p.rs
theochap: Sounds good, will open a separate PR to do it!

PR #1497, opened 2025-04-23
chore(bin/node): adding more flexibility to verbosity levels

theochap:
## Description
This PR expands the number of possible verbosity level configurations for kona. This matches reth's configuration options and can be helpful to debug the node's state in Kurtosis with finer grained precision.


refcell (approved):
Nice, thanks!

PR #1500, opened 2025-04-23
fix(bin/node): Discover Subcommand

refcell:
### Description

Fixes up the discover subcommand.

### DIFF
--- a/bin/node/src/commands/discover.rs
+++ b/bin/node/src/commands/discover.rs
@@ -142,7 +142,7 @@ impl Discovery {
                     }
                 }
                 _ = disc_ticker.tick() => {
-                    if let Ok(c) = handler.peer_count().try_recv() {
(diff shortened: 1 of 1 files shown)


bin/node/src/commands/discover.rs
```diff
@@ -142,7 +142,7 @@ impl Discovery {
                     }
                 }
                 _ = disc_ticker.tick() => {
-                    if let Ok(c) = handler.peer_count().try_recv() {
+                    if let Ok(c) = handler.peer_count().await {
```
refcell, line 145: TODO: pull this into a spawned task that attempts to grab the pc and sends it over a channel here to await instead of using the `disc_ticker` and then blocking.

refcell:
[Ticketed](https://github.com/op-rs/kona/issues/1539). This is not that important so throwing onto the backlog.

PR #1503, opened 2025-04-24
fix(proof-interop): Operate provider off of `local-safe` heads

clabby:
## Overview

Fixes an issue in the interop client program where the `OracleInteropProvider` was operating off of the `cross-safe` heads, rather than the `local-safe` heads. After this change, the global interop provider is instantiated with the `local-safe` headers that were generated during the runs of the first sub-problem.

### DIFF
--- a/bin/client/src/interop/consolidate.rs
+++ b/bin/client/src/interop/consolidate.rs
@@ -2,8 +2,9 @@
 
 use super::FaultProofProgramError;
 use crate::interop::util::fetch_output_block_hash;
-use alloc::{sync::Arc, vec::Vec};
+use alloc::sync::Arc;
 use core::fmt::Debug;
+use kona_executor::TrieDBProvider;
 use kona_preimage::{HintWriterClient, PreimageOracleClient};
 use kona_proof::{CachingOracle, l2::OracleL2ChainProvider};
 use kona_proof_interop::{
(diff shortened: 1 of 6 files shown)


bin/client/src/interop/consolidate.rs
```diff
@@ -27,65 +28,79 @@ where
...
+        // Fetch the rollup config for the given L2 chain ID.
+        let rollup_config = ROLLUP_CONFIGS
+            .get(&cross_safe_output.chain_id)
+            .or_else(|| boot.rollup_configs.get(&cross_safe_output.chain_id))
+            .ok_or(FaultProofProgramError::MissingRollupConfig(cross_safe_output.chain_id))?;
+
+        // Initialize the local provider for the current L2 chain.
+        let mut local_provider = OracleL2ChainProvider::new(
+            cross_safe_head_hash,
+            Arc::new(rollup_config.clone()),
```
refcell, line 68: I wonder if we can figure out a better way to reference the value in `kona_registry::ROLLUP_CONFIGS` across threads without cloning. Often time, we don't need the owned value. Bit of a footgun though since in some cases where updates are applied like derivation pipeline, we do want to reference the rollup config (system config) that is updated.

refcell (approved):

bin/client/src/interop/consolidate.rs
clabby: Yeah, this is something I've noted down that I'm also not a fan of. I'll look to express this in a ticket so that we can get around to cleaning it up.

That said, thankfully it's not a terribly expensive clone.

PR #1505, opened 2025-04-24
fix(node): Unsafe Block Signer

refcell:
### Description

#1498 broke the unsafe block signer by using the `batcher_address` on the `RollupConfig` which is different than the unsafe block signer.

This PR fixes the `kona-node` to use the correct unsafe block signer, allowing the user to pass in a custom unsafe block signer if desired.
Configuring the unsafe block signer can be done with `--p2p.unsafe.block.signer` cli flag (optional). By default, the `kona-node` will fetch
the unsafe block signer from the superchain-registry.


refcell:
## 📚 $\text{Stack Overview}$

* #1505 👈
* `main`

theochap (approved):

PR #1506, opened 2025-04-24
feat(interop): Add message expiry check

clabby:
## Overview

Adds a new check to the `MessageGraph` for the message expiry window.

spec ref: https://github.com/ethereum-optimism/specs/pull/671

### DIFF
--- a/crates/protocol/interop/src/constants.rs
+++ b/crates/protocol/interop/src/constants.rs
@@ -7,7 +7,7 @@ pub const CROSS_L2_INBOX_ADDRESS: Address = address!("42000000000000000000000000
 
 /// The expiry window for relaying an initiating message (in seconds).
 /// <https://specs.optimism.io/interop/messaging.html#message-expiry-invariant>
-pub const MESSAGE_EXPIRY_WINDOW: u64 = 180 * 24 * 60 * 60;
+pub const MESSAGE_EXPIRY_WINDOW: u64 = 30 * 24 * 60 * 60;
 
 /// The current version of the [SuperRoot] encoding format.
 ///
--- a/crates/protocol/interop/src/graph.rs
+++ b/crates/protocol/interop/src/graph.rs
@@ -1,7 +1,7 @@
 //! Interop [MessageGraph].
 
 use crate::{
-    RawMessagePayload,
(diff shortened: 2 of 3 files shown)


theochap (approved):
Looks good!

crates/protocol/interop/src/graph.rs
```diff
@@ -13,6 +13,9 @@ use kona_genesis::RollupConfig;
 use kona_registry::{HashMap, ROLLUP_CONFIGS};
 use tracing::{info, warn};
 
+/// The message expiry window (30 days, in seconds).
+const MESSAGE_EXPIRY_WINDOW: u64 = 30 * 24 * 60 * 60;
```
refcell, line 17: It's honestly unfortunate rust alloc doesn't have a time module for these constants

clabby: `core` does! Though, just a type. Also have to still use `from_secs`. Will update to that though.

refcell: RIP https://github.com/rust-lang/rust/blob/master/library/core/src/time.rs#L27-L39

crates/protocol/interop/src/constants.rs
```diff
@@ -1,13 +1,14 @@
...
 use alloy_primitives::{Address, address};
+use core::time::Duration;
 
 /// The address of the L2 cross chain inbox predeploy proxy.
 pub const CROSS_L2_INBOX_ADDRESS: Address = address!("4200000000000000000000000000000000000022");
 
 /// The expiry window for relaying an initiating message (in seconds).
 /// <https://specs.optimism.io/interop/messaging.html#message-expiry-invariant>
-pub const MESSAGE_EXPIRY_WINDOW: u64 = 180 * 24 * 60 * 60;
+pub const MESSAGE_EXPIRY_WINDOW: Duration = Duration::from_secs(30 * 24 * 60 * 60);
```
emhane, line 11: seems like it makes more sense to keep this as u64, since down below we call `as_secs` to get the u64 back for doing the subtraction?

emhane (changes requested):

crates/protocol/interop/src/constants.rs
clabby: I don't really mind either way personally.

Issue #1374, opened 2025-04-07
feat(bin/node): P2P Advertising Flags


refcell:
### Description

The `op-node` allows the ip address along with the tcp and udp ports that are advertised in its ENR to be configured via [these CLI args](https://github.com/ethereum-optimism/optimism/blob/b4070c5244c8c7ba6d37b9e31fa7d43ada25070d/op-node/flags/p2p_flags.go#L32-L34).

This ticket is to make these options available in the `kona-node` and wired up with `kona-p2p`.

### Task List

- [x] IP Advertising Flag (`--p2p.advertise.ip`)
- [x] TCP Port Advertising Flag (`--p2p.advertise.tcp`)
- [x] UDP Port Advertising Flag (`--p2p.advertise.udp`)
- [x] Wire up these flags into `kona-p2p`.

PR #1509, opened 2025-04-24
feat(bin/node): add advertise p2p flags

theochap:
## Description
This PR adds flags to allow specifying custom p2p advertise address for the discovery layer. This allows a local node to define a custom address to advertise to external peers. 

This is used in the Kurtosis integration of the kona-node [here](https://github.com/ethpandaops/optimism-package/pull/229/files#diff-a03b79cfb91d58ebac65800353104cf8516c01df6a4e37272337d2dc08ad8392R201-R206).

Closes [#1374](https://github.com/op-rs/kona/issues/1374)

### DIFF
--- a/bin/node/src/flags/p2p.rs
+++ b/bin/node/src/flags/p2p.rs
@@ -38,7 +38,29 @@ pub struct P2PArgs {
     /// The hex-encoded 32-byte private key for the peer ID.
     #[arg(long = "p2p.priv.raw", env = "KONA_NODE_P2P_PRIV_RAW")]
     pub private_key: Option<B256>,
-    /// IP to bind LibP2P and Discv5 to.
+
+    /// IP to advertise to external peers from Discv5.
+    /// Optional argument. Use the `p2p.listen.ip` if not set.
+    #[arg(long = "p2p.advertise.ip", env = "KONA_NODE_P2P_ADVERTISE_IP")]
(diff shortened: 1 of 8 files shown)


bin/node/src/flags/p2p.rs
```diff
@@ -36,13 +36,35 @@ pub struct P2PArgs {
...
+    #[arg(long = "p2p.advertise.ip", env = "KONA_NODE_P2P_ADVERTISE_IP")]
+    pub advertise_ip: Option<IpAddr>,
+    /// TCP port to advertise to external peers from the discovery layer. Same as `p2p.listen.tcp`
+    /// if set to zero.
+    #[arg(
+        long = "p2p.advertise.tcp",
+        default_value = "0",
+        env = "KONA_NODE_P2P_ADVERTISE_TCP_PORT"
+    )]
+    pub advertise_tcp_port: u16,
```
refcell, line 51: Only note that the approach to use `0` as a way to re-use the `p2p.listen.tcp` port prevents people from using `0` as a way to grab an ephemeral/dynamically assigned port. e.g. there may be the edge case here where the user doesn't want the advertised ports to be the same as the listening ports, but also wants to use `0` in order to get a dynamically assigned port. Probably not a big deal, but worth mentioning.

refcell (approved):
This looks good to me

PR #1516, opened 2025-04-24
feat(bin/node): also advertise udp port

theochap:
## Description
Small PR to add the discv5 UDP port to the advertised addresses from the network.
This matches the behavior of `op-node` more closely, see [the op-node codebase](https://github.com/ethereum-optimism/optimism/blob/174e55f0a1e73b49b80a561fd3fedd4fea5770c6/op-node/p2p/discovery.go#L65-L69)

### DIFF
--- a/crates/node/p2p/src/discv5/builder.rs
+++ b/crates/node/p2p/src/discv5/builder.rs
@@ -1,22 +1,61 @@
 //! Contains a builder for the discovery service.
 
-use discv5::{Config, Discv5, Enr, ListenConfig, enr::CombinedKey};
-use std::{net::SocketAddr, path::PathBuf};
+use discv5::{Config, Discv5, Enr, enr::CombinedKey};
+use std::{net::IpAddr, path::PathBuf};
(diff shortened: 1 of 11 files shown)


crates/node/p2p/src/discv5/builder.rs
```diff
@@ -69,12 +107,6 @@ impl Discv5Builder {
         self
     }
 
-    /// Sets the listen config for the discovery service.
-    pub fn with_listen_config(mut self, listen_config: ListenConfig) -> Self {
-        self.listen_config = Some(listen_config);
-        self
-    }
```
theochap, line 76: This was dead-code

refcell: Yea I think a bunch of dead code snuck in since we've been iterating pretty quickly

refcell: Thanks for removing

refcell (approved):
Nice work

refcell (approved):

PR #1523, opened 2025-04-25
fix(bin/node): fix the unsafe block signer to be compatible with unknown chain ids

theochap:
## Description
This PR fixes the unsafe block signer import method to be compatible with unknown chain ids. Before we were using the `GlobalArgs` to import the rollup config from a registry indexed by chain id. This PR fixes this to use the rollup config provided as an argument (either at a custom path, or retrieved from the registry)


refcell (approved):
Oh good catch thank you... we may want to come back and clean all this up. We're passing a lot of arguments around and obviously it's becoming error prone

PR #1525, opened 2025-04-25
fix(node/p2p): ensure that the discovery and the gossip keys match

theochap:
## Description
This PR ensures that the keys advertised by discv5 for the local node matches the one from the gossip layer. Currently the node uses different keys for discv5 and the gossip layer which may cause some of the instability we have observed at the gossip layer (nodes get dropped unexpectedly).

### DIFF
--- a/crates/node/p2p/README.md
+++ b/crates/node/p2p/README.md
@@ -16,8 +16,9 @@ Contains a gossipsub driver to run discv5 peer discovery and block gossip.
 ```rust,no_run
 use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 use alloy_primitives::address;
-use kona_p2p::{AdvertisedIpAndPort, Network};
+use kona_p2p::{LocalNode, Network};
 use libp2p::Multiaddr;
+use discv5::enr::CombinedKey;
 
 // Construct the Network
 let signer = address!("bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb");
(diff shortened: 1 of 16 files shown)


crates/node/p2p/README.md
```diff
@@ -16,16 +16,21 @@ Contains a gossipsub driver to run discv5 peer discovery and block gossip.
...
 let signer = address!("bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb");
 let gossip = SocketAddr::new(IpAddr::V4(Ipv4Addr::UNSPECIFIED), 9099);
 let mut gossip_addr = Multiaddr::from(gossip.ip());
 gossip_addr.push(libp2p::multiaddr::Protocol::Tcp(gossip.port()));
 let advertise_ip = IpAddr::V4(Ipv4Addr::UNSPECIFIED);
-let disc = AdvertisedIpAndPort::new(advertise_ip, 9097, 9098);
+
+let CombinedKey::Secp256k1(k256_key) = CombinedKey::generate_secp256k1() else {
+    unreachable!()
+};
```
emhane, line 32:
```suggestion
let CombinedKey::Secp256k1(k256_key) = CombinedKey::generate_secp256k1().expect("X should happen");
```
better use an expect statement here

emhane (approved):
lgtm! good find

crates/node/p2p/README.md
theochap: The `CombinedKey` is just an enum that doesn't implement `Result` or `Option` so the `.expect` patter doesn't seem to work here (I tried to do it locally)

clabby (approved):

PR #1527, opened 2025-04-25
chore(node/p2p): Peer Count Metric

### LINKED ISSUE #1528: feat(node/p2p): Peer Count Gauge Metric
### Description

Add a gauge metric to the `kona-p2p` crate behind a `metrics` feature flag.

This follows the [`op-node`'s peer count metric](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/metrics/metrics.go#L237-L242).

It should be incremented when a new connection is established, or decremented when a connection is dropped. (to ensure notifications aren't missed, the gauge can just be _set_ to the current peer count value when one of these notifications is encountered).

refcell:
### Description

Introduces a peer count metric.

Closes #1528

### DIFF
--- a/crates/node/p2p/src/macros.rs
+++ b/crates/node/p2p/src/macros.rs
@@ -1,0 +1,10 @@
+//! Macros for recording metrics.
+
+/// Sets a metric value.
+#[macro_export]
+macro_rules! set {
+    ($metric:ident, $value:expr) => {
+        #[cfg(feature = "metrics")]
+        $crate::metrics::$metric.set($value);
+    };
+}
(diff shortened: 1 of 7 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1527 👈
* `main`

crates/node/p2p/src/macros.rs
```diff
@@ -0,0 +1,10 @@
+//! Macros for recording metrics.
+
+/// Sets a metric value.
+#[macro_export]
+macro_rules! set {
```
emhane, line 5:
```suggestion
macro_rules! record_metric {
```
a more helpful name would be nice

refcell:
> Looks great, it could be useful to think of a way to emit that kind of metrics inside the kona-node binary as well

We should be able to pick up the metrics through the exported prometheus metrics. Planning on setting up a simple grafana dashboard so we can start to look at these metrics. Maybe it'll help us understand what's going on here.

PR #1538, opened 2025-04-25
chore(node/sources): Runtime Loading Call

refcell:
### Description

Improves the api for runtime loading to imitate how alloy returns a call object that implements the `Future` trait.

This way, the block at which to execute the call can be set on the returned object rather than having two methods in the public api.


refcell:
## 📚 $\text{Stack Overview}$

* #1538 👈
* `main`

crates/node/sources/src/runtime/call.rs
```diff
@@ -0,0 +1,93 @@
...
+#[pin_project]
+pub struct RuntimeCall {
+    #[pin]
+    loader: RuntimeLoader,
+    block_info: Option<BlockInfo>,
+    #[pin]
+    state: RuntimeCallState,
+}
+
+impl core::fmt::Debug for RuntimeCall {
```
theochap, line 20: Out of curiosity: are we manually implementing debug because the automatic trait derives are not compatible with `#[pin_project]`? Otherwise, is there a specific reason we're doing it?

crates/node/sources/src/runtime/error.rs
```diff
@@ -9,6 +11,9 @@ pub enum RuntimeLoaderError {
     /// Transport error
     #[error(transparent)]
     Transport(#[from] RpcError<TransportErrorKind>),
+    /// Transport error string.
+    #[error("Transport error: {0}")]
+    TransportString(String),
```
theochap, line 16: nit: it is not super clear what's the difference between a transport error and a transport string error. Is there more context we could add to the error variant name/error message? If not, maybe we could rename the `Transport` error above `TransportRPC` and this `TransportString` to `Transport`

crates/node/sources/src/runtime/call.rs
```diff
@@ -0,0 +1,93 @@
...
+#[pin_project(project = RuntimeCallStateProj)]
+enum RuntimeCallState {
+    Initial,
+    LoadingLatest(#[pin] Pin<Box<dyn Future<Output = Result<RuntimeConfig, RuntimeLoaderError>>>>),
+    LoadingWithBlockInfo(
+        #[pin] Pin<Box<dyn Future<Output = Result<RuntimeConfig, RuntimeLoaderError>>>>,
+    ),
+    #[allow(unused)]
+    Complete(Result<RuntimeConfig, RuntimeLoaderError>),
+}
```
theochap, line 39: All these structs are internal but I feel it would be super helpful (given that this a not trivial piece of code with `Pin`s and such) if we could have some short comments for both the enum and the `poll` method.

crates/node/sources/Cargo.toml
```diff
@@ -32,6 +33,7 @@ lru.workspace = true
 url.workspace = true
 tracing.workspace = true
 thiserror.workspace = true
+pin-project = "1.1.3"
```
theochap, line 36: Question: how are we handling the interaction between crate and workspace dependencies? Are we ok to define small crates like `pin-project` at the individual crate level, or we want to define everything at workspace to have a better grasp of the crates we import?

refcell: No we should move this to the workspace level. All imports should be at the workspace level, this is a mistake committing this on my end - I forgot to hoist it to the workspace manifest

refcell:
Closing for now, tracked by https://github.com/op-rs/kona/issues/1595

PR #1540, opened 2025-04-25
feat(proof-interop): Block replacement transaction

clabby:
## Overview

Implements the block replacement transaction in the superchain consolidation logic for the proof, specified [here](https://specs.optimism.io/interop/derivation.html#optimistic-block-deposited-transaction).


crates/proof/proof-interop/src/consolidation.rs
```diff
@@ -205,6 +215,41 @@ where
...
+        let source = InteropBlockReplacementDepositSource::new(old_output_root);
+        let output_root_preimage = {
+            let mut raw_output = [0u8; 128];
+            raw_output[31] = 0x0;
+            raw_output[32..64].copy_from_slice(old_header.state_root.as_ref());
+            raw_output[64..96]
+                .copy_from_slice(old_header.withdrawals_root.unwrap_or_default().as_ref());
+            raw_output[96..128].copy_from_slice(old_header.hash().as_ref());
+            raw_output
+        };
```
clabby, line 236: Will follow up in another PR with a dedicated `OutputRoot` type here that can handle encoding in the background. We do this enough that manual encoding is lazy and ugly. https://github.com/op-rs/kona/issues/1541

clabby: #1544

bin/client/src/interop/consolidate.rs
```diff
@@ -111,6 +111,12 @@ where
 
     // Ensure that the post-state matches the claimed post-state.
     if post_commitment != boot.claimed_post_state {
+        tracing::error!(
+            target: "client_interop",
+            "Post state validation failed. Claimed: {}, Actual: {}",
+            boot.claimed_post_state,
+            post_commitment
+        );
```
emhane, line 119:
```suggestion
        tracing::error!(
            target: "client_interop",
            claimed=%boot.claimed_post_state,
            actual=%post_commitment,
            "Post state validation failed.",
        );
```
let's use the kv format for logs

emhane (dismissed):

crates/proof/proof-interop/src/consolidation.rs
```diff
@@ -142,6 +144,29 @@ where
...
+            let original_optimistic_block = transition_state
+                .pending_progress
+                .iter_mut()
+                .find(|block| block.block_hash == header.hash())
+                .ok_or(MessageGraphError::EmptyDependencySet)?;
+
+            // Filter out all transactions that are not deposits to start.
+            let mut transactions = transactions
+                .into_iter()
+                .filter(|t| !t.is_empty() && t[0] == OpTxType::Deposit)
```
refcell, line 161: What tx type is this? Would be nice to be explicit and use a strong method to check if the tx type is a deposit tx (e.g. `t.is_deposit()`) rather than having to check if empty and the first byte.

refcell (approved):
One style nit/suggestion, otherwise looks good

crates/proof/proof-interop/src/consolidation.rs
clabby: So the transactions in the transactions trie are already 2718-encoded, meaning to use [`OpTxEnvelope::is_deposit`](https://docs.rs/op-alloy-consensus/latest/op_alloy_consensus/transaction/enum.OpTxEnvelope.html#method.is_deposit), we'd have to decode them just for this check.

I'd prefer not to mainly for the redundancy there, but not super opinionated; They're re-validated later on in the pre-validation execution step. You have a strong pref here?

refcell: No that's fine. There's a lot of repeated logic here and we've seen the issue in revm (albeit a magic number issue `0x7F` vs `0x7E`). It would be nice to consolidate this filter a bit so it can be unit tested and in one spot rather than duped. Not a blocker.

PR #1542, opened 2025-04-25
fix(host-interop): Ignore block data hint if chain hasn't progressed

clabby:
## Overview

There is a small issue in this hint currently, where it is possible for the agreed block hash to equal the disputed block hash. This is because the `SuperRoot` progresses every 1s, but the chains within it do not necessarily. In the event that the agreed block hash equals the disputed block hash, the block data hint can be ignored.

### DIFF
--- a/bin/host/src/interop/handler.rs
+++ b/bin/host/src/interop/handler.rs
@@ -36,7 +36,7 @@ use kona_protocol::BlockInfo;
 use kona_registry::ROLLUP_CONFIGS;
 use std::sync::Arc;
 use tokio::task;
-use tracing::warn;
+use tracing::{debug, warn};
 
 /// The [HintHandler] for the [InteropHost].
 #[derive(Debug, Clone, Copy)]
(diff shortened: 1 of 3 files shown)


bin/host/src/interop/handler.rs
```diff
@@ -407,6 +407,13 @@ impl HintHandler for InteropHintHandler {
                 let disputed_block_hash = B256::from_slice(&hint.data.as_ref()[32..64]);
                 let chain_id = u64::from_be_bytes(hint.data.as_ref()[64..72].try_into()?);
 
+                // Return early if the agreed and disputed block are the same. This can occur when
+                // the chain has not progressed past its prestate, but the super root timestamp has
+                // progressed.
+                if agreed_block_hash == disputed_block_hash {
```
emhane, line 413: probably we want to add a trace log here, including the text from the code comment informing the debugging user why this might have happened

emhane (approved):

bin/host/src/interop/handler.rs
```diff
@@ -36,7 +36,7 @@ use kona_protocol::BlockInfo;
 use kona_registry::ROLLUP_CONFIGS;
 use std::sync::Arc;
 use tokio::task;
-use tracing::warn;
+use tracing::{debug, warn};
```
refcell, line 39: FWIW, don't need to have these imports if you do this in the top-level `lib.rs`

```
#[macro_use]
extern crate tracing;
```

clabby: will follow up - agree this is cleaner.

PR #1547, opened 2025-04-26
chore(ci): Remove interop proof from codecov ignore

clabby:
## Overview

Removes the `kona-proof-interop` and client/host interop variants from the `codecov.yml`'s ignore list.


emhane (approved):
whoop whoop

PR #1548, opened 2025-04-27
perf(ci): Cache forge artifacts in action tests

clabby:
## Overview

Caches foundry artifacts in the action tests workflow. Speeds up the job significantly, with contracts often taking 4-5 minutes to build.


emhane (approved):
nice!

PR #1549, opened 2025-04-28
feat(protocol): Complete `Predeploys` definition

clabby:
## Overview

Small aesthetic / consistency cleanup across the codebase - completes the definition of predeploys within `kona-protocol`'s `Predeploys` struct, and uses them everywhere rather than the odd one-off local constants in a few crates.

### DIFF
--- a/crates/protocol/protocol/src/predeploys.rs
+++ b/crates/protocol/protocol/src/predeploys.rs
@@ -12,9 +12,82 @@ pub struct Predeploys;
 
 impl Predeploys {
     /// List of all predeploys.
-    pub const ALL: [Address; 1] = [Self::L2_TO_L1_MESSAGE_PASSER];
+    pub const ALL: [Address; 18] = [
+        Self::WETH9,
+        Self::L2_CROSS_DOMAIN_MESSENGER,
+        Self::L2_STANDARD_BRIDGE,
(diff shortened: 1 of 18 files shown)


crates/protocol/protocol/src/predeploys.rs
```diff
@@ -12,9 +12,82 @@ pub struct Predeploys;
...
+        Self::PROXY_ADMIN,
+        Self::BASE_FEE_VAULT,
+        Self::L1_FEE_VAULT,
+        Self::SCHEMA_REGISTRY,
+        Self::EAS,
+        Self::OPERATOR_FEE_VAULT,
+        Self::CROSS_L2_INBOX,
+    ];
+
+    /// The WETH9 predeploy address.
```
emhane, line 36: would be really nice if we could link to the solidity code for each one, like this
```suggestion
    /// The WETH9 predeploy address.
    /// `WETH9.sol`: <https://optimistic.etherscan.io/address/0x4200000000000000000000000000000000000006#code#L1>
```

emhane (approved):

crates/protocol/protocol/src/predeploys.rs
clabby: Good thinking. Will do when I'm at my desk

PR #1551, opened 2025-04-28
fix: allow advertise port to have dynamic value

### LINKED ISSUE #1514: fix(bin/node): allow advertised ports to be automatically chosen by the client
## Description

Relevant discussion: https://github.com/op-rs/kona/pull/1509#discussion_r2059107984

We should have a way to randomly assign a port to advertise if advertise port is manually set to 0. For that we can change the CLI type for the advertised port to become an `Option<u16>` and distinguish between the `None` and the `Some(0)` cases.

varun-doshi:
Closes #1514 

`advertise_tcp_port` can now be:
- Value -> if user specifies
- 0 -> dynamic port assigned
- None -> equal to `listen_tcp_port`

### DIFF
--- a/bin/node/src/flags/p2p.rs
+++ b/bin/node/src/flags/p2p.rs
@@ -51,7 +51,7 @@ pub struct P2PArgs {
         default_value = "0",
         env = "KONA_NODE_P2P_ADVERTISE_TCP_PORT"
     )]
-    pub advertise_tcp_port: u16,
+    pub advertise_tcp_port: Option<u16>,
     /// UDP port to advertise to external peers from the discovery layer.
     /// Same as `p2p.listen.udp` if set to zero.
     #[arg(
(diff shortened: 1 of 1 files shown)


bin/node/src/flags/p2p.rs
```diff
@@ -304,11 +304,7 @@ impl P2PArgs {
         let advertise_ip = self.advertise_ip.unwrap_or(self.listen_ip);
 
         // If the advertise tcp port is null, use the listen tcp port
-        let advertise_tcp_port = if self.advertise_tcp_port != 0 {
-            self.advertise_tcp_port
-        } else {
-            self.listen_tcp_port
-        };
+        let advertise_tcp_port = self.advertise_tcp_port.unwrap_or(self.listen_tcp_port);
```
refcell, line 307: I don't think this is going to work as intended. Since the `default_value = 0`, from the CLI, you won't be able to disable the advertise port. Let's add unit tests here to verify e.g. `mock-command --p2p.advertise.tcp none` and without the flag specified (default).

refcell:
Going to close this for now since having a dynamic port advertised is the same as not setting the advertised port, since the ENR is automatically updated.

PR #1553, opened 2025-04-28
feat(bin/node): Extend P2P Configurability

refcell:
### Description

Extends P2P configurability options in the cli.

### DIFF
--- a/bin/node/src/commands/discover.rs
+++ b/bin/node/src/commands/discover.rs
@@ -125,6 +125,7 @@ impl Discovery {
             .with_local_node(ip_and_port)
             .with_store_interval(std::time::Duration::from_secs(5))
             .with_chain_id(self.l2_chain_id)
+            // Disable forwarding since gossip isn't important for this command.
             .disable_forward();
         let discovery = discovery_builder.build()?;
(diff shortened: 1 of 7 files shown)


bin/node/src/commands/discover.rs
```diff
@@ -120,13 +120,13 @@ impl Discovery {
...
             .with_local_node(ip_and_port)
-            .with_chain_id(self.l2_chain_id);
-        let mut discovery = discovery_builder.build()?;
-        discovery.interval = std::time::Duration::from_secs(2);
-        discovery.forward = false;
-        discovery.remove_interval = Some(std::time::Duration::from_secs(2));
-        discovery.store_interval = std::time::Duration::from_secs(5);
+            .with_store_interval(std::time::Duration::from_secs(5))
+            .with_chain_id(self.l2_chain_id)
+            .disable_forward();
```
theochap, line 128: Question: That may be obvious but why do we want to disable enr forwarding by default for the discover command? We never want to use bootstrap ENRs to test the discovery layer? If so maybe a small comment to explain would be nice

theochap (approved):
Nice! Thanks for that

PR #1554, opened 2025-04-28
fix(node/p2p): Wait for Swarm Address Dialing

refcell:
### Description

Waits for the libp2p swarm to establish listening on its address before proceeding with peer connection.

Avoids the situation where it takes the swarm time to listen on the given address while the network service is sending over peers for it to connect to.


refcell:
## 📚 $\text{Stack Overview}$

* #1554 👈
* `main`

theochap (approved):

PR #1555, opened 2025-04-29
fix(std-fpvm): Large file IO

clabby:
## Overview

Small fix for large IO operations using `NativeClientIO` in the FPVM environment. This hasn't come up before for us, since all reads are generally within the size of an operating system page on the host machine. However, with the `L2PayloadWitness` hint, we easily breach that boundary.

### DIFF
--- a/bin/host/src/backend/online.rs
+++ b/bin/host/src/backend/online.rs
@@ -97,8 +97,9 @@ where
     async fn route_hint(&self, hint: String) -> PreimageOracleResult<()> {
         trace!(target: "host_backend", "Received hint: {hint}");
 
-        let parsed_hint =
-            hint.parse::<Hint<C::HintType>>().map_err(|_| PreimageOracleError::KeyNotFound)?;
+        let parsed_hint = hint
+            .parse::<Hint<C::HintType>>()
+            .map_err(|e| PreimageOracleError::HintParseFailed(e.to_string()))?;
         if self.proactive_hints.contains(&parsed_hint.ty) {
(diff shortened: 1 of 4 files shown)


crates/proof/std-fpvm/src/io.rs
```diff
@@ -30,7 +30,7 @@ cfg_if! {
             fn read(fd: FileDescriptor, buf: &mut [u8]) -> IOResult<usize> {
                 unsafe {
                     let mut file = File::from_raw_fd(fd as i32);
```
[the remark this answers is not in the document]
clabby: In this case, this is what we want.

bin/host/src/backend/online.rs
```diff
@@ -97,8 +97,9 @@ where
     async fn route_hint(&self, hint: String) -> PreimageOracleResult<()> {
         trace!(target: "host_backend", "Received hint: {hint}");
 
-        let parsed_hint =
-            hint.parse::<Hint<C::HintType>>().map_err(|_| PreimageOracleError::KeyNotFound)?;
+        let parsed_hint = hint
+            .parse::<Hint<C::HintType>>()
+            .map_err(|e| PreimageOracleError::Other(format!("failed to parse hint: {e}")))?;
```
emhane, line 102: can we do this properly and add another variant instead pls

emhane (approved):
nice! smol nitpick about error variant

bin/host/src/backend/online.rs
refcell: + a unit test

refcell (approved):

PR #1556, opened 2025-04-29
chore(client): Bump heap size

clabby:
## Overview

Bumps the heap size of the client program to 200MB. As chains start to raise their block sizes, this should give us a bit of headroom.

In the near future, we should consider writing our own allocator for the FPVM to `mmap` new memory when necessary.


emhane (changes requested):
have we got any benchmarks to suggest this is needed?

clabby:
> have we got any benchmarks to suggest this is needed?

Screenshot 2025-04-29 at 9 13 27 AM

memory usage has been steadily on the rise over the past 30d on a few networks on VM Runner. We haven't OOM'd yet, but it's likely we will soon. I'm going to get together a bump allocator that actually `mmap`s so we don't have to concern ourselves with this, but for now, just adding a bit more headroom.

clabby:
Closing in favor of #1567

PR #1558, opened 2025-04-29
feat(node/p2p): disable dynamic ENR updates for static IPs

theochap:
## Description
This PR makes sure that dynamic ENR updates (at the discovery layer) are disabled when explicitly setting the advertised IP. This respects the behavior of the op-node that treats the advertised IP as static IP, hence that shouldn't be dynamically updated by the discovery layer. See [here](https://github.com/ethereum-optimism/optimism/blob/174e55f0a1e73b49b80a561fd3fedd4fea5770c6/op-node/p2p/discovery.go#L63)

Small note: there is also a small [one liner](https://github.com/op-rs/kona/pull/1558/files#diff-d72b5a9e5e10500d4de868bc3be0bcdc32530ba30a380f1c8b4919cfebdb92d3R374) that is an artifact of #1525 (it seems I missed that one)

### DIFF
--- a/bin/node/src/flags/p2p.rs
+++ b/bin/node/src/flags/p2p.rs
@@ -371,7 +371,7 @@ impl P2PArgs {
             discovery_interval: Duration::from_secs(self.discovery_interval),
             discovery_address,
             gossip_address,
-            keypair: self.keypair().unwrap_or_else(|_| Keypair::generate_secp256k1()),
+            keypair,
             unsafe_block_signer: self.unsafe_block_signer(config, args, l1_rpc).await?,
             gossip_config,
             scoring: self.scoring,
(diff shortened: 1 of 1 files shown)


bin/node/src/flags/p2p.rs
```diff
@@ -42,6 +42,10 @@ pub struct P2PArgs {
 
     /// IP to advertise to external peers from Discv5.
     /// Optional argument. Use the `p2p.listen.ip` if not set.
+    ///
+    /// Technical note: if this argument is set, the dynamic ENR updates from the discovery layer
+    /// will be disabled. This is to allow the advertised IP to be static (to use in a network
+    /// behind a NAT for instance).
```
refcell, line 48: Nice doc comment

refcell (approved):

PR #1560, opened 2025-04-29
fix(bin/node): Regerated Keypair

refcell:
### Description

Previously the keypair was being regenerated even though we constructed and used the keypair earlier on.


refcell:
## 📚 $\text{Stack Overview}$

* #1560 👈
* `main`

PR #1561, opened 2025-04-29
fix(node/p2p): Fix multiaddress translation

theochap:
## Description
Small PR that fixes the enr to multiaddress translation inside the kona-node. The produced multiaddresses did not contain the p2p (peer id) and the udp protocols from the ENRs.

### DIFF
--- a/crates/node/p2p/Cargo.toml
+++ b/crates/node/p2p/Cargo.toml
@@ -35,6 +35,7 @@ snap.workspace = true
 futures.workspace = true
 discv5 = { workspace = true, features = ["libp2p"] }
 libp2p = { workspace = true, features = ["macros", "tokio", "tcp", "noise", "gossipsub", "ping", "yamux"] }
+libp2p-identity.workspace = true
 openssl = { workspace = true, features = ["vendored"] }
 
 # Cryptography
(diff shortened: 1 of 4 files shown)


crates/node/p2p/Cargo.toml
```diff
@@ -35,6 +35,7 @@ snap.workspace = true
 futures.workspace = true
 discv5 = { workspace = true, features = ["libp2p"] }
 libp2p = { workspace = true, features = ["macros", "tokio", "tcp", "noise", "gossipsub", "ping", "yamux"] }
+libp2p-identity.workspace = true
```
refcell, line 38: I don't think we need this, the `libp2p` crate itself re-exports the `libp2p-identity` crate through [`libp2p::identity`](https://docs.rs/libp2p/latest/libp2p/identify/index.html)

crates/node/p2p/src/peers/utils.rs
```diff
@@ -1,23 +1,44 @@
...
+        addr.push(Protocol::Udp(socket.port()));
     }
-    None
+
+    let CombinedPublicKey::Secp256k1(pub_key) = enr.public_key() else {
+        return None;
+    };
+
+    let pub_key = libp2p_identity::secp256k1::PublicKey::try_from_bytes(&pub_key.encode()).ok()?;
+    let pub_key = libp2p_identity::PublicKey::from(pub_key);
```
refcell, line 37:
```suggestion
    let pub_key = libp2p::identity::secp256k1::PublicKey::try_from_bytes(&pub_key.encode()).ok()?;
    let pub_key = libp2p::identity::PublicKey::from(pub_key);
```

refcell (approved):
Very nice!

PR #1590, opened 2025-04-29
perf(std-fpvm): Switch to `buddy_system_allocator`

clabby:
## Overview

Switches `kona-std-fpvm` over to the `buddy_system_allocator`, now that memory usage isn't a very sensitive resource in the FPVM variant of the proof program (after #1567)

The `buddy_system_allocator` is much more efficient for some of the large allocations we have to do, for example during channel decompression. It does, however, come with the tradeoff of a larger memory footprint, since it allocates in `2^n` byte blocks.

### Benchmarks (`linked_list_allocator` (current) vs. `buddy_system_allocator`)

buddy_benchmarks

### DIFF
--- a/crates/proof/std-fpvm/src/malloc.rs
+++ b/crates/proof/std-fpvm/src/malloc.rs
@@ -7,11 +7,14 @@
 /// The global allocator for the program in embedded environments.
 #[cfg(any(target_arch = "mips64", target_arch = "riscv64"))]
 pub mod global_allocator {
-    use linked_list_allocator::LockedHeap;
+    use buddy_system_allocator::LockedHeap;
+
+    /// The maximum block size, as a power of two, for the buddy system allocator.
+    const HEAP_ORDER: usize = 32;
 
     /// The global allocator for the program.
     #[global_allocator]
-    static ALLOCATOR: LockedHeap = LockedHeap::empty();
+    static ALLOCATOR: LockedHeap<HEAP_ORDER> = LockedHeap::empty();
 
     /// Initialize the [SpinLockedAllocator] with the following parameters:
     /// * `heap_start_addr` is the starting address of the heap memory region,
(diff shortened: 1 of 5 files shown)


crates/proof/std-fpvm/src/malloc.rs
```diff
@@ -7,11 +7,14 @@
 /// The global allocator for the program in embedded environments.
 #[cfg(any(target_arch = "mips64", target_arch = "riscv64"))]
 pub mod global_allocator {
-    use linked_list_allocator::LockedHeap;
+    use buddy_system_allocator::LockedHeap;
+
+    /// The maximum block size, as a power of two, for the buddy system allocator.
+    const HEAP_ORDER: usize = 32;
```
refcell, line 13: Should this be configurable?

refcell (approved):

crates/proof/std-fpvm/src/malloc.rs
clabby: Don't think we need it to be, but could be! Buddy system allows for anything from `2^0 .. 2^n` blocks to be allocated, and `2^32` is 4 gigabytes. There's no single allocation we ever make that's over `40MB` from what I've seen so far, so I think this # of "buckets" should be sufficient.

Inphi: Do you expect this maximum of `40MB` for a single allocation to change after Pectra with more blobs?

clabby: @Inphi Not because of pectra, but yes. It's likely we see larger decompressed batch sizes due to block size increases on L2s, and that's the largest allocations we usually see. But each of the blobs themselves (containing _compressed_ data) are constant sized, so individually, the allocations for the blob data will remain the same.

clabby: Also note that we're specifically talking about _individual_ allocations here. Memory usage with pectra should go up, just individual allocation size probably won't due to pectra itself.

PR #1592, opened 2025-04-29
fix(node/p2p): enable identify protocol. Additional small fixes

### LINKED ISSUE #1396: fix(node/p2p): Peers stop sending blocks in sync
Peers stop sending unsafe blocks when syncing. Swarm has 13 peers but all peers are useless not sending any unsafe blocks. Investigation needed to discover root cause of this.

theochap:
## Description
This PR brings an additional wave of fixes to the p2p layer for the kona node. In particular this PR:
- Enables the `identify` protocol for node discovery. This helps kickstart the gossip layer from the op-node
- Does a small change to the RPCClients to remove the `is_local` flag
- Removes the udp protocol from the converted multiaddresses from ENRs (it produces malformed multiaddresses)
- Does a couple of small log visibility changes that were useful to debug the kona-node inside kurtosis

## Note
Merging this PR will fix the p2p gossip for kona-nodes and allow them to receive/gossip blocks on the p2p layer. Kona-nodes can also communicate with op-nodes (tested in Kurtosis)

To some extent, merging this PR will fix #1396 

## What remains to do
It seems that the peers get graylisted after a while and eventually get ejected from the gossip layer. We'll need to fine-tune network parameters to make sure this doesn't happen

### DIFF
--- a/crates/node/p2p/src/gossip/builder.rs
+++ b/crates/node/p2p/src/gossip/builder.rs
@@ -124,7 +124,7 @@ impl GossipDriverBuilder {
     pub fn build(mut self) -> Result<GossipDriver, GossipDriverBuilderError> {
         // Extract builder arguments
         let timeout = self.timeout.take().unwrap_or(Duration::from_secs(60));
-        let keypair = self.keypair.take().unwrap_or(Keypair::generate_secp256k1());
+        let keypair = self.keypair.take().ok_or(GossipDriverBuilderError::MissingKeyPair)?;
         let chain_id = self.chain_id.ok_or(GossipDriverBuilderError::MissingChainID)?;
         let addr = self.gossip_addr.take().ok_or(GossipDriverBuilderError::GossipAddrNotSet)?;
         let signer_recv = self.signer.ok_or(GossipDriverBuilderError::MissingUnsafeBlockSigner)?;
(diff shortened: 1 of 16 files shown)


crates/node/p2p/src/gossip/builder.rs
```diff
@@ -124,7 +124,7 @@ impl GossipDriverBuilder {
     pub fn build(mut self) -> Result<GossipDriver, GossipDriverBuilderError> {
         // Extract builder arguments
         let timeout = self.timeout.take().unwrap_or(Duration::from_secs(60));
-        let keypair = self.keypair.take().unwrap_or(Keypair::generate_secp256k1());
+        let keypair = self.keypair.take().ok_or(GossipDriverBuilderError::MissingKeyPair)?;
```
refcell, line 127: Nice, makes a lot more sense to leave it up to the user to set sensible defaults

crates/node/p2p/src/gossip/driver.rs
```diff
@@ -178,12 +178,12 @@ impl GossipDriver {
 
         match self.swarm.dial(addr.clone()) {
             Ok(_) => {
-                event!(tracing::Level::TRACE, peer=%addr, "Dialed peer");
+                trace!(peer=?addr, "Dialed peer");
```
refcell, line 181: nit, add a target here

```suggestion
                trace!(target: "gossip", peer=?addr, "Dialed peer");
```

refcell (approved):
Small nit, otherwise nice work!

PR #1597, opened 2025-04-30
feat(protocol): Interop upgrade transactions

### LINKED ISSUE #1475: feat(derive): Interop network upgrade transactions
## Overview

The interop hardfork will come with several network upgrade transactions. The spec hasn't yet landed, but once it does, we'll need to add these into our `kona-hardforks` crate.

I'll update this ticket once they're available.

clabby:
## Overview

Adds a new module to `kona-hardforks` for the interop upgrade transactions, specified in the [interop derivation spec](https://specs.optimism.io/interop/derivation.html#network-upgrade-transactions).

closes #1475


crates/protocol/hardforks/src/interop.rs
```diff
@@ -0,0 +1,223 @@
...
+            },
+            TxDeposit {
+                source_hash: Self::upgrade_l2_to_l2_xdm_proxy_source(),
+                from: Address::ZERO,
+                to: TxKind::Call(Predeploys::L2_TO_L2_XDM),
+                mint: None,
+                value: U256::ZERO,
+                gas_limit: 50_000,
+                is_system_transaction: false,
+                input: super::upgrade_to_calldata(Self::NEW_CROSS_L2_INBOX_IMPL),
```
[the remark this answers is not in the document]
clabby: Good catch chat gippity

crates/protocol/hardforks/Cargo.toml
```diff
@@ -27,6 +27,8 @@ op-alloy-consensus.workspace = true
 
 [dev-dependencies]
 alloy-primitives = { workspace = true, features = ["rand", "arbitrary"] }
+revm.workspace = true
+op-revm.workspace = true
```
teddyknox, line 31: Q: If this change includes this cargo module in the revm and op-revm workspace, what was the state before?

crates/protocol/hardforks/src/bytecode/crossl2inbox_interop.hex
teddyknox, line 1: Matches [spec](https://github.com/ethereum-optimism/specs/blob/5f81bfa3fef346d8312e5a4b344bb839bc8234dd/specs/static/bytecode/interop-cross-l2-inbox-deployment.txt).

crates/protocol/hardforks/src/bytecode/l2tol2_xdm_interop.hex
teddyknox, line 1: Matches [spec](https://github.com/ethereum-optimism/specs/blob/5f81bfa3fef346d8312e5a4b344bb839bc8234dd/specs/static/bytecode/interop-l2-to-l2-cross-domain-messenger-deployment.txt).

teddyknox (approved):
Main comment is that derivation should be updated to not accept an upgrade block that contains user transactions. This behavior can be added in a follow-up PR if it isn't already programmed in.

crates/protocol/hardforks/Cargo.toml
clabby: This is just a dev dependency for the `hardforks` crate - added a test utility in this PR that actually deploys the bytecode and checks for the deployed codehash.

clabby:
> Main comment is that derivation should be updated to not accept an upgrade block that contains user transactions. This behavior can be added in a follow-up PR if it isn't already programmed in.

Good callout - The specs state "The execution payload MUST set noTxPool to true for this block, excluding user transactions from the block."

and this should already be the case:

https://github.com/op-rs/kona/blob/d9f240a80f0175c86ec62f14d574c757ad06b517/crates/protocol/derive/src/attributes/stateful.rs#L187

PR #1600, opened 2025-04-30
fix(bin/node): Enable Metrics

refcell:
### Description

Enables metrics by toggling `kona-p2p` `metrics` feature flag on for the `kona-node` binary.

### DIFF
--- a/bin/node/Cargo.toml
+++ b/bin/node/Cargo.toml
@@ -19,13 +19,14 @@ workspace = true
 kona-derive.workspace = true
 kona-sources.workspace = true
 kona-cli.workspace = true
-kona-p2p.workspace = true
 kona-engine.workspace = true
 kona-genesis.workspace = true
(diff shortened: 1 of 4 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1600 👈
* `main`

theochap (approved):
Looks good! Just one question: maybe we should precise in the CLI args (for the prometheus metrics) that the binary must be compiled with the metrics feature to actually retrieve the metrics?

clabby (approved):

clabby:
> maybe we should precise in the CLI args (for the prometheus metrics) that the binary must be compiled with the metrics feature to actually retrieve the metrics?

This sounds good. Can be done in a follow up, enabling by default is a sane setting and can't imagine many people will want to disable them at comptime 👍

PR #1602, opened 2025-04-30
feat(protocol): Interop transition batch validity rule

clabby:
## Overview

Adds a new rule to `SingleBatch` validity for the interop hardfork, where in the transition block, the batch _must_ be empty.

spec ref: https://github.com/ethereum-optimism/specs/pull/692

### DIFF
--- a/crates/protocol/protocol/src/batch/single.rs
+++ b/crates/protocol/protocol/src/batch/single.rs
@@ -7,6 +7,7 @@ use alloy_primitives::{BlockHash, Bytes};
 use alloy_rlp::{RlpDecodable, RlpEncodable};
 use kona_genesis::RollupConfig;
 use op_alloy_consensus::OpTxType;
+use tracing::warn;
 
 /// Represents a single batch: a single encoded L2 block
 #[derive(Debug, Default, RlpDecodable, RlpEncodable, Clone, PartialEq, Eq)]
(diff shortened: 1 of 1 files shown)


crates/protocol/protocol/src/batch/single.rs
```diff
@@ -155,6 +156,19 @@ impl SingleBatch {
             }
         }
 
+        // If this is the first block in the interop hardfork, and the batch contains any
+        // transactions, it must be dropped.
+        if !cfg.is_interop_active(self.timestamp - cfg.block_time) &&
+            cfg.is_interop_active(self.timestamp) &&
```
sebastianst, line 162: May be worth adding a `is_interop_activation_block` function? We have this in Go and use it also e.g. in the sequencer for the upgrade block.

crates/protocol/protocol/src/batch/single.rs
```diff
@@ -580,4 +594,36 @@ mod tests {
...
+        let l1_blocks = vec![BlockInfo::default(), BlockInfo::default()];
+        let l2_safe_head = L2BlockInfo {
+            block_info: BlockInfo { timestamp: 0, ..Default::default() },
+            ..Default::default()
+        };
+        let inclusion_block = BlockInfo::default();
+        assert_eq!(
+            single_batch.check_batch(&cfg, &l1_blocks, l2_safe_head, &inclusion_block),
+            BatchValidity::Drop
+        );
```
sebastianst, line 627: We also capture and assert logs in our Go tests to make sure we drop batches for the expected reason, not any other. Is there something like this available?

sebastianst (approved):
Place and logic of the check lgtm!

crates/protocol/protocol/src/batch/single.rs
clabby: Yeah; I can add that in. We have a trace collector.

crates/protocol/protocol/src/batch/single.rs
clabby: For sure. I'll ticket that out for a follow up.

clabby: https://github.com/op-rs/kona/issues/1603

theochap (approved):

PR #1609, opened 2025-05-01
fix(node/p2p): Use Compat Metrics Crate

### LINKED ISSUE #1347: feat(node/p2p): Gossip Message Metrics
### Description

Inside the node's libp2p gossipsub driver, events + messages are [handled according to their enum variants](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/gossip/driver.rs#L123-L162). Certain variants are ignored since there's no handling to be done, but we'll still want to record metrics for when those events happen.

For example, when the [`libp2p::gossipsub::Event::SlowPeer`](https://docs.rs/libp2p/latest/libp2p/gossipsub/enum.Event.html) variant is received, a metric should be updated to include that.

This ticket is to extend P2P metrics to the gossip driver, specifically its event handling.

### LINKED ISSUE #1530: feat(node/p2p): Total Gossip Event Count
### Description

Add a [counter vec](https://docs.rs/prometheus/latest/prometheus/type.IntCounterVec.html) metric to the `kona-p2p` crate behind a `metrics` feature flag.

This mirrors the [`op-node`'s gossip events total metric](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/metrics/metrics.go#L136C32-L136C42).

Whenever [a new gossip event is received](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/net/driver.rs#L94-L99), record this under the new counter vec.

### LINKED ISSUE #1531: feat(node/p2p): Dial Count Metrics
### Description

Add a [counter vec](https://docs.rs/prometheus/latest/prometheus/type.IntCounterVec.html) metric to the `kona-p2p` crate behind a `metrics` feature flag.

This mirrors the [`op-node`'s dials metric](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/metrics/metrics.go#L140C2-L140C7).

Whenever [a peer is successfully dialed](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/gossip/driver.rs#L164-L173) record this under the metric.

### LINKED ISSUE #1533: feat(node/p2p): Broadcasted Unsafe Blocks
### Description

Add a [int gauge](https://docs.rs/prometheus/latest/prometheus/type.IntGauge.html) metric to the `kona-p2p` crate behind a `metrics` feature flag. This metric tracks the number of unsafe blocks that are broadcasted from the `kona-p2p` network driver through the [broadcast wrapper](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/net/broadcast.rs#L32-L49).

refcell:
### Description

Uses the [`metrics`](https://crates.io/crate/metrics) crate which is compatible with `metrics-exporter-prometheus` instead of the incompatible `prometheus` crate.

This allows us to have visibility into kurtosis through the already provided grafana service.

Closes #1347 
Closes #1530 
Closes #1531 
Closes #1533


refcell:
## 📚 $\text{Stack Overview}$

* #1609 👈
* `main`

crates/node/p2p/src/gossip/driver.rs
```diff
@@ -220,14 +227,34 @@ impl GossipDriver {
...
+                    gauge,
+                    crate::Metrics::GOSSIP_EVENT,
+                    "unsubscribed",
+                    topic.to_string()
+                );
             }
             libp2p::gossipsub::Event::SlowPeer { peer_id, .. } => {
                 trace!(target: "gossip", "Slow peer: {:?}", peer_id);
-                // TODO: Metrice slow peer count
+                kona_macros::inc!(gauge, crate::Metrics::GOSSIP_EVENT, "slow_peer", "slow_peer");
```
theochap, line 248: nit: Can we put the `peer_id` as the metric value?

crates/node/p2p/src/gossip/driver.rs
```diff
@@ -220,14 +227,34 @@ impl GossipDriver {
...
             }
-            _ => {
-                trace!(target: "gossip", "Ignoring non-message gossipsub event: {:?}", event)
+            libp2p::gossipsub::Event::GossipsubNotSupported { peer_id } => {
+                trace!(target: "gossip", "Peer: {:?} does not support gossipsub", peer_id);
+                kona_macros::inc!(
+                    gauge,
+                    crate::Metrics::GOSSIP_EVENT,
+                    "not_supported",
+                    "not_supported"
```
theochap, line 256: nit: Can we put the `peer_id` as the value for this metric?

crates/node/p2p/src/net/driver.rs
```diff
@@ -89,7 +89,7 @@ impl Network {
                         }
                     }
                     event = self.gossip.select_next_some() => {
-                        trace!("Received event: {:?}", event);
+                        kona_macros::inc!(gauge, crate::Metrics::GOSSIP_EVENT, "total", "total");
```
theochap, line 92: nit: would it be possible to put an event ID or string as the value for this one?

crates/node/p2p/src/gossip/driver.rs
```diff
@@ -238,22 +265,40 @@ impl GossipDriver {
...
                 }
                 return None;
             }
+            SwarmEvent::IncomingConnectionError { error, .. } => {
+                trace!(target: "gossip", "Incoming connection error: {:?}", error);
+                kona_macros::inc!(
+                    gauge,
+                    crate::Metrics::GOSSIPSUB_CONNECTION,
+                    "type",
+                    "incoming_error"
```
theochap, line 293: nit: Maybe it would be more helpful to replace the `type` key with `incoming_error` (or `error`) and put the `error` (string formatted) as a value. We could do that here (https://github.com/op-rs/kona/pull/1609/files#diff-7b46094fda6fd309a217bf2c133fdcbc33b9dbdb9c8290d08d553ff276d7ca30R280) as well

crates/node/p2p/src/gossip/driver.rs
refcell: Are you referring to the label value so instead of (`slow_peer`, `slow_peer`), it'd be (`slow_peer`, peer_id)?

theochap (commented):
This looks super useful. Left a few comments on similar patterns where I believe having a bit more precision would help quite a bit. Worst case, to get aggregate metric counts we can always group the metrics by keys

crates/node/p2p/src/net/driver.rs
refcell: Doesn't seem like there's an easy way to do this. We track individual events separately so it wouldn't provide any additional metric coverage.

crates/node/p2p/src/gossip/driver.rs
refcell: The errors are in the logs and putting the errors in the metrics will cause a lot of bloat. I've updated with the connection id and peer id for the other error type (outgoing). Let's start with this and we can always update later.

crates/node/p2p/src/gossip/driver.rs
theochap: Yes!

theochap (approved):

PR #1611, opened 2025-05-01
feat(interop): Use `FpvmOpEvmFactory` in interop proof program

### LINKED ISSUE #1543: chore(interop): Use FPVM EVM factory in client + host
## Overview

When we migrated over to `op-alloy`, I overlooked using the new `FpvmOpEvmFactory` that supports accelerated precompiles in the client + host. Both the `transition.rs` and hint handler for interop need to be updated.

clabby:
## Overview

Uses the `FpvmOpEvmFactory` in the interop proof program, adding support for accelerated precompiles.

closes #1543

### DIFF
--- a/bin/client/src/interop/consolidate.rs
+++ b/bin/client/src/interop/consolidate.rs
@@ -3,6 +3,7 @@
 use super::FaultProofProgramError;
 use crate::interop::util::fetch_output_block_hash;
 use alloc::sync::Arc;
+use alloy_evm::{EvmFactory, FromRecoveredTx, FromTxWithEncoded};
 use core::fmt::Debug;
 use kona_executor::TrieDBProvider;
 use kona_preimage::{HintWriterClient, PreimageOracleClient};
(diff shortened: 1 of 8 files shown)


bin/client/src/interop/consolidate.rs
```diff
@@ -20,13 +23,16 @@ use tracing::{error, info};
 /// superchain and ensuring that all dependencies are satisfied.
 ///
 /// [OptimisticBlock]: kona_proof_interop::OptimisticBlock
-pub(crate) async fn consolidate_dependencies<P, H>(
+pub(crate) async fn consolidate_dependencies<P, H, Evm>(
     oracle: Arc<CachingOracle<P, H>>,
     mut boot: BootInfo,
+    evm_factory: Evm,
```
theochap, line 29: Are we ever going to mutate the `Evm` factory? If no, maybe we could simplify the where bounds by wrapping it in an `Arc` (that way we gain `Send + Sync + Clone + 'static` at least)?

clabby: Nope - but good idea, should look around for cases where we have long constraints on immutables :) There's a follow up to this that removes the new parameter here, needed to refactor the type constraints on the FPVM-oriented EVM factory a bit and didn't want to bloat this PR. #1613

theochap (approved):
Looks good!

PR #1632, opened 2025-05-02
chore(genesis): Use `abi_decode_validate`

clabby:
## Overview

`alloy-sol-types` `v1.1.0` re-introduces validation when ABI decoding. This PR reworks the `SystemConfig` event update parsing logic to use that, rather than the hacky external checks that we had to use with `v1.0.0`.


refcell (approved):
So much nicer thanks

PR #1633, opened 2025-05-02
chore(node/rpc): Refactor p2p rpc

theochap:
## Description
Refactoring PR that tidies up the RPC implementation of kona-p2p in preparation of the implementation of the node rpc api.

Please note that I had to slightly change the feature-gating of the `Metrics` struct in the `p2p` crate to allow access to the `Metrics` struct inside `net` without having to feature-gate it's usage every time

Progress towards #1634


crates/node/p2p/src/rpc/types.rs
refcell, line 1: Would like to see these types split into separate modules in the future instead of having a large 376 line file. Not a blocker for this PR since I know it was ported over. Bad practice set by me 😅

refcell (approved):
I'm happy with this

PR #1635, opened 2025-05-02
fix(node/engine): Don't use Genesis in EngineState

refcell:
### Description

Removes the genesis block info fallback in the `EngineStateBuilder`.

By letting the engine state have default block info for the safe and finalized heads, EL sync is correctly kicked off since the hashes are zeroed.


refcell:
## 📚 $\text{Stack Overview}$

* #1635 👈
* `main`

PR #1640, opened 2025-05-02
fix(node/service): Wait to Kickstart Derivation

refcell:
### Description

Waits for the engine to finish syncing prior to progressing on the derivation pipeline.


crates/node/service/src/actors/derivation.rs
```diff
@@ -30,6 +30,10 @@ where
     pipeline: P,
     /// The latest L2 safe head.
     l2_safe_head: L2BlockInfo,
+    /// A receiver that tells derivation to begin.
+    sync_complete_rx: UnboundedReceiver<()>,
```
clabby, line 34: Can this be a `oneshot`?

refcell: No because in order to not block on the engine sync check, we spawn the check in a thread which will send over the cloned channel. With a oneshot channel, we can't clone the sender since there would only be one sender allowed.

PR #1643, opened 2025-05-04
chore(interop): Refactor `MessageGraph` test utilities

clabby:
## Overview

Refactors the `MessageGraph` test utilities in `kona-interop` to be a bit more intuitive.

Note: The errors that caused block invalidity are still opaque to the consumer, and a follow up PR (#1644) exposes them in the API such that the consumer can inspect why a block was deemed to be invalid.


emhane (approved):
left some comments, but in general lgtm

PR #1644, opened 2025-05-04
feat(interop): Bubble up message validity errors

clabby:
## Overview

Refactors the API of the `MessageGraphError` to return the reason for a block's invalidation. Also adjusts the `MessageGraph` tests to assert the specific reasons for a graph resolution failure on the mock chain.

### DIFF
--- a/crates/protocol/interop/src/graph.rs
+++ b/crates/protocol/interop/src/graph.rs
@@ -103,13 +103,9 @@ where
             if let Err(e) = self.check_single_dependency(message).await {
                 warn!(
                     target: "message_graph",
-                    executing_chain_id =  message.executing_chain_id,
+                    executing_chain_id = message.executing_chain_id,
                     message_hash = ?message.inner.payloadHash,
-                    warn!(
-                    target: "message_graph",
-                    executing_chain_id =  message.executing_chain_id,
-                    message_hash = ?message.inner.payloadHash,
-                    err = e,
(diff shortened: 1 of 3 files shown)


clabby:
## 📚 PR Stack

- https://github.com/op-rs/kona/pull/1645
- https://github.com/op-rs/kona/pull/1644
- `main`

crates/protocol/interop/src/graph.rs
```diff
@@ -87,57 +87,27 @@ where
...
+        // message. Errors are de-duplicated by chain ID in a map, since a single invalid
+        // message is cause for invalidating a block.
+        for message in self.messages.iter() {
+            if let Err(e) = self.check_single_dependency(message).await {
                 warn!(
                     target: "message_graph",
                     executing_chain_id =  message.executing_chain_id,
                     message_hash = ?message.inner.payloadHash,
-                    "Invalid ExecutingMessage found",
+                    "Invalid ExecutingMessage found: {e}",
```
theochap, line 108: nit: instead of formatting inside the warn string we could put the error as a field like 
```suggestion
                    warn!(
                    target: "message_graph",
                    executing_chain_id =  message.executing_chain_id,
                    message_hash = ?message.inner.payloadHash,
                    err = e,
                    "Invalid ExecutingMessage found",
```

theochap (approved):
Looks great

PR #1647, opened 2025-05-04
feat(genesis): Add `block_number_from_timestamp` helper

clabby:
## Overview

Adds a helper function to the `RollupConfig` to compute a block number from a given timestamp, relative to the config's L2 genesis time and block time.


emhane:
a nice thing is if the base pr is linked in the pr description as
> Based on <pr-link>

 makes reviewing easier

theochap (approved):

Issue #1651, opened 2025-05-05
[TRACKING] chore(kona): code cleanup/ergonomics


theochap:
## Description
This issue tracks a list of potential improvements to the codebase that will improve code quality/maintainability/ergonomics. The items on this list are low-priority and should be addressed on a best effort basis. It can be a great way to get introduced to the codebase if you're an external contributor.

## Tracker

- [x] Remove the `no_std` requirement from the `kona-rpc` crate. #1659 
- [ ] Change the type builders (`NetworkBuilder`, ...) to become infallible by specifying mandatory configuration at initialization
- [ ] Split P2p RPC types in separate files (https://github.com/op-rs/kona/pull/1633#discussion_r2071923307)

clabby:
2 more:
1. https://github.com/op-rs/kona/pull/1816#discussion_r2098407831

Issue #1634, opened 2025-05-02
[TRACKING] testing(node): bootstrap testing for `kona-node` using the `devstack`


theochap:
## Description
This issue tracks the efforts to bootstrap testing for the `kona-node` using the monorepo's `devstack` (https://github.com/ethereum-optimism/optimism/tree/develop/devnet-sdk). We'll start with a kurtosis testing integration (can be done without updating `ethereum-optimism`) and follow-up with in-memory (will require to update the monorepo)

## Tracker
- #1633 Refactor and clean-up `kona-p2p` RPC in preparation of the `node-rpc` integration
- Add `rollup-client` RPC endpoints to `kona-node` (see https://github.com/ethereum-optimism/optimism/blob/174e55f0a1e73b49b80a561fd3fedd4fea5770c6/op-service/sources/rollupclient.go)
- Bootstrap `kurtosis`-based tests in `kona-node` with the `devstack`
- Update the monorepo's `devstack` to support launching `in-memory` `kona-nodes` for testing
- Bootstrap `in-memory`-based tests in `kona-node` with the `devstack`

PR #1652, opened 2025-05-05
feat(node/net): skeleton impl of the rollup RPC api

theochap:
## Description
This PR provides a skeleton implementation of the `rollup-client` endpoints for kona. This PR:
- takes the network builder outside of `kona-p2p` into a `kona-net` crate.
- adds a new `rollup` router to `kona-rpc`
- plugs the rollup router into `kona-net`

For now the routes of the rollup router are empty. I will populate them in the next PR and connect the router to the engine driver.

Progress towards #1634 and #1651


crates/node/net/src/driver.rs
```diff
@@ -146,13 +154,21 @@ impl Network {
...
                             continue;
                         };
                         req.handle(&self.gossip, &handler);
                     },
+                    req = rollup_rpc.recv() => {
+                        let Some(req) = req else {
+                            trace!("Receiver `None` rpc request");
+                            continue;
+                        };
+                        // TODO(@theochap): add engine driver here.
```
refcell, line 169: How is the engine driver going to slot in here? The `kona-node-service` crate is what orchestrates all the different actors (and their drivers). If we pull the engine driver into the `NetworkDriver` here, how do we access it from inside `kona-node-service`?

PR #1655, opened 2025-05-06
feat(proof): Derivation over generic L1/L2/DA providers

hashcashier:
This PR refactors the derivation pipeline in `kona-proof` to accept generic `ChainProvider`, `L2ChainProvider` and `DataAvailabilityProvider` types. This allows finer-grain customization of the derivation pipeline with less code duplication. 

For example, the following projects would need only plug in their custom providers instead of vendoring in the entire pipeline code just to change the `OracleDerivationPipeline`/`OracleDataProvider` type aliases:
* [Kailua](https://github.com/risc0/kailua/blob/15a2a8f8135eb25d055962ecde78ba3197844c99/crates/common/src/kona/pipeline.rs#L15)
* [Hokulea](https://github.com/Layr-Labs/hokulea/blob/4f8b33af384ef19d630eaf0f24885f11d3ea15d5/crates/proof/src/pipeline.rs#L47)
* [Hana](https://github.com/celestiaorg/hana/blob/50efa651802cdc601934ebb00e4af2865a204da9/crates/oracle/src/pipeline.rs#L32)

Decreasing the amount of vendoring required should also serve to keep these projects up to date with the latest changes in the standard Kona pipeline.

### DIFF
--- a/bin/client/src/interop/transition.rs
+++ b/bin/client/src/interop/transition.rs
@@ -19,7 +19,7 @@ use kona_proof::{
     executor::KonaExecutor,
     l1::{OracleBlobProvider, OracleL1ChainProvider, OraclePipeline},
     l2::OracleL2ChainProvider,
-    sync::new_pipeline_cursor,
+    sync::new_oracle_pipeline_cursor,
 };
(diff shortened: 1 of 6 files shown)


clabby (approved):
Looks good. Thanks Rami!

Issue #1616, opened 2025-05-01
feat(node/engine): Record Unsafe Block Height


refcell:
### Description

Add a [gauge metric](https://docs.rs/metrics/0.24.2/metrics/struct.Gauge.html) to the `kona-engine` crate that tracks the unsafe block height. The unsafe block height is updated when an `InsertUnsafeTask` is executed on the engine state. Specifically, the unsafe block height is set in the task [here](https://github.com/op-rs/kona/blob/main/crates/node/engine/src/task_queue/tasks/insert/task.rs#L206-L207). Since mutation to the `EngineState` is encapsulated in methods, the metric can be updated when the [`set_unsafe_head`](https://github.com/op-rs/kona/blob/main/crates/node/engine/src/state/core.rs#L107-L111) method is called.

### Output

Use `kona-macros` to set the block height on a new gauge metric when the [`set_unsafe_head`](https://github.com/op-rs/kona/blob/main/crates/node/engine/src/state/core.rs#L107-L111) method is called. Set the height as the [`L2BlockInfo.block_info.number`](https://github.com/op-rs/kona/blob/main/crates/protocol/protocol/src/block.rs#L69).

Unit test the metric output. The `metrics-exporter-prometheus` crate has [unit tests](https://github.com/metrics-rs/metrics/blob/main/metrics-exporter-prometheus/src/exporter/builder.rs#L569-L591) that should be used as a resource.

clabby:
Closed by #1741

PR #1656, opened 2025-05-06
feat(node/engine): record unsafe block height metric

7suyash7:
#1616 

Adds a Prometheus gauge metric named `engine_unsafe_head_height`. Tracks the block number of the L2 unsafe head and is updated whenever the `EngineState::set_unsafe_head` method is called.

The metrics implementation is gated behind the `metrics` feature flag. Includes a basic unit test to verify the metric's behaviour.


crates/node/engine/src/lib.rs
```diff
@@ -6,6 +6,10 @@
 )]
 #![cfg_attr(docsrs, feature(doc_cfg, doc_auto_cfg))]
 
+#[cfg(feature = "metrics")]
+#[macro_use]
+extern crate kona_macros;
```
refcell, line 11: Nice!

7suyash7: Hi, if this looks good, can I pick up the other metrics that need to be added here?

refcell:
@clabby implemented this in https://github.com/op-rs/kona/pull/1741

Issue #1653, opened 2025-05-05
Add `rollup-client` RPC endpoints to `kona-node`


theochap:
## Description
To be able to integrate `kona-node` inside the `devnet-sdk` we need to expose the `rollup-client` RPC endpoints from `op-node` (see https://github.com/ethereum-optimism/optimism/blob/174e55f0a1e73b49b80a561fd3fedd4fea5770c6/op-service/sources/rollupclient.go)

## Tracker
- [x] #1659 
- [ ] #1692
- [ ] #1664
- [ ] #1697

PR #1664, opened 2025-05-06
feat(engine/rpc): skeleton implementation of the engine rpc

theochap:
## Description
This PR provides a skeleton implementation of the engine state queries and rollup RPC. In particular, it:
- Adds a query method to the engine actor to gather the engine state
- Adds an empty implementation for a rollup RPC query handler

## Note
Progress towards #1653


crates/node/engine/src/rpc.rs
```diff
@@ -0,0 +1,92 @@
...
+    pub output_root: B256,
+    /// A reference to the L2 block.
+    pub block_ref: L2BlockInfo,
+    /// The withdrawal storage root.
+    pub withdrawal_storage_root: B256,
+    /// The state root.
+    pub state_root: B256,
+    /// The status of the node sync.
+    pub sync_status: SyncStatus,
+}
```
refcell, line 30: I think it'd be cleaner if we move these types into their own module to avoid this file from growing. Not a blocker though

theochap: Moved back these types to their original files

crates/node/service/src/actors/engine.rs
```diff
@@ -136,6 +149,36 @@ impl NodeActor for EngineActor {
...
+                                } else {
+                                    warn!(target: "engine", "Engine query channel closed unexpectedly, shutting down query receiver task.");
+                                    return;
+                                }
+                            }
+
+                    }
+                }
+            });
+        }
```
refcell, line 180: Good candidate for a separate function to spawn the query server. Not a blocker but worth a ticket to follow up + cleanup

crates/node/service/src/service/validator.rs
```diff
@@ -153,11 +155,15 @@ pub trait ValidatorNodeService {
...
 
+        let rollup_rpc = RollupRpc::new(engine_query_sender);
+
         // The RPC Server should go last to let other actors register their rpc modules.
         let rpc = if let Some(mut rpc) = self.rpc() {
             if let Some(p2p_module) = p2p_module {
                 rpc = rpc.merge(p2p_module.into_rpc()).expect("failed to merge p2p rpc module");
             }
+
+            rpc = rpc.merge(rollup_rpc.into_rpc()).expect("failed to merge engine rpc module");
```
refcell, line 166: These `.expect` will need to be removed. We should ticket + follow up with proper error bubbling

refcell (approved):

crates/node/service/src/actors/engine.rs
theochap: Good suggestion will do it in this PR.

crates/node/service/src/service/validator.rs
theochap: #1704

crates/node/service/src/actors/engine.rs
theochap: Done! I refactored that method, which did simplify quite a bit the main `start` method of the engine actor

PR #1666, opened 2025-05-06
fix(node/engine): Attributes Tx Mismatch

refcell:
### Description

Fixes the attributes mismatch checks for deposit transactions where the l2 block returned by the engine api can contain mismatching deposit txs.

Since a `mint` value of `Some(0)` is functionally equivalent to `None`, the `kona-node` needs to account for this mismatch between deposit transactions.

Instead of modifying the `OpAttributesWithParent`, this PR just performs reconciliation for now.


refcell:
## 📚 $\text{Stack Overview}$

* #1666 👈
* `main`

clabby (commented):
Rather than adjusting the `mint` value out of band, maybe we just update `kona-derive`?

Updating this should do the trick, if I'm not mistaken:
https://github.com/op-rs/kona/blob/0324411840e948369c9464285efe7a1d94eeff3d/crates/protocol/protocol/src/deposits.rs#L204-L209

refcell:
> Rather than adjusting the `mint` value out of band, maybe we just update `kona-derive`?
> 
> Updating this should do the trick, if I'm not mistaken:
> 
> https://github.com/op-rs/kona/blob/0324411840e948369c9464285efe7a1d94eeff3d/crates/protocol/protocol/src/deposits.rs#L204-L209

Either way works - Can go with changing `kona-derive`.

clabby (approved):
g2g - let's get back to removing this ambiguity once `mint` isn't an option.

PR #1667, opened 2025-05-06
fix(node/engine): Round Robin Task Execution

refcell:
### Description

Allows tasks to be processed in a round-robin fashion based on their type.

Since consolidation tasks may need to wait for data in the derivation pipeline, we still want the engine to execute unsafe payload tasks in the meantime.

This round-robin type-based policy preserves strict ordering of the different task types, while allowing progress to be made on different tasks.

### DIFF
--- a/crates/node/engine/src/task_queue/core.rs
+++ b/crates/node/engine/src/task_queue/core.rs
@@ -1,8 +1,8 @@
 //! The [Engine] is a task queue that receives and executes [EngineTask]s.
 
 use super::{EngineTaskError, EngineTaskExt};
-use crate::{EngineState, EngineTask};
-use std::collections::VecDeque;
+use crate::{EngineState, EngineTask, EngineTaskType};
+use std::collections::{HashMap, VecDeque};
 
 /// The [Engine] task queue.
 ///
@@ -18,7 +18,9 @@ pub struct Engine {
     /// The state of the engine.
     state: EngineState,
     /// The task queue.
-    tasks: VecDeque<EngineTask>,
+    tasks: HashMap<EngineTaskType, VecDeque<EngineTask>>,
+    /// The current task being executed.
+    cursor: EngineTaskType,
 }
 
 impl Engine {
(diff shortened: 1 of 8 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1667 👈
* #1666
* `main`

crates/node/engine/src/task_queue/tasks/task.rs
```diff
@@ -34,6 +62,16 @@ impl EngineTask {
...
     }
+
+    /// Returns the [`EngineTaskType`] for the [`EngineTask`].
+    pub const fn ty(&self) -> EngineTaskType {
+        match &self {
+            Self::ForkchoiceUpdate(_) => EngineTaskType::ForkchoiceUpdate,
+            Self::InsertUnsafe(_) => EngineTaskType::InsertUnsafe,
+            Self::BuildBlock(_) => EngineTaskType::BuildBlock,
+            Self::Consolidate(_) => EngineTaskType::Consolidate,
+        }
```
theochap, line 73: Nit: to automatically build that type and all the associated impls we can use `strum`. Can also understand if we want to avoid adding external dependencies if not absolutely required

crates/node/engine/src/task_queue/core.rs
```diff
@@ -46,20 +52,39 @@ impl Engine {
...
                 e => return e,
             }
+            let ty = task.ty();
+            if let Some(queue) = self.tasks.get_mut(&ty) {
+                queue.pop_front();
+                if queue.is_empty() {
+                    self.tasks.remove(&ty);
+                }
+            };
         }
```
theochap, line 88: Is there an easy way to test that specific piece of code in isolation? I think this will help make sure this behaves properly and we do get the round robin behavior

theochap (commented):
Looks great, would be nice to have a couple of tests for that

crates/node/engine/src/task_queue/tasks/task.rs
refcell: Yea, let's leave the optimization for future PRs. Important to just get this working nicely

crates/node/engine/src/task_queue/core.rs
refcell: This is a good candidate for a refactor to encapsulate the queue behavior. For now, this is working when running live against `op-reth` so I'd suggest we ticket out a refactor for later.

theochap (approved):
Looks great

PR #1670, opened 2025-05-07
Add permissions sanity check TIP to RELEASES.md

lewejahlil:
Adds reminder to verify permissions on crates affected in release


clabby (approved):
legend

PR #1674, opened 2025-05-07
fix(node/service): Mark Engine Ready on Sender

refcell:
### Description

Previously, the engine ready signal when sync is complete was continuously being sent because there was no way to tell if it had been received.

This PR closes the channel. When the channel is closed by the receiver after it has been marked `engine_ready`, no more engine ready signals are sent.


refcell:
## 📚 $\text{Stack Overview}$

* #1674 👈
* `main`

clabby (approved):

PR #1675, opened 2025-05-07
feat(node/service): Flush Channel on Invalid Payloads

### LINKED ISSUE #1239: feat(node): Channel flushing on invalid blocks
## Overview

When the engine gets an `INVALID` response for a new block during block building, it is reduced to deposits only and the channel + remaining buffered batches are flushed. This will need to be covered through a response to the derivation actor, after it has sent safe attributes off for processing (or in the sequencer, which isn't implemented yet.)

Spec: https://specs.optimism.io/protocol/derivation.html#l1-sync-payload-attributes-processing

refcell:
### Description

Wires up the engine actor to send a signal to the derivation actor.

This is necessary when the engine api marks a payload `INVALID`.
Post-Holocene, the payload becomes deposit-only and is re-executed.
At the same time, the derivation pipeline needs to be flushed, so
the channel and remaining buffered batches are cleared.

Makes progress on #1239

### DIFF
--- a/crates/node/service/src/actors/derivation.rs
+++ b/crates/node/service/src/actors/derivation.rs
@@ -5,7 +5,7 @@ use async_trait::async_trait;
 use kona_derive::{
     errors::{PipelineError, PipelineErrorKind, ResetError},
     traits::{Pipeline, SignalReceiver},
-    types::{ActivationSignal, ResetSignal, StepResult},
+    types::{ActivationSignal, ResetSignal, Signal, StepResult},
 };
 use kona_protocol::{BlockInfo, L2BlockInfo, OpAttributesWithParent};
 use thiserror::Error;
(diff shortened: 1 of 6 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1675 👈
* `main`

crates/node/service/src/actors/derivation.rs
```diff
@@ -202,6 +233,18 @@ where
...
                     self.process(InboundDerivationMessage::NewDataAvailable).await?;
                 }
+                signal = self.derivation_signal_rx.recv() => {
+                    let Some(signal) = signal else {
+                        error!(
+                            target: "derivation",
+                            ?signal,
+                            "DerivationActor failed to receive signal"
+                        );
+                        return Ok(());
```
theochap, line 243: Should we return an error in that case (or the one above)?

refcell: I think you're right we should error here since it's critical to consensus

theochap (approved):

PR #1676, opened 2025-05-07
feat(node/service): Re-import Deposits Only Payload

### LINKED ISSUE #1239: feat(node): Channel flushing on invalid blocks

refcell:
### Description

Re-imports deposits-only payload on invalid payload post-holocene.

Closes #1239

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -118,14 +118,14 @@ dependencies = [
 
 [[package]]
 name = "alloy-consensus"
-version = "0.15.9"
+version = "0.15.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1676 👈
* #1675
* `main`

theochap (approved):

PR #1677, opened 2025-05-07
feat(node/service): Update L2 Safe Head

### LINKED ISSUE #1240: feat(node): Wait for L2 safe head updates in `DerivationActor`
## Overview

The derivation actor currently derives safe payload attributes, and sends them off to the engine actor to either be built or checked for equivalence with existing blocks [via L1 consolidation](https://specs.optimism.io/protocol/derivation.html#l1-consolidation-payload-attributes-matching) (depending on if `safe_head < unsafe_head`.)

When attributes are sent off, the pipeline must wait to continue processing payloads until the L2 safe head has advanced. This synchronization step should happen here:

https://github.com/op-rs/kona/blob/3f70082d241a0b2cd68d45ef9cbaf708175782f5/crates/node/service/src/actors/derivation.rs#L201

refcell:
### Description

Only produce payload attributes when the l2 safe head is updated.

Closes #1240.

### DIFF
--- a/crates/node/service/src/actors/derivation.rs
+++ b/crates/node/service/src/actors/derivation.rs
@@ -225,6 +225,7 @@ where
         if *self.engine_l2_safe_head.borrow() == self.l2_safe_head {
             return Ok(());
         }
+        self.l2_safe_head = *self.engine_l2_safe_head.borrow();
 
         // Advance the pipeline as much as possible, new data may be available or there still may be
         // payloads in the attributes queue.
--- a/crates/node/service/src/actors/derivation.rs
+++ b/crates/node/service/src/actors/derivation.rs
@@ -225,7 +225,6 @@ where
         if *self.engine_l2_safe_head.borrow() == self.l2_safe_head {
             return Ok(());
         }
-        self.l2_safe_head = *self.engine_l2_safe_head.borrow();
 
         // Advance the pipeline as much as possible, new data may be available or there still may be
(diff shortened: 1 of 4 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1677 👈
* `main`

clabby (approved):

crates/node/service/src/actors/derivation.rs
```diff
@@ -214,6 +221,11 @@ where
             return Ok(());
         }
 
+        // The L2 Safe Head must be advanced before producing new payload attributes.
+        if *self.engine_l2_safe_head.borrow() == self.l2_safe_head {
+            return Ok(());
+        }
```
theochap, line 227: Question: Is there a risk of race conditions where we send to the `sync_complete_rx` but not yet to the `engine_l2_safe_head` channel which causes the `process` method to complete but without actually processing the next payload? If we're relying on some assumption maybe we should mention it there. If there is a risk, would using `.changed` instead of `.borrow` prevent race conditions?

refcell: This shouldn't be an issue because it will just get re-processed and executed the next time around `process` is called.

refcell: To expand - once the sync complete rx is sent, it will keep trying to call `process` frequently to make progress.

PR #1686, opened 2025-05-08
feat(executor): Public `compute_receipts_root` method

hashcashier:
This PR refactors `compute_receipts_root` as its own publicly exposed function outside of the `StatelessL2Builder` implementation to enable unvendored reuse by downstream packages such as [in Kailua](https://github.com/risc0/kailua/blob/15a2a8f8135eb25d055962ecde78ba3197844c99/crates/common/src/client/stitching.rs#L178).

### DIFF
--- a/crates/proof/executor/src/builder/assemble.rs
+++ b/crates/proof/executor/src/builder/assemble.rs
@@ -46,8 +46,7 @@ where
             |tx, buf| buf.put_slice(tx.as_ref()),
         )
         .root();
-        let receipts_root =
-            Self::compute_receipts_root(&ex_result.receipts, self.config, timestamp);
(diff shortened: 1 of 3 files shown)


clabby (changes requested):
Glad to expose this, but would prefer this to be an associated function on `StatelessL2Builder`

hashcashier:
@clabby `StatelessL2Builder` has several type parameters that are unrelated to the function and would have to be specified just to call it.

clabby:
> @clabby `StatelessL2Builder` has several type parameters that are unrelated to the function and would have to be specified just to call it.

Ah good point. This is fine then; Not a huge deal!

clabby (approved):

PR #1691, opened 2025-05-08
fix(node/service): EL Sync Only

refcell:
### Description

Makes `kona-node` slightly more opinionated by only supporting EL sync.

Previously, the node broke since it couldn't kick off EL sync post-finalization.
Since the finalized and safe heads were populated using the execution engine, a
forkchoice state that kicks off EL sync (zero hash in finalized) wasn't constructed.

This PR removes the unused CL sync, deprecating the `--l2.enginekind` flag, and
moves the `kona-node` over to only support EL sync by initializing finalized
and safe heads to zero on startup, like the `op-node` does. This allows the first
unsafe insert task to kick off el sync.

### DIFF
--- a/bin/node/src/commands/node.rs
+++ b/bin/node/src/commands/node.rs
@@ -3,7 +3,7 @@
 use alloy_rpc_types_engine::JwtSecret;
 use anyhow::{Result, bail};
 use clap::Parser;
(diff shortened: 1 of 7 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1691 👈
* `main`

PR #1700, opened 2025-05-09
feat(supervisor): boilerplate

### LINKED ISSUE #1671: feat(supervisor): boilerplate setup
# Description

This issue tracks all tasks related to scaffolding and setting up the Rust op-supervisor service. The goal is to establish the foundational structure, configuration, and development environment to enable further development of core modules.

# Tasks

- [ ] Create base project structure with Cargo workspaces
- [ ] Add cli, flags, and main modules
- [ ] Implement argument parsing using clap with required flags
- [ ] Setup logging and basic observability (tracing + Prometheus metrics integration)

dhyaniarun1993:
Closes #1671

### DIFF
--- a/bin/supervisor/src/cli.rs
+++ b/bin/supervisor/src/cli.rs
@@ -1,0 +1,74 @@
+//! Contains the supervisor CLI.
+
+use crate::flags::{GlobalArgs, MetricsArgs, SupervisorArgs};
+use clap::Parser;
+
+use anyhow::Result;
+use kona_cli::cli_styles;
+use kona_supervisor_service::{Service, Config};
+use std::net::SocketAddr;
+use tracing::info;
+
+/// CLI for the Rust implementation of the OP Supervisor.
+#[derive(Parser, Debug)]
+#[command(name = "op-supervisor", about = "Rust implementation of the OP Supervisor", styles = cli_styles())]
+pub struct Cli {
+    /// Global args
+    #[command(flatten)]
+    pub global: GlobalArgs,
+
+    /// Prometheus metrics args
+    #[command(flatten)]
+    pub metrics: MetricsArgs,
+
+    /// Supervisor args
+    #[command(flatten)]
+    pub supervisor: SupervisorArgs,
+}
+
+impl Cli {
+    /// Runs the CLI.
+    pub fn run(self) -> Result<()> {
+        self.init_telemetry(&self.global, &self.metrics)?;
+
+        Self::run_until_ctrl_c(async move {
+            let config = Config {
+                rpc_addr: SocketAddr::new(self.supervisor.rpc_address, self.supervisor.rpc_port),
+            };
+            let mut service = Service::new(config.clone())?;
+            service.run().await?; // run() now returns Result<()> and populates the handle internally
+
+            tokio::signal::ctrl_c().await?;
+            info!("Shutdown signal received. Initiating service shutdown...");
+    
+            service.shutdown().await?; // Call shutdown on the service instance itself
+    
+            info!("Supervisor service shut down gracefully.");
+            Ok(())
+        })
+    }
+
+    /// Run until ctrl-c is pressed.
+    pub fn run_until_ctrl_c<F>(fut: F) -> Result<()>
+    where
+        F: std::future::Future<Output = Result<()>>,
+    {
+        let rt = Self::tokio_runtime().map_err(|e| anyhow::anyhow!(e))?;
+        rt.block_on(fut)
+    }
+
+    /// Creates a new default tokio multi-thread [Runtime](tokio::runtime::Runtime) with all
+    /// features enabled
+    pub fn tokio_runtime() -> Result<tokio::runtime::Runtime, std::io::Error> {
+        tokio::runtime::Builder::new_multi_thread().enable_all().build()
+    }
+
+    /// Initializes the telemetry stack and Prometheus metrics recorder.
+    pub fn init_telemetry(&self, args: &GlobalArgs, metrics: &MetricsArgs) -> anyhow::Result<()> {
+        // Filter out discovery warnings since they're very very noisy.
+        let filter = tracing_subscriber::EnvFilter::from_default_env();
+
+        args.init_tracing(Some(filter))?;
+        metrics.init_metrics()
+    }
+}
--- a/bin/supervisor/src/flags/globals.rs
+++ b/bin/supervisor/src/flags/globals.rs
@@ -1,0 +1,58 @@
+use clap::{ ArgAction, Args};
+use kona_cli::init_tracing_subscriber;
+use tracing_subscriber::EnvFilter;
+
+/// Global configuration arguments.
+#[derive(Args, Debug)]
+pub struct GlobalArgs {
+    /// Verbosity level (0-5).
+    /// If set to 0, no logs are printed.
+    /// By default, the verbosity level is set to 3 (info level).
+    #[arg(
+        long, 
+        short, 
+        global = true, 
+        default_value = "3", 
+        action = ArgAction::Count,
+    )]
+    pub v: u8,
+}
+
+impl GlobalArgs {
+    /// Initializes the telemetry stack.
+    pub fn init_tracing(&self, filter: Option<EnvFilter>) -> anyhow::Result<()> {
+        Ok(init_tracing_subscriber(self.v, filter)?)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use clap::Parser;
+
+    // Helper struct to parse GlobalArgs within a test CLI structure
+    #[derive(Parser, Debug)]
+    struct TestCli {
+        #[command(flatten)]
+        global: GlobalArgs,
+    }
+
+    #[test]
+    fn test_default_verbosity_level() {
+        let cli = TestCli::parse_from(["test_app"]);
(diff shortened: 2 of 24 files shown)


bin/supervisor/src/flags/supervisor.rs
```diff
@@ -0,0 +1,185 @@
+use clap::Args;
+use std::net::IpAddr;
+
+/// Supervisor configuration arguments.
+#[derive(Args, Debug)]
+pub struct SupervisorArgs {
+    /// L1 RPC source.
+    #[arg(long, env = "L1_RPC", help = "L1 RPC source")]
```
emhane, line 8: there is some way you can get clap to use the rustdocs as help, not sure if that is what happens by default. I know if you have some escape chars in there you should use `verbatim_doc_comment` attribute to get it to print unescaped as help

feel free to leave the doc without trailing punctuation for sake of simplicity if that's why the help is now double up. is safer that way since means we don't have to maintain the invariant docs = help message manually.

bin/supervisor/src/flags/supervisor.rs
```diff
@@ -0,0 +1,185 @@
...
+    /// Path to the dependency-set JSON config file.
+    #[arg(long = "dependency-set", env = "DEPENDENCY_SET", help = "Path to dependency-set config")]
+    pub dependency_set: Option<String>,
+
+    /// IP address for the Supervisor RPC server to listen on.
+    #[arg(
+        long = "rpc.addr",
+        env = "RPC_ADDR",
+        help = "Supervisor RPC listening address",
+        default_value = "127.0.0.1"
```
emhane, line 50: probably want to make this all interfaces instead, no?

bin/supervisor/src/flags/supervisor.rs
```diff
@@ -0,0 +1,185 @@
...
+    )]
+    pub rpc_address: IpAddr,
+
+    /// Port for the Supervisor RPC server to listen on.
+    #[arg(
+        long = "rpc.port",
+        env = "RPC_PORT",
+        help = "Supervisor RPC listening port",
+        default_value_t = 9000
```
emhane, line 59: is this the port Go uses? this is the l1 cL p2p default port...

emhane (changes requested):
first pass, some qs and some requested changes

bin/supervisor/src/flags/supervisor.rs
dhyaniarun1993: Ohh yea, my bad. Thanks for pointing it out.
Will update.

bin/supervisor/src/flags/supervisor.rs
dhyaniarun1993: Not sure what which port go uses. Checking...
Yea, l1 cl p2p default port is 9000. Will change it to go default

bin/supervisor/src/cli.rs
```diff
@@ -0,0 +1,74 @@
...
+    /// Run until ctrl-c is pressed.
+    pub fn run_until_ctrl_c<F>(fut: F) -> Result<()>
+    where
+        F: std::future::Future<Output = Result<()>>,
+    {
+        let rt = Self::tokio_runtime().map_err(|e| anyhow::anyhow!(e))?;
+        rt.block_on(fut)
+    }
+
+    /// Creates a new default tokio multi-thread [Runtime](tokio::runtime::Runtime) with all
```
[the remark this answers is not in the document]
dhyaniarun1993: Updated

bin/supervisor/src/flags/supervisor.rs
dhyaniarun1993: Updated.

bin/supervisor/src/flags/supervisor.rs
dhyaniarun1993: Updated to the go-supervisor default port.

clabby (approved):
Looks good to start out w/ :)

We'll probably also need a `kona-supervisor-db` crate; The supervisor DB itself is deserving of isolation.

It looks like we're heading in this direction already, but to emphasize, it'd be grand if the service as a whole has a programatic entrypoint via `kona-supervisor-core`. Will give us more flexibility as to using this as an execution extension to reth at some point.

crates/supervisor/core/src/rpc/metrics.rs
```diff
@@ -0,0 +1,111 @@
...
+/// # Usage
+/// ```ignore
+/// async fn my_rpc_method(&self, arg: u32) -> RpcResult<String> {
+///     observe_rpc_call!("my_rpc_method_name", {
+///         // ... actual RPC logic ...
+///         if arg == 0 { Ok("success".to_string()) } else { Err(ErrorObject::owned(1, "failure", None::<()>)) }
+///     })
+/// }
+/// ```
+#[macro_export]
```
emhane, line 95: nice docs with example on macro, should we open an issue to remove the ignore attribute when the rpc logic has been written?

crates/supervisor/core/src/rpc/server.rs
```diff
@@ -0,0 +1,95 @@
...
+use jsonrpsee::{
+    core::RpcResult,
+    types::{ErrorObject, error::ErrorCode},
+};
+use kona_interop::{DerivedIdPair, ExecutingDescriptor, SafetyLevel, SuperRootResponse};
+use kona_supervisor_rpc::SupervisorApiServer;
+use std::sync::Arc;
+use tracing::{trace, warn};
+
+/// The server-side implementation struct for the `SupervisorApi`.
```
emhane, line 16:
```suggestion
/// The server-side implementation struct for the [`SupervisorApiServer`].
```
let's link docs wherever we can

crates/supervisor/rpc/src/jsonrpsee.rs
emhane, line 1: this trait exists, let's add the methods there instead - doesn't add much overhead or modify anything existing
https://github.com/op-rs/kona/blob/e9b7d9cfd9092e3f37c60fa05fa7c3384cada3c8/crates/node/rpc/src/jsonrpsee.rs#L148-L160

crates/supervisor/service/src/service.rs
```diff
@@ -0,0 +1,75 @@
...
+    pub rpc_addr: SocketAddr,
+    // Add other configuration fields as needed (e.g., connection details for L1/L2 nodes)
+}
+
+/// The main service structure for the Kona Supervisor.
+/// orchestrates the various components of the supervisor
+#[derive(Debug)]
+pub struct Service {
+    config: Config,
+    _supervisor: Arc<dyn kona_supervisor_core::SupervisorService + Send + Sync>,
```
emhane, line 24:
```suggestion
    _supervisor: Arc<dyn SupervisorService + Send + Sync>,
```
no reason to use long path here

crates/supervisor/service/src/service.rs
```diff
@@ -0,0 +1,75 @@
...
+            "Supervisor RPC server started successfully and listening on {}",
+            self.config.rpc_addr
+        );
+
+        Ok(())
+    }
+
+    pub async fn shutdown(mut self) -> Result<()> {
+        if let Some(handle) = self.rpc_server_handle.take() {
+            info!("Sending stop signal to RPC server...");
```
emhane, line 64:
```suggestion
            info!(target: "supervisor::service", "Sending stop signal to RPC server...");
```
let's add a simplified target here for log filtering with `RUST_LOG=supervisor::service=trace`

crates/supervisor/core/src/rpc/server.rs
```diff
@@ -0,0 +1,95 @@
...
+/// It holds a reference to the core Supervisor logic.
+#[derive(Debug, Clone)]
+pub struct SupervisorRpc {
+    /// Reference to the core Supervisor logic.
+    /// Using Arc allows sharing the Supervisor instance if needed,
+    supervisor: Arc<dyn SupervisorService + Send + Sync + 'static>,
```
emhane, line 22: no need to dynamic dispatch here, let's use and unbounded a generic with a default instead and trait bound the generic on the impl bodies which need it

emhane (changes requested):
nice! let's dedup the supervisor rpc api trait right away in this pr, and only do the impl of the `kona_rpc::jsonrpsee::SupervisorApiServer` in this new `kona-supervisor` crate.

crates/supervisor/rpc/src/jsonrpsee.rs
emhane: makes more sense to keep this trait and remove the original one, will open issue

emhane (approved):

crates/supervisor/rpc/src/jsonrpsee.rs
dhyaniarun1993: Yea, I think it's better to keep all supervisor related code inside `supervisor` and node can depend on this.

emhane: https://github.com/op-rs/kona/issues/1731

bin/supervisor/src/flags/globals.rs
```diff
@@ -0,0 +1,57 @@
...
+    /// By default, the verbosity level is set to 3 (info level).
+    #[arg(
+        long,
+        short,
+        global = true,
+        default_value = "3",
+        action = ArgAction::Count,
+    )]
+    pub v: u8,
+}
```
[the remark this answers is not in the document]
emhane: https://github.com/op-rs/kona/issues/1745

PR #1702, opened 2025-05-09
fix(node/engine): Pre-Holocene Deposits Only

### LINKED ISSUE #1699: fix(node/engine): Pre-Holocene Deposits Only
### Description

Currently the `kona-node` logic [re-attempts a deposits-only payload](https://github.com/op-rs/kona/blob/main/crates/node/engine/src/task_queue/tasks/build/task.rs#L248-L263) regardless of whether the payload is post-holocene hardfork.

This PR is to fix this logic so the task is _only_ re-executed when the payload is post-Holocene.

refcell:
### Description

Fixes the engine to only re-attempt deposits only payload if holocene is active for the given payload timestamp.

Closes #1699

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,2 @@
 * @clabby @refcell @emhane @theochap
+bin/supervisor @dhyaniarun1993 @itschaindev @sadiq1971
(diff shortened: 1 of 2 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1702 👈
* `main`

PR #1708, opened 2025-05-09
fix(node/engine): Derivation Sync Start

refcell:
### Description

Fixes derivation to use the engine safe head instead of keeping its own local view.

### DIFF
--- a/crates/node/service/src/actors/derivation.rs
+++ b/crates/node/service/src/actors/derivation.rs
@@ -30,8 +30,6 @@ where
 {
     /// The derivation pipeline.
     pipeline: P,
-    /// The latest L2 safe head.
-    l2_safe_head: L2BlockInfo,
     /// The l2 safe head from the engine.
     engine_l2_safe_head: WatchReceiver<L2BlockInfo>,
     /// A receiver that tells derivation to begin.
@@ -74,7 +72,6 @@ where
     #[allow(clippy::too_many_arguments)]
     pub const fn new(
         pipeline: P,
-        l2_safe_head: L2BlockInfo,
         engine_l2_safe_head: WatchReceiver<L2BlockInfo>,
         sync_complete_rx: UnboundedReceiver<()>,
         derivation_signal_rx: UnboundedReceiver<Signal>,
@@ -84,7 +81,6 @@ where
     ) -> Self {
         Self {
             pipeline,
-            l2_safe_head,
             engine_l2_safe_head,
             sync_complete_rx,
             derivation_signal_rx,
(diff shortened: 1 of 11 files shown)


theochap (approved):
Looks great!

crates/node/service/src/actors/derivation.rs
```diff
@@ -265,11 +262,10 @@ where
...
         }
 
-        // The L2 Safe Head must be advanced before producing new payload attributes.
-        if self.engine_l2_safe_head.borrow().block_info.number <=
-            self.l2_safe_head.block_info.number
-        {
-            debug!(target: "derivation", engine_safe_head = ?self.engine_l2_safe_head.borrow().block_info.number, l2_safe_head = ?self.l2_safe_head.block_info.number, "L2 safe head unchanged");
+        // Wait for the engine to initialize unknowns prior to kicking off derivation.
+        let engine_safe_head = *self.engine_l2_safe_head.borrow();
+        if engine_safe_head.block_info.hash.is_zero() {
```
clabby, line 267: We also need to add a condition here that the safe head has actually changed. Can use the "has seen" functionality of the watch channel for this thankfully :D A combo of [`has_changed`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.has_changed) + [`borrow_and_update`](https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.borrow_and_update) will do the trick.

crates/node/service/src/actors/derivation.rs
```diff
@@ -259,18 +256,35 @@ where
...
             return Ok(());
         }
 
-        // The L2 Safe Head must be advanced before producing new payload attributes.
-        if self.engine_l2_safe_head.borrow().block_info.number <=
-            self.l2_safe_head.block_info.number
-        {
-            debug!(target: "derivation", engine_safe_head = ?self.engine_l2_safe_head.borrow().block_info.number, l2_safe_head = ?self.l2_safe_head.block_info.number, "L2 safe head unchanged");
+        // If the safe head hasn't changed, don't step on the pipeline.
+        if self.engine_l2_safe_head.has_changed().map_or(false, |v| !v) {
```
clabby, line 279: As is, this check would fail to skip derivation if the channel has closed (meaning the engine actor has shut off.) We should hard error in that case.

clabby (approved):

crates/node/service/src/actors/derivation.rs
refcell: Good point

PR #1709, opened 2025-05-09
fix(node/rpc): removed the panic and added error when building RPC actor

### LINKED ISSUE #1704: fix(node/rpc): don't panic when merging new RPC routes to RPCActor
## Description
When building the RPC Actor we are using the following pattern:

```rust
        // The RPC Server should go last to let other actors register their rpc modules.
        let rpc = if let Some(mut rpc) = self.rpc() {
            if let Some(p2p_module) = p2p_module {
                rpc = rpc.merge(p2p_module.into_rpc()).expect("failed to merge p2p rpc module");
            }

            rpc = rpc.merge(rollup_rpc.into_rpc()).expect("failed to merge engine rpc module");
...
```

We should gracefully return an error instead of panicking in that case.

Relevant comment: https://github.com/op-rs/kona/pull/1664/files/1ac3be64c68d06816ccaaa1ca5fe1a3a45229389#r2080498084

PoulavBhowmick03:
Potentially Fixes #1704

### DIFF
--- a/crates/node/service/src/service/standard/error.rs
+++ b/crates/node/service/src/service/standard/error.rs
@@ -1,6 +1,7 @@
 //! Contains the error type for the [`crate::RollupNode`].
 
 use crate::SyncStartError;
+use jsonrpsee::server::RegisterMethodError;
 use kona_derive::errors::PipelineErrorKind;
 use kona_engine::EngineStateBuilderError;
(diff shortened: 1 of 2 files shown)


theochap (approved):
Thanks for taking care of that!

refcell (approved):
Nice work!

PR #1713, opened 2025-05-12
chore(supervisor): Dedup `SupervisorApi`

emhane:
Closes https://github.com/op-rs/kona/issues/1731


emhane:
blocked by https://github.com/alloy-rs/op-alloy/pull/518

crates/supervisor/core/src/supervisor.rs
```diff
@@ -2,15 +2,32 @@ use core::fmt::Debug;
...
+#[derive(Debug, Error, PartialEq, Eq)]
 pub enum SupervisorError {
     /// Indicates that a feature or method is not yet implemented.
     #[error("functionality not implemented")]
     Unimplemented,
+    /// Data availability errors.
+    ///
+    /// Spec <https://github.com/ethereum-optimism/specs/blob/main/specs/interop/supervisor.md#protocol-specific-error-codes>.
+    #[error(transparent)]
+    InvalidInboxEntry(#[from] InvalidInboxEntry),
```
itschaindev, line 21: There is another [InvalidInboxEntry](https://github.com/op-rs/kona/blob/9a7885769387b67409263224d9ae986d5e5fac60/crates/protocol/interop/src/errors.rs#L123).  Should we rename it there otherwise it'll cause confusion?

itschaindev (changes requested):

dhyaniarun1993 (approved):

crates/supervisor/core/src/supervisor.rs
emhane: good catch! that one should have been removed, thought i did...

emhane: ah yes, it was removed https://github.com/op-rs/kona/pull/1778

clabby (approved):

PR #1733, opened 2025-05-12
chore(workspace): Update REVM Deps

refcell:
### Description

Small pr to update `revm` + it's associated evm dependencies.


clabby (approved):

clabby:
Action tests flake looks unrelated to these changes (same isPrague race condition we've been seeing from the in-mem geth backend.) Re-running.

Issue #1620, opened 2025-05-01
feat(node/engine): Record Pending Safe Block Height


refcell:
### Description

Add a [gauge metric](https://docs.rs/metrics/0.24.2/metrics/struct.Gauge.html) to the `kona-engine` crate that tracks the pending safe block height. Since mutation to the `EngineState` is encapsulated in methods, the metric can be updated when the [`set_pending_safe_head`](https://github.com/op-rs/kona/blob/main/crates/node/engine/src/state/core.rs#L119) method is called.

### Output

Use `kona-macros` to set the block height on a new gauge metric when the [`set_pending_safe_head`](https://github.com/op-rs/kona/blob/main/crates/node/engine/src/state/core.rs#L119) method is called. Set the height as the [`L2BlockInfo.block_info.number`](https://github.com/op-rs/kona/blob/main/crates/protocol/protocol/src/block.rs#L69).

Unit test the metric output. The `metrics-exporter-prometheus` crate has [unit tests](https://github.com/metrics-rs/metrics/blob/main/metrics-exporter-prometheus/src/exporter/builder.rs#L569-L591) that should be used as a resource.

PR #1741, opened 2025-05-12
feat(engine): Chain label metrics

clabby:
## Overview

Adds metrics for `kona-engine`'s `EngineState`.

closes https://github.com/op-rs/kona/issues/1621
closes https://github.com/op-rs/kona/issues/1620
closes https://github.com/op-rs/kona/issues/1619
closes https://github.com/op-rs/kona/issues/1618
closes https://github.com/op-rs/kona/issues/1617


crates/node/engine/src/state/core.rs
```diff
@@ -123,23 +125,73 @@ impl EngineState {
...
+    use metrics_exporter_prometheus::PrometheusBuilder;
+    use rstest::rstest;
+
+    #[rstest]
+    #[case::set_unsafe(EngineState::set_unsafe_head, "unsafe", 1)]
+    #[case::set_cross_unsafe(EngineState::set_cross_unsafe_head, "cross-unsafe", 2)]
+    #[case::set_local_safe(EngineState::set_local_safe_head, "local-safe", 3)]
+    #[case::set_safe_head(EngineState::set_safe_head, "safe", 4)]
+    #[case::set_finalized_head(EngineState::set_finalized_head, "finalized", 5)]
+    fn test_chain_label_metrics(
```
refcell, line 176: Great test case.

refcell (approved):

crates/node/engine/src/metrics/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+
+    /// Describes metrics used in [`kona_engine`][crate].
+    pub fn describe() {
+        metrics::describe_gauge!(Self::BLOCK_LABELS, "Blockchain head labels");
+    }
+
+    /// Initializes metrics to `0` so they can be queried immediately by consumers of prometheus
+    /// metrics.
+    pub fn zero() {
+        // Blockchain head labels
```
emhane, line 29: can you explain why you chose to use gauge here pls? just curious

clabby: Counter would be the proper one to use, but `kona-macros` doesn't look to support counters well atm. We can follow up with moving over to counters with some added support there if we want.

PR #1742, opened 2025-05-12
chore(engine): Remove pending/backup unsafe head

clabby:
## Overview

Removes the pending and backup unsafe heads from the `EngineState` to simplify. These were necessary pre-Holocene, and since `kona-node` only supports execution layer sync, it will never actually need this functionality in the wild.


refcell (approved):
ty 👑

emhane (approved):
nice small scope

PR #1747, opened 2025-05-13
fix(node/p2p): Async Broadcast with Backoff

### LINKED ISSUE #1529: fix(node/p2p): Broadcast Unsafe Blocks
### Description

The `kona-p2p` networking stack attempts to broadcast unsafe blocks strictly in order on each iteration of the driver loop.

Currently, this is a [blocking operation](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/net/broadcast.rs#L32-L49). Instead of blocking, the forwarding can be spawned into a thread. It should hold a lock on the unsafe block list and only be spawned if there isn't already another thread broadcasting the blocks.

refcell:
### Description

Broadcasts blocks asynchronously with a backoff.

Closes #1529

### DIFF
--- a/crates/node/p2p/src/net/broadcast.rs
+++ b/crates/node/p2p/src/net/broadcast.rs
@@ -1,8 +1,15 @@
 //! Broadcast handles broadcasting unsafe blocks in-order.
 
+use backon::{ExponentialBuilder, Retryable};
 use op_alloy_rpc_types_engine::OpNetworkPayloadEnvelope;
-use std::collections::VecDeque;
-use tokio::sync::broadcast::{Receiver as BroadcastReceiver, Sender as BroadcastSender};
+use std::{
+    collections::VecDeque,
+    sync::{Arc, Mutex},
+};
+use tokio::{
+    sync::broadcast::{Receiver as BroadcastReceiver, Sender as BroadcastSender},
+    time::Duration,
+};
 
 /// The `Broadcast` struct is responsible for broadcasting unsafe blocks in order.
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1747 👈
* `main`

crates/node/p2p/src/net/broadcast.rs
```diff
@@ -34,17 +39,40 @@ impl Broadcast {
...
+        if self.broadcasting.lock().map_or(true, |v| *v) {
+            trace!(target: "net", "Broadcasting in flight, skipping");
+            return;
+        }
+        if let Ok(mut broadcasting) = self.broadcasting.lock() {
+            *broadcasting = true;
+        }
+        let flag = Arc::clone(&self.broadcasting);
+        let channel = self.channel.clone();
+        let mut drained = self.buffer.drain(..).collect::<VecDeque<_>>();
```
clabby, line 51: Can we just `mem::take` here instead? No need to iterate and drain if we're doing the full thing.

crates/node/p2p/src/net/broadcast.rs
```diff
@@ -34,17 +39,40 @@ impl Broadcast {
...
+                if let Some(block) = drained.front() {
+                    if let Err(e) = channel.send(block.clone()) {
+                        trace!(target: "net", ?e, "Failed to broadcast block");
+                        count += 1;
+                    } else {
+                        count = 0;
+                    }
+                    drained.pop_front();
+                }
             }
```
clabby, line 72: We can use `backon` here instead of a hand-rolled backoff strategy; Once this is rebased on #1753 should be able to introduce it.

crates/node/p2p/src/net/broadcast.rs
```diff
@@ -34,17 +39,40 @@ impl Broadcast {
...
-        }
+            if let Ok(mut broadcasting) = flag.lock() {
+                *broadcasting = true;
```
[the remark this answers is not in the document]
clabby: Yeah this looks right.

clabby:
Actions flake looks unrelated (reeeee.)

Can re-run and it'll pass.

crates/node/p2p/src/net/broadcast.rs
refcell: Good catch thanks

PR #1755, opened 2025-05-13
fix(node/p2p): Bootnodes in Bootstrap + Backoff Discv5 Restart

refcell:
### Description

Fixes the `kona-node` bootstrapping to use the manually passed bootnodes in the bootnode bootstrapping to ensure they are prioritized.

Previously, the custom bootnodes were only bootstrapped via the bootstore which is a lower priority than the bootnodes.

This PR also updates the initialization of the discovery service to use an exponential backoff strategy via [backon](https://crates.io/crate/backon).
Before, kona's discovery service would just try to restart in a loop.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4598,6 +4598,7 @@ dependencies = [
  "alloy-rpc-types-engine",
  "arbitrary",
  "arbtest",
+ "backon",
  "derive_more",
  "dirs",
  "discv5",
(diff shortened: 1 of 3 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #1755 👈
* `main`

theochap (approved):

PR #1758, opened 2025-05-14
feat(node): Version info metrics

clabby:
## Overview

Adds metrics for the version information of the node, including:
1. The binary version
2. The build timestamp
3. Any bin features enabled
4. The commit sha of the build
5. The target triple of the binary
6. The build profile

Also uses this rich version information for the `clap` version.

### DIFF
--- a/bin/node/build.rs
+++ b/bin/node/build.rs
@@ -1,0 +1,92 @@
+//! Derived from [`reth-node-core`][reth-build-script]
+//!
+//! [reth-build-script]: https://github.com/paradigmxyz/reth/blob/805fb1012cd1601c3b4fe9e8ca2d97c96f61355b/crates/node/core/build.rs
+
+#![allow(missing_docs)]
+
+use std::{env, error::Error};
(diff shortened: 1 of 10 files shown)


bin/node/build.rs
```diff
@@ -0,0 +1,89 @@
...
+    emitter.add_instructions(&build_builder)?;
+
+    let cargo_builder = CargoBuilder::default().features(true).target_triple(true).build()?;
+
+    emitter.add_instructions(&cargo_builder)?;
+
+    let git_builder =
+        Git2Builder::default().describe(false, true, None).dirty(true).sha(false).build()?;
+
+    emitter.add_instructions(&git_builder)?;
```
theochap, line 25: Super nit, but would you mind adding a quick comment explaining what are we tracking?

theochap (approved):
Amazing!

bin/node/build.rs
clabby: done :)

PR #1769, opened 2025-05-14
fix(bin/node): Argument Defaults

refcell:
### Description

Instead of redefining the argument defaults, use the values defined in the clap fields.


clabby (approved):
nice.

PR #1773, opened 2025-05-14
chore(sources): Refactor sync start to use alloy providers

clabby:
## Overview

Refactors sync start to use alloy providers over the derivation wrappers. This simplifies usage a bit, with not having to have mutable references to the caching derivation providers.

Also removes `L2BlockInfo::from_rpc_block_and_genesis`; alloy now has helper functions to convert RPC blocks into `alloy-consensus` block types.


theochap (approved):

crates/node/engine/src/client.rs
```diff
@@ -105,7 +105,7 @@ impl EngineClient {
         let Some(block) = block else {
             return Ok(None);
         };
-        Ok(Some(L2BlockInfo::from_rpc_block_and_genesis(block, &self.cfg.genesis)?))
+        Ok(Some(L2BlockInfo::from_block_and_genesis(&block.into_consensus(), &self.cfg.genesis)?))
```
refcell, line 108: This was erroring iirc

crates/protocol/protocol/src/block.rs
```diff
@@ -141,27 +141,6 @@ impl L2BlockInfo {
...
-        );
-        if block_info.number == genesis.l2.number {
-            if block_info.hash != genesis.l2.hash {
-                return Err(FromBlockError::InvalidGenesisHash);
-            }
-            return Ok(Self { block_info, l1_origin: genesis.l1, seq_num: 0 })
-        }
-        Self::from_block_and_genesis(&block.into_consensus(), genesis)
-    }
-
```
refcell, line 164: We need this because the hash produced by `into_consensus` wasn't right.

If we're going to delete this, we should add a test to ensure `from_block_and_genesis` works as intended.

refcell (changes requested):

crates/protocol/protocol/src/block.rs
clabby: Also seems to be working on the dev instance without this function - just have to make sure the rpc block is fetched with `BlockTransactionsKind::Full`.

PR #1774, opened 2025-05-14
fix(kurtosis): fix kurtosis configuration

theochap:
## Description
Fixes the kurtosis network params to be not fail when running with `just kurtosis-up`. Also ensures that the sequencer is run with `op-geth` as `op-reth` is not yet fully compatible with sequencer mode. Change log levels to debug


refcell (approved):
Thanks king!

PR #1771, opened 2025-05-14
[DRAFT] feat(node/engine): add reset event

theochap:

PR #1789, opened 2025-05-15
feat(node): Propagate engine reset to pipeline

clabby:
## Overview

Builds on #1771, adding support for propagating the engine reset back to the derivation pipeline.


crates/node/service/src/actors/engine.rs
```diff
@@ -146,14 +148,22 @@ impl EngineLauncher {
         let state = self.state_builder().build().await?;
         let (engine_state_send, _) = tokio::sync::watch::channel(state);
 
-        Ok(Engine::new(state, engine_state_send))
+        let mut engine = Engine::new(state, engine_state_send);
+
+        engine
+            .reset(Arc::new(self.client()), &self.config)
+            .await
+            .expect("TODO: Handled in follow-up PR");
```
clabby, line 156: #1790 gets rid of initialization here in favor of a more streamlined path.

crates/node/service/src/actors/engine.rs
```diff
@@ -203,6 +213,21 @@ impl NodeActor for EngineActor {
                           let sent = self.engine_l2_safe_head_tx.send_if_modified(update);
                           trace!(target: "engine", ?sent, "Attempted L2 Safe Head Update");
                         }
+                        Err(EngineTaskError::Reset(e)) => {
+                            warn!(target: "engine", ?e, "Received reset request");
```
emhane, line 217:
```suggestion
                        Err(EngineTaskError::Reset(err)) => {
                            warn!(target: "engine", ?err, "Received reset request");
```
thinks they key `err` makes logs a bit better readable than `e` for newly onboarded, but no biggie

emhane (approved):
super nice !

bin/node/src/commands/node.rs
```diff
@@ -111,6 +111,7 @@ impl NodeCommand {
         let engine_client = kona_engine::EngineClient::new_http(
             self.l2_engine_rpc.clone(),
             self.l2_provider_rpc.clone(),
+            self.l2_provider_rpc.clone(),
```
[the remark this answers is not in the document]
clabby: Ah yes - good catch. Typo!

Didn't matter here since it isn't used for JWT auth, but fixed it up.

PR #1792, opened 2025-05-16
feat(node): Add kurtosis e2e testing skeleton

theochap:
## Description
This PR adds the necessary infrastructure to automatically deploy and run e2e tests for the node with Kurtosis. 

Next PR will add that to CI to make sure we don't break the node tests when adding new features.

Note that I had to port the mise config from the ethereum optimism.


tests/devnets/simple-kona.yaml
```diff
@@ -0,0 +1,75 @@
...
+
+optimism_package:
+  chains:
+    - participants:
+      # Note: it seems that op-reth isn't fully compatible with the sequencer mode.
+      # So we use op-geth for now.
+      - el_type: op-geth
+        cl_type: op-node
+        el_log_level: "debug"
+        cl_log_level: "debug"
```
emhane, line 12: nice !!

tests/Justfile
refcell, line 1: Can we add a very small `README.md` to this repo with a short `Usage` section on executing the `just` targets

tests/devnets/simple-kona.yaml
```diff
@@ -0,0 +1,75 @@
...
+        count: 1
+      - el_type: op-reth
+        cl_type: kona-node
+        el_log_level: "debug"
+        cl_image: "ghcr.io/op-rs/kona/kona-node:theo-kickstart-testing"
```
refcell, line 17: Is it possible to pull this in via env variable? Then we could reference a locally built image

tests/go.mod
```diff
@@ -0,0 +1,171 @@
...
+	gopkg.in/yaml.v2 v2.4.0 // indirect
+	gopkg.in/yaml.v3 v3.0.1 // indirect
+	lukechampine.com/blake3 v1.3.0 // indirect
+	rsc.io/tmplfunc v0.0.3 // indirect
+)
+
+replace github.com/ethereum-optimism/optimism/op-node v0.10.14 => github.com/ethereum-optimism/optimism v1.13.2
+
+// TODO Working on a local fork of op-node. Remove that line before merging.
+replace github.com/ethereum-optimism/optimism v1.13.2 => ../../optimism
```
refcell, line 169: Is this still `TODO`?

refcell (commented):
Few suggestions/comments

tests/devnets/simple-kona.yaml
theochap: I think we can use go templates to define external json configs to inject in the yaml file. I am not sure that environment variables would work though. Can try to do it in a follow-up PR. In the meantime I removed the `cl_image` field inside this yaml file

tests/go.mod
theochap: Good catch! Removed

refcell (approved):
Nice!

PR #1793, opened 2025-05-16
feat: dependency set

### LINKED ISSUE #1682: feat(protocol/interop): Implement dependency-set
# Description

Implement dependency-set and helper method(probably inside interop crate)

dhyaniarun1993:
Closes #1682


crates/protocol/interop/src/depset.rs
```diff
@@ -0,0 +1,225 @@
+use crate::MESSAGE_EXPIRY_WINDOW;
+use alloy_primitives::ChainId;
+use kona_registry::HashMap;
+
+/// Configuration for a dependency of a chain
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+pub struct ChainDependency {
+    /// chain_index is the unique short identifier for this chain
```
emhane, line 10:
```suggestion
    /// Chain_index is the unique short identifier for this chain
```
for docs we use proper sentences (or skipping trailing punctuation if you like), for comments not

crates/protocol/interop/src/depset.rs
```diff
@@ -0,0 +1,225 @@
...
+    pub chain_index: u32,
+
+    /// activation_time is the timestamp (in seconds) when the chain becomes part of the dependency
+    /// set. This is the minimum timestamp of the inclusion of an executing message that can be
+    /// considered valid.
```
emhane, line 15:
```suggestion
    /// Timestamp (in seconds) when the chain becomes part of the dependency
    /// set. This is the minimum timestamp of the inclusion of an executing message that can be
    /// considered valid.
```
try to avoid using field names in docs since they may change. if they are publicly accessible fields, then its ok to use a doc link e.g. (activation_time)(Self::activation_time).

emhane (approved):
lgtm, remaining nits on docs

emhane:
pending @itschaindev or @sadiq1971

crates/protocol/interop/src/depset.rs
```diff
@@ -0,0 +1,224 @@
...
+use alloy_primitives::ChainId;
+use kona_registry::HashMap;
+
+/// Configuration for a dependency of a chain
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
+pub struct ChainDependency {
+    /// The unique short identifier for this chain.
+    pub chain_index: u32,
```
itschaindev, line 11: What will we use this for? Since ChainId already uniquely identifies the chain.

itschaindev (commented):
lgtm, just a small comment.

crates/protocol/interop/src/depset.rs
dhyaniarun1993: Thanks for pointing that out. It's part of the current spec. 
I believe both are same at the moment but would be different later on. From what I can understand it's supposed to an internal, potentially abstracted, and locally scoped identifier. Let's cross check this with the team.

dhyaniarun1993: Axel Reponse
```
It is for storage compression - ChainIDs are uint64 while Indexes are (IIRC) uint32, so we slice off half the storage by mapping them down. Right now they are just identical, but the DependencySet allows you to map to specific indexes if you like. I think we are finding that the Index mapping should common among nodes on a superchain (for database compatability), but they don't need to be.
```

PR #1801, opened 2025-05-17
fix(node): Zero `safe` + `finalized` hashes while EL syncs

clabby:
## Overview

Sets the `safe` and `finalized` hashes to `B256::ZERO` while the EL syncs, and also only finalizes the unsafe ref if the client is undergoing its _initial_ sync.

Also moves the priority of the `sync_complete_rx.recv()` future in the biased select within the derivation actor. Since the futures are polled in-order, that future always resolves once the channel has been closed - it must be placed last.

### DIFF
--- a/crates/node/engine/src/task_queue/tasks/insert/task.rs
+++ b/crates/node/engine/src/task_queue/tasks/insert/task.rs
@@ -123,23 +123,38 @@ impl EngineTaskExt for InsertUnsafeTask {
             safe_block_hash: state.safe_head().block_info.hash,
             finalized_block_hash: state.finalized_head().block_info.hash,
         };
-        if state.sync_status == SyncStatus::ExecutionLayerNotFinalized {
-            // Use the new payload as the safe and finalized block for the FCU.
-            fcu.safe_block_hash = self.envelope.payload.block_hash();
-            fcu.finalized_block_hash = self.envelope.payload.block_hash();
 
-            // Update the local engine state to match.
-            state.set_unsafe_head(new_unsafe_ref);
-            state.set_safe_head(new_unsafe_ref);
-            state.set_local_safe_head(new_unsafe_ref);
-            state.set_finalized_head(new_unsafe_ref);
-        } else if state.sync_status == SyncStatus::ExecutionLayerWillStart {
-            // For the first FCU sent to the EL, set the `safe` and `finalized` hashes to 0. This is
-            // a special case to trigger optimistic head sync on `op-reth`.
-            state.sync_status = SyncStatus::ExecutionLayerStarted;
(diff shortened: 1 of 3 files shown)


crates/node/engine/src/task_queue/tasks/insert/task.rs
```diff
@@ -130,12 +130,21 @@ impl EngineTaskExt for InsertUnsafeTask {
...
+            // initial sync of the node. Update the safe and finalized heads to the
+            // new unsafe ref.
+            let genesis_hash = self.rollup_config.genesis.l2.hash;
+            if state.finalized_head().block_info.hash == genesis_hash &&
+                state.safe_head().block_info.hash == genesis_hash
+            {
+                state.set_safe_head(new_unsafe_ref);
+                state.set_local_safe_head(new_unsafe_ref);
+                state.set_finalized_head(new_unsafe_ref);
+            }
```
theochap, line 144: Just to be sure: how does this piece of code prevent reth from saturating its state with new hashes and keep the initial sync going? As far as I understand here:
- When we are sending the initial FCU (with zeroed out hashes) we kick off optimistic sync on `op-reth`
- Then we send the reset signal that updates the safe and finalized head to the genesis hash
- Then we get the new unsafe payloads and update the safe and finalized heads *once*, which would cause the EL to override its state and stop syncing

Don't we want to just remove the genesis branch here and wait for the `ExecutionLayer` to have finished to send a FCU with the most recent heads?

theochap: Also nit but it may help readability if we used a match here instead of if/else statements

clabby: What this branch does is prevent promoting the latest unsafe ref to `finalized` and `safe` if we've already found them (i.e., it's not the first sync.) Without this branch, when EL sync finishes on startup, the node will wrongfully promote the latest unsafe block to `safe` and `finalized` respectively. This is bad, following the first sync, since there's no way for us to know whether or not that block is `safe` or `finalized`.

The outer branch here should _only_ be hit once at startup, when EL sync finishes. And the inner branch that's added here should only be hit on the first ever sync.

clabby: The other branch, where we set the `safe_block_hash` and `finalized_block_hash` to `Default::default()` (`B256::ZERO`) occurs when payloads are inserted _during_ EL sync, as a means of not creating a higher priority backfill job for the EL. In normal operation (after the first sync), only when EL sync finishes can we actually inform the execution layer of the `safe`/`finalized` labels that we found via sync start.

theochap (approved):

PR #1804, opened 2025-05-17
fix(node): Build task + consolidate condition

clabby:
## Overview

Fixes the condition under which block building is triggered during payload attributes consolidation, and also fixes the build task.

### DIFF
--- a/crates/node/engine/src/task_queue/tasks/build/task.rs
+++ b/crates/node/engine/src/task_queue/tasks/build/task.rs
@@ -3,18 +3,18 @@
 use super::BuildTaskError;
 use crate::{
     EngineClient, EngineForkchoiceVersion, EngineGetPayloadVersion, EngineState, EngineTaskError,
-    EngineTaskExt, Metrics,
+    EngineTaskExt, ForkchoiceTask, Metrics,
 };
 use alloy_provider::ext::EngineApi;
 use alloy_rpc_types_engine::{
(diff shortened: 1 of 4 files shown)


crates/node/engine/src/task_queue/tasks/build/task.rs
theochap, line 1: Is there any logical changes to this file or is that mostly refactoring?

theochap: It's a bit hard to follow but it seems that there shouldn't be *any* logical changes.

clabby: Two logical changes - V4 payloads are now correctly parsed, and the FCU uses the attributes' parent hash as the `head_block_hash` when initiating the build job.

crates/protocol/protocol/src/block.rs
theochap, line 1: Same for here, looks like it is mostly refactoring. What are the logical changes I should be aware of?

theochap (approved):

crates/protocol/protocol/src/block.rs
clabby: `L2BlockInfo::from_payload_and_genesis` was fixed to properly construct the `L2BlockInfo`, using `alloy`'s native functionality to convert an `OpExecutionPayload` into a `Block<OpTxEnvelope>`. Previously it only took an `ExecutionPayload`, which doesn't extend to the OP Stack's modification for Isthmus (supported by `OpExecutionPayload`.)

PR #1808, opened 2025-05-19
feat(supervisor/storage): derivation schema

### LINKED ISSUE #1749: feat(supervisor/state-manager): derivation storage schema
## DerivationStorage tracks how each L2 block was derived from an L1 block.
Spec can be found here: #1684 

Database Schema:

### Derived Block Pairs:

* **Purpose:** Stores pairs of source and derived block.
* **Primary Key:** `derived_pairs:{chain_index}:{derived_block_num}`
    * `chain_index`: Identifier for the blockchain.
    * `derived_block_num`: The block number of the derived block.
* **Value:** Serialized `DerivedBlockPair` struct (format to be determined):

    ```
    type DerivedBlockPair struct {
        Source  BlockRef
        Derived BlockRef
    }

* **Secondary Index (Lookup by Source Block Number):**
    * **Key:** `derived_from_source:{chain_index}:{source_block_num}`
        * `chain_index`: Identifier for the blockchain.
        * `source_block_num`: The block number of the source block.
    * **Value:** Serialized list of strings, where each string is in the format `{derived_block_num}` (e.g., `["0:1000", "0:1005"]`).

Database Operations:
StoreDerived(source BlockRef, derived BlockRef)
SourceBlock(derived)BlockRef

Note: There might be additional operations required. Also research on what should we do with revision.

dhyaniarun1993:
Closes #1749


crates/supervisor/storage/src/models/mod.rs
```diff
@@ -50,7 +57,7 @@ macro_rules! impl_compression_for_compact {
 }
 
 // Implement compression logic for all value types stored in tables
-impl_compression_for_compact!(BlockRef, LogEntry);
+impl_compression_for_compact!(BlockRef, LogEntry, DerivedBlockPair, U64List);
 
 tables! {
```
itschaindev, line 62: I'm assuming these tables are chain specific. Would the DB handling for each chain be a part of another pr?

crates/supervisor/storage/src/models/common.rs
```diff
@@ -0,0 +1,98 @@
+//! Common model types used across various storage tables.
+
+use reth_codecs::Compact;
+use serde::{Deserialize, Serialize};
+use std::ops::{Deref, DerefMut};
+
+/// Wrapper for `Vec<u64>` to represent a list of numbers.
+#[derive(Debug, Clone, PartialEq, Eq, Default, Serialize, Deserialize, Compact)]
+pub struct U64List(pub Vec<u64>);
```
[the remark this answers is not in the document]
itschaindev: +1

crates/supervisor/storage/src/models/mod.rs
sadiq1971: yes

crates/supervisor/storage/src/models/mod.rs
```diff
@@ -69,6 +76,22 @@ tables! {
...
+    }
+
+    /// A table mapping a source block number to a list of its derived block numbers.
+    /// - Key: `u64` — source block number
+    /// - Value: [`U64List`] — list of derived block numbers
+     table SourceToDerivedBlockNumbers {
+        type Key = u64;
+        type Value = U64List;
+    }
 }
```
sadiq1971, line 95: Instead of maintaining two table we can create a DUP SORT table like this:
```
table DerivedBlocks {
        type Key = u64; // source block
        type SubKey = u64 // derived block
        type Value = DerivedBlockPair;
    }
```

crates/supervisor/storage/src/models/common.rs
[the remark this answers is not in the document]
dhyaniarun1993: `Vec<u64>` doesn't implement `Compress` and `Decompress` trait which is required by the table.
I will add a todo here to add support for `Vec<u64>`.

crates/supervisor/storage/src/models/mod.rs
dhyaniarun1993: We need to have both lookup to fetch derived blocks from source block and source block by derived block. I don't think we can achieve this by `DUP SORT` table

itschaindev (approved):

emhane (approved):

PR #1809, opened 2025-05-19
feat(engine): Simplify EL sync startup routine

clabby:
## Overview

Simplifies the EL sync startup routine, removing the `ExecutionLayerNotFinalized` stage. Rather than performing the sync start lookup at startup, the sync start routine is performed _after_ EL sync finishes now.

This also improves the security of the sync start routine. Because we no longer blindly promote `safe` + `finalized` when the first EL sync finishes, we can find a `safe` block by traversing backwards from the `unsafe` head, looking for an L2 block whose L1 origin is at least one sequence window behind the `unsafe` head's L1 origin block.

### DIFF
--- a/crates/node/engine/src/sync.rs
+++ b/crates/node/engine/src/sync.rs
@@ -14,9 +14,7 @@
 ///
 /// 1. Sync hasn't started yet. [`SyncStatus::ExecutionLayerWillStart`]
 /// 2. The sync has started. [`SyncStatus::ExecutionLayerStarted`]
-/// 3. The sync has finished but the last block is not finalized.
-///    [`SyncStatus::ExecutionLayerNotFinalized`]
-/// 4. The sync has finished. [`SyncStatus::ExecutionLayerFinished`]
+/// 3. The sync has finished. [`SyncStatus::ExecutionLayerFinished`]
 ///
 /// Once the execution layer sync has finished, the last block is marked
 /// as finalized and consolidation is performed.
@@ -29,27 +27,20 @@ pub enum SyncStatus {
     ExecutionLayerWillStart = 1,
     /// Execution sync has started.
     ExecutionLayerStarted = 2,
-    /// Execution sync has finished but the last block is not finalized.
-    ExecutionLayerNotFinalized = 3,
     /// Execution sync has finished.
     /// At this point, consolidation is being performed.
-    ExecutionLayerFinished = 4,
+    ExecutionLayerFinished = 3,
 }
 
 impl SyncStatus {
     /// Returns true if the execution layer sync has started.
     pub const fn has_started(&self) -> bool {
-        matches!(self, Self::ExecutionLayerStarted)
+        matches!(self, Self::ExecutionLayerStarted | Self::ExecutionLayerFinished)
     }
 
     /// Returns true if syncing is in progress.
     pub const fn is_syncing(&self) -> bool {
-        matches!(
-            self,
-            Self::ExecutionLayerWillStart |
-                Self::ExecutionLayerStarted |
-                Self::ExecutionLayerNotFinalized
-        )
+        matches!(self, Self::ExecutionLayerWillStart | Self::ExecutionLayerStarted)
     }
 
     /// Returns true if syncing is finished
--- a/crates/node/engine/src/task_queue/tasks/insert/task.rs
+++ b/crates/node/engine/src/task_queue/tasks/insert/task.rs
@@ -41,18 +41,8 @@ impl InsertUnsafeTask {
         Self { client, rollup_config, version, envelope }
     }
 
-    /// Checks the response of the `engine_newPayload` call, and updates the sync status if
-    /// necessary.
-    fn check_new_payload_status(
-        &self,
-        state: &mut EngineState,
-        status: &PayloadStatusEnum,
-    ) -> bool {
-        debug!(target: "engine", ?status, "Checking payload status");
-        if matches!(status, PayloadStatusEnum::Valid) {
-            debug!(target: "engine", "Valid new payload status. Finished execution layer sync");
-            state.sync_status = SyncStatus::ExecutionLayerFinished;
-        }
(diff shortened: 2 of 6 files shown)


crates/node/engine/src/task_queue/tasks/insert/task.rs
```diff
@@ -66,7 +71,11 @@ impl InsertUnsafeTask {
...
             state.sync_status == SyncStatus::ExecutionLayerStarted
         {
-            state.sync_status = SyncStatus::ExecutionLayerNotFinalized;
+            info!(
+                target: "engine",
+                "Finished execution layer sync."
+            );
+            state.sync_status = SyncStatus::ExecutionLayerFinished;
         }
         matches!(status, PayloadStatusEnum::Valid | PayloadStatusEnum::Syncing)
```
theochap, line 80: It seems that `check_new_payload_status` and `check_forkchoice_updated_status` have exactly the same body. Would it be possible to remove one of them?

crates/node/engine/src/sync.rs
```diff
@@ -29,27 +27,20 @@ pub enum SyncStatus {
     ExecutionLayerWillStart = 1,
```
theochap, line 27: Question: Is there a reason we keep around the `ExecutionLayerWillStart` variant? It seems we can merge this one with `ExecutionLayerStarted` and remove it.

crates/node/engine/src/task_queue/tasks/insert/task.rs
clabby: oops; Yep - we can remove the state update from the payload status check.

crates/node/service/src/actors/engine.rs
```diff
@@ -225,18 +238,9 @@ impl NodeActor for EngineActor {
...
-                                  *head = state_safe_head;
-                                  return true;
-                              }
-                              false
-                          };
-                          let sent = self.engine_l2_safe_head_tx.send_if_modified(update);
-                          trace!(target: "engine", ?sent, "Attempted L2 Safe Head Update");
+                            trace!(target: "engine", "[ENGINE] tasks drained");
+                            // Update the l2 safe head if needed.
+                            self.maybe_update_safe_head();
```
theochap, line 243: Maybe it would be safer to move the `self.maybe_update_safe_head()` call outside of the `match` statement so that we can capture all the updates in a single place. For instance, when we raise `EngineTaskError::Reset`, it seems we're not calling `maybe_update_safe_head()` at all currently

crates/node/engine/src/sync.rs
clabby: Don't think so, other than for implementation clarity. The default state isn't actually that the EL sync has started, but that state holds no real significance. Can remove it if you prefer!

crates/node/service/src/actors/engine.rs
clabby: yeah good catch. Believe it currently works because the next call to `drain` will attempt the update again, but that would unblock derivation sooner.

crates/node/sources/src/sync/forkchoice.rs
```diff
@@ -46,22 +46,21 @@ impl L2ForkchoiceState {
...
-                            .await?
-                            .ok_or(SyncStartError::BlockNotFound(BlockNumberOrTag::Finalized))?,
-                        Err(e) => return Err(e.into()),
+        let finalized = {
+            let rpc_block =
+                match l2_provider.get_block(BlockNumberOrTag::Finalized.into()).full().await {
+                    Ok(Some(block)) => block,
+                    Ok(None) => {
+                        l2_provider.get_block(cfg.genesis.l2.number.into()).full().await?.ok_or(
+                            SyncStartError::BlockNotFound(BlockNumberOrTag::Finalized.into()),
```
theochap, line 55: I think that here `alloy` converts `BlockNumberOrTag::Finalized` into `BlockId::Number(1)`. Don't know if that's intentional or not, but it *seems* a bit buggy (because it can also indicates the block number 1 afaik). Maybe we should mention it somewhere in the error docs?

clabby: Ah this is actually a small bug; Should be `SyncStartError::BlockNotFound(cfg.genesis.l2.number.into())`

clabby: Fixed that up

crates/node/sources/src/sync/mod.rs
```diff
@@ -49,297 +37,74 @@ pub async fn find_starting_forkchoice(
...
+            break;
+        } else {
+            let block = l2_provider
+                .get_block(safe_cursor.block_info.parent_hash.into())
+                .full()
+                .await?
+                .ok_or(SyncStartError::BlockNotFound(safe_cursor.block_info.parent_hash.into()))?;
+            safe_cursor =
+                L2BlockInfo::from_block_and_genesis(&block.into_consensus(), &cfg.genesis)?;
         }
```
theochap, line 103: Really nice simplifications!

crates/node/engine/src/sync.rs
theochap: I may be missing something but that state doesn't seem to actually track whether the EL is currently waiting for peers to initiate the backfill routine. Don't we just change the state machine to `ExecutionLayerStarted` whenever we receive the very first `UnsafePayload`? This *seems* pretty independent from EL sync.

I would be somewhat biased to trying to simplify as much as possible the data structures. If that enum only has two variants, it could even be a boolean which I guess would reduce the mental overhead. Happy to hear pushback if you don't think that's relevant/good!

clabby: Correct - right now, we just immediately switch over once we insert the first payload. We can tweak it a bit to support this, though.

Let's swap it over for now. Always can add those metrics later, deleting code is good :)

theochap (approved):
Looks great

PR #1810, opened 2025-05-19
chore(engine): Remove `SyncStatus`

clabby:
## Overview

Removes the `SyncStatus` enum in favor of just using a `bool` in the `EngineState`.


theochap (approved):
Lovely, stunning

Issue #1562, opened 2025-04-29
feat(node/p2p): `opp2p_peers` RPC Method Support


refcell:
### Description

Similar to how the `op-node` [supports the `opp2p_peers` endpoint](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/p2p/rpc_server.go#L153-L190), add support for `opp2p_peers` in the `kona-node` [here](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/rpc/server.rs#L61-L64).

PR #1811, opened 2025-05-20
feat(node/p2p): enable support for the peers RPC endpoint

theochap:
## Description

To be able to properly test the `kona-node` synchronization primitives we need to add RPC methods that would allow to interact with the node's internal state.

This PR starts the implementation of the `opp2p_peers` endpoint. A series of follow-up PRs will complete the implementation

Progress towards #1562

### DIFF
--- a/crates/node/p2p/src/rpc/request.rs
+++ b/crates/node/p2p/src/rpc/request.rs
@@ -1,10 +1,20 @@
 //! Contains the p2p RPC request type.
 
+use std::collections::HashMap;
+
 use crate::{Discv5Handler, GossipDriver};
-use discv5::multiaddr::Protocol;
+use alloy_primitives::map::foldhash::fast::RandomState;
+use discv5::{
+    enr::{NodeId, k256::ecdsa},
+    multiaddr::Protocol,
+};
+use libp2p::PeerId;
 use tokio::sync::oneshot::Sender;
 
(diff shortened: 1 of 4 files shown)


crates/node/p2p/src/rpc/request.rs
```diff
@@ -47,6 +66,130 @@ impl P2pRpcRequest {
...
+            error!(target: "p2p::rpc", "Failed to get total connected peers. The number of connected peers is too large and overflows u32.");
+            return;
+        };
+
+        let peer_ids: Box<dyn Iterator<Item = PeerId>> = if connected {
+            Box::new(gossip.swarm.connected_peers().cloned())
+        } else {
+            Box::new(gossip.peerstore.keys().cloned())
+        };
+        let peer_ids: Vec<PeerId> = peer_ids.collect();
```
clabby, line 85: nit - can remove the need for the intermediate boxed trait object.
```suggestion
        let peer_ids: Vec<PeerId> = if connected {
            gossip.swarm.connected_peers().cloned().collect()
        } else {
            gossip.peerstore.keys().cloned().collect()
        };
```

clabby (approved):
Nice - good start.

crates/node/p2p/src/rpc/request.rs
theochap: Very good suggestion!

PR #1816, opened 2025-05-20
feat(node-service): Handle derivation reset events

clabby:
## Overview

Forwards derivation reset events to the engine. In the case of the `HoloceneActivation` reset, the reset can be handled internal to the pipeline without informing the engine.

Additionally, adds functionality to block the derivation actor while it is waiting for a signal to be sent back. This will prevent it from attempting to advance while the engine is resetting.

### DIFF
--- a/crates/node/service/src/actors/derivation.rs
+++ b/crates/node/service/src/actors/derivation.rs
@@ -5,14 +5,14 @@ use async_trait::async_trait;
 use kona_derive::{
     errors::{PipelineError, PipelineErrorKind, ResetError},
     traits::{Pipeline, SignalReceiver},
-    types::{ActivationSignal, ResetSignal, Signal, StepResult},
+    types::{ActivationSignal, Signal, StepResult},
 };
 use kona_protocol::{BlockInfo, L2BlockInfo, OpAttributesWithParent};
 use thiserror::Error;
 use tokio::{
     select,
     sync::{
-        mpsc::{UnboundedReceiver, UnboundedSender, error::SendError},
+        mpsc::{UnboundedReceiver, UnboundedSender},
         watch::Receiver as WatchReceiver,
     },
 };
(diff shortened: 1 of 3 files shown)


crates/node/service/src/actors/derivation.rs
```diff
@@ -53,16 +54,23 @@ where
...
     derivation_signal_rx: UnboundedReceiver<Signal>,
+
     /// The receiver for L1 head update notifications.
     l1_head_updates: UnboundedReceiver<BlockInfo>,
-    /// The sender for derived [OpAttributesWithParent]s produced by the actor.
+    /// The sender for derived [`OpAttributesWithParent`]s produced by the actor.
     attributes_out: UnboundedSender<OpAttributesWithParent>,
+    /// The reset request sender, used to handle [`PipelineErrorKind::Reset`] events and forward
+    /// them to the engine.
+    reset_request_tx: UnboundedSender<()>,
```
theochap, line 64: How about we use a `watch::Sender`/`watch::Receiver` instead? In that case if we have several reset signal we can just reset the engine once instead of resetting it several times

crates/node/service/src/actors/derivation.rs
```diff
@@ -53,16 +54,23 @@ where
...
 
     /// A flag indicating whether the derivation pipeline is ready to start.
     engine_ready: bool,
     /// A flag indicating whether or not derivation is idle. Derivation is considered idle when it
     /// has yielded to wait for more data on the DAL.
     derivation_idle: bool,
+    /// A flag indicating whether or not derivation is waiting for a signal. When waiting for a
+    /// signal, derivation cannot process any incoming events.
+    waiting_for_signal: bool,
```
theochap, line 73: Question: can we merge this `waiting_for_signal` boolean with the `engine_ready` one and only have one couple of `watch::Receiver<()>`/`watch::Sender<()>` between the engine and the derivation pipeline to kickoff derivation?

theochap: Note: if that's too much work it's absolutely fine to do it in a separate PR. Not want to block on that

theochap: We just have to be careful with race conditions, because we are starting to have a few intricate channels between the engine and the derivation actor

theochap (approved):

PR #1827, opened 2025-05-21
fix(p2p/rpc): fix formatting for the `opp2p-peers` rpc method

theochap:
## Description

This PR fixes the formatting of the `opp2p-peers` rpc method by ensuring that a `NodeId` is correctly formatted. The `Display` impl of the `NodeId` only partially exposes the `NodeId` hex-string.


clabby (approved):
Looks like this needs to be rebased on #1824 , but the peer ID changes lgtm.

PR #1830, opened 2025-05-21
feat(supervisor/storage): log storage 

### LINKED ISSUE #1746: feat(supervisor/state-manager):  log storage
## Implement Log Storage for the supervisor. 
Spec can be found here #1684

Database Schema:
**1. Log:**

* **Purpose:** Stores information about individual blockchain logs, including their hash and optional executing message data.
* **Key:** `{chain_index}:{block_num}:{log_idx}`
    * `chain_index`: Identifier for the blockchain.
    * `block_num`: The block number where the log originated.
    * `log_idx`: The index of the log within the block.
* **Value:** Serialized `Log` struct (format to be determined, e.g., JSON, Protocol Buffers):

    ```
    struct Log {
        Hash                        string            // Hex-encoded logHash of the initiating log
        ExecutingMessage object | null // Structured data for ExecutingMessage, or null if not applicable
    }
    ```

**2. Block References (Metadata):**

* **Purpose:** Stores metadata information about individual blockchain blocks.
* **Key:** `{chain_index}:{block_num}`
    * `chain_index`: Identifier for the blockchain.
    * `block_num`: The block number.
* **Value:** Serialized `BlockRef` struct (format to be determined):

    ```
    struct BlockRef {
        Hash             string
        Number        unsigned 64-bit integer
        ParentHash  string 
        Time             unsigned 64-bit integer 
    }
    ```
Possible Database Operations(High Level):
```
StoreBlockLogs(BlockRef, []Log)
Rewind(BlockNumber)
LastestBlock()BlockRef
FindBlock(BlockNumber)BlockRef
ContainsLog(BlockNumber, LogIdx, Timestamp, Hash) BlockRef
BlockLogs(BlockNumber)[]Log
```

sadiq1971:
Closes #1746


crates/supervisor/storage/src/error.rs
```diff
@@ -0,0 +1,42 @@
...
+/// This type is used as the source for most [`StorageError`] variants to allow
+/// flexibility in wrapping reth database errors uniformly.
+pub(crate) type SourceError = Box<dyn std::error::Error + Send + Sync>;
+
+/// Errors that may occur while interacting with supervisor log storage.
+///
+/// This enum is used across all implementations of the Storge traits.
+#[allow(dead_code)]
+#[derive(Debug, Error)]
+pub(crate) enum StorageError {
```
emhane, line 14: can we make this public? we should allow developers to extend the library in ways we can't imagine in the current moment

crates/supervisor/storage/src/error.rs
```diff
@@ -0,0 +1,42 @@
+use thiserror::Error;
+
+/// A dynamic error type for encapsulating low-level database/backend errors.
+///
+/// This type is used as the source for most [`StorageError`] variants to allow
+/// flexibility in wrapping reth database errors uniformly.
+pub(crate) type SourceError = Box<dyn std::error::Error + Send + Sync>;
```
emhane, line 7: is there no type safe way to do this?

crates/supervisor/storage/src/log/mdbx.rs
```diff
@@ -0,0 +1,311 @@
+//! Reth's MDBX-backed implementation of [`crate::log::LogStorage`] for supervisor state.
```
emhane, line 1:
```suggestion
//! Reth's MDBX-backed abstraction of [`LogStorage`](crate::log::LogStorage) for superchain state.
```

emhane (changes requested):
first pass

crates/supervisor/storage/src/error.rs
```diff
@@ -0,0 +1,42 @@
...
+    #[error("Transaction initialization failed")]
+    TransactionInit(#[source] SourceError),
+
+    /// Failed to commit a transaction.
+    #[error("Transaction commit failed")]
+    TransactionCommit(#[source] SourceError),
+
+    /// Failed to initialize a cursor or dup-sorted cursor.
+    #[error("Cursor initialization failed")]
+    CursorInit(#[source] SourceError),
```
dhyaniarun1993, line 37: I think we can simplify the error handling by using a single error variant for all database operations. In this we lose the ability to distinguish between read, write, and init errors at the type level but we can still inspect the source error if needed. Let me know what you think

```suggestion
    /// DatabaseError
    #[error("Database error")]
    Database(#[source] DatabaseError),
```

dhyaniarun1993 (changes requested):

crates/supervisor/storage/src/log/mdbx.rs
```diff
@@ -0,0 +1,311 @@
...
+where
+    TX: DbTxMut,
+{
+    type Error = StorageError;
+
+    fn store_block_logs(&self, block: &BlockInfo, logs: Vec<Log>) -> Result<(), Self::Error> {
+        debug!(target: "supervisor_storage", "Storing {} logs for block {}", logs.len(), block.number);
+        let tx = self.tx;
+        if let Err(e) = tx.put::<BlockRefs>(block.number, (*block).into()) {
+            error!(target: "supervisor_storage", "Failed to write block for {}: {e}", block.number);
```
dhyaniarun1993, line 56: I think we should do structured logging. 
```suggestion
            error!(
                target: "supervisor_storage",
                block_number = block.number,
                error = %e,
                "Failed to write block"
            );
```

sadiq1971: may be should go with full error trace? like `?e`

crates/supervisor/storage/src/log/mdbx.rs
```diff
@@ -0,0 +1,363 @@
...
+                target: "supervisor_storage",
+                block_number = block_number,
+                error = ?e,
+                "Failed to read block",
+            );
+            StorageError::Database(e)
+        })?;
+
+        let block = block_option.ok_or_else(|| {
+            warn!(target: "supervisor_storage", block_number = block_number, "Block not found");
```
emhane, line 104:
```suggestion
            warn!(target: "supervisor_storage", block_number, "Block not found");
```
you can just do this and it will use the variable name as key name

crates/supervisor/storage/src/log/mdbx.rs
```diff
@@ -0,0 +1,363 @@
...
+            error!(target: "supervisor_storage", error = ?e, "Failed to get dup cursor");
+            StorageError::Database(e)
+        })?;
+
+        for log in logs {
+            if let Err(e) = cursor.append_dup(block.number, log.into()) {
+                error!(
+                    target: "supervisor_storage",
+                    block_number = block.number,
+                    error = ?e,
```
emhane, line 75:
```suggestion
            if let Err(err) = cursor.append_dup(block.number, log.into()) {
                error!(
                    target: "supervisor_storage",
                    block_number = block.number,
                    ?err,
```
nitpick, this will use the variable name err as key so to stdout it will print `err=<the error>`

crates/supervisor/storage/src/log/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+mod mdbx;
+
+use kona_protocol::BlockInfo;
+use kona_supervisor_types::Log;
+
+/// Provides read-only access to log and block metadata stored in the supervisor database.
+///
+/// This trait is used by components that need to inspect the state of execution logs
+/// and block metadata without modifying storage.
+#[allow(dead_code)]
```
emhane, line 17:
```suggestion
#[expect(dead_code)]
```
`expect` will emit a warning when the lint expectation is no longer fulfilled, this helps us maintain the code so we don't leave redundant `allow` attributes in the codebase

crates/supervisor/storage/src/log/mod.rs
```diff
@@ -0,0 +1,65 @@
...
+pub(crate) trait LogStorageReader {
```
emhane, line 18: can this be public too? it doesn't have to be exported from crate root so that in itself is an indication that it is not intended for use in other crates as long as the `log` module is public

emhane (commented):
super nice - very nice docs, thanks a lot ! remaining some nitpicks

emhane (approved):

crates/supervisor/storage/src/log/mdbx.rs
emhane: ok in that case makes sense to rename the variable to `error`

crates/supervisor/storage/src/error.rs
```diff
@@ -0,0 +1,20 @@
...
+use thiserror::Error;
+
+/// Errors that may occur while interacting with supervisor log storage.
+///
+/// This enum is used across all implementations of the Storge traits.
+#[derive(Debug, Error)]
+pub enum StorageError {
+    /// DatabaseError
+    #[error("Database error")]
+    Database(#[source] DatabaseError),
```
dhyaniarun1993, line 11:
```suggestion
   #[error(transparent)]
   Database(#[from] DatabaseError),
```

crates/supervisor/storage/src/models/log.rs
```diff
@@ -32,20 +111,93 @@ pub struct LogEntry {
...
     pub hash: B256,
 }
 
+/// Compact encoding for [`ExecutingMessageEntry`] used in log storage.
+///
+/// This format ensures deterministic encoding and lexicographic ordering by
+/// placing `log_index` first, which is used as the subkey in dup-sorted tables.
+///
+/// ## Encoding Layout (ordered):
+/// - `log_index: u32` – Subkey for dup sort ordering.
```
dhyaniarun1993, line 134: I think there is a typo. This isn't the subKey. Right?
This is the initiating message identifier. SubKey would be `LogEntry.index`.

crates/supervisor/storage/src/models/log.rs
```diff
@@ -32,20 +111,93 @@ pub struct LogEntry {
...
+/// - `chain_id: u64`
+/// - `block_number: u64`
+/// - `timestamp: u64`
+/// - `hash: B256` – 32-byte message hash.
+impl Compact for ExecutingMessageEntry {
```
dhyaniarun1993, line 139: Can you add some test cases for this?

crates/supervisor/types/src/message.rs
```diff
@@ -0,0 +1,17 @@
+use alloy_primitives::B256;
+
+/// A parsed executing message extracted from a log emitted by the
+/// `CrossL2Inbox` contract on an L2 chain.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct ExecutingMessage {
```
dhyaniarun1993, line 6: Can we use `ExecutingMessage` from `kona-interop`?

crates/supervisor/storage/src/providers/log_provider.rs
```diff
@@ -0,0 +1,383 @@
...
+
+impl<TX> LogProvider<'_, TX>
+where
+    TX: DbTxMut + DbTx,
+{
+    fn store_block_logs(&self, block: &BlockInfo, logs: Vec<Log>) -> Result<(), StorageError> {
+        debug!(target: "supervisor_storage", block_number = block.number, "Storing logs");
+
+        if let Ok(latest_block) = self.get_latest_block() {
+            if latest_block.number + 1 != block.number || latest_block.hash != block.parent_hash {
```
dhyaniarun1993, line 52: I have introduced `is_parent_of` method in `BlockInfo` for this in my PR.

crates/supervisor/storage/src/providers/log_provider.rs
```diff
@@ -0,0 +1,383 @@
...
+                    target: "supervisor_storage",
+                    latest_block = ?latest_block,
+                    incoming_block = ?block,
+                    "Incoming block does not follow latest stored block"
+                );
+                return Err(StorageError::ConflictError(
+                    "Incoming block does not follow latest stored block".into(),
+                ));
+            }
+        }
```
dhyaniarun1993, line 63: Shouldn't we check for genesis?

If we think that it will be decided later on who will be responsible for this check and we can add a todo `todo: analyze if we should check for genesis block`

crates/supervisor/storage/src/providers/log_provider.rs
```diff
@@ -0,0 +1,383 @@
...
+                target: "supervisor_storage",
+                block_number = block_number,
+                error = ?e,
+                "Failed to read block",
+            );
+            StorageError::Database(e)
+        })?;
+
+        let block = block_option.ok_or_else(|| {
+            warn!(target: "supervisor_storage", block_number = block_number, "Block not found");
```
dhyaniarun1993, line 108:
```suggestion
            warn!(target: "supervisor_storage", block_number, "Block not found");
```

crates/supervisor/storage/src/providers/log_provider.rs
sadiq1971: So, the current scenario is if there is no block in db it will insert what is being sent at first.
I am not sure everytime we need to start from genesis or not. Or we can also start from a specific height. That's why the check is not here in db level.

dhyaniarun1993: Ok. I will check this at the chaindb issue.

crates/supervisor/storage/src/providers/log_provider.rs
dhyaniarun1993: I have initiated it.

crates/supervisor/storage/src/providers/log_provider.rs
```diff
@@ -0,0 +1,385 @@
...
+#[allow(dead_code)]
+impl<TX> LogProvider<'_, TX>
+where
+    TX: DbTxMut + DbTx,
+{
+    fn store_block_logs(&self, block: &BlockInfo, logs: Vec<Log>) -> Result<(), StorageError> {
+        debug!(target: "supervisor_storage", block_number = block.number, "Storing logs");
+
+        if let Ok(latest_block) = self.get_latest_block() {
+            if block.is_parent_of(&latest_block) {
```
dhyaniarun1993, line 53: This looks incorrect
```suggestion
            if !latest_block.is_parent_of(&block) {
```

dhyaniarun1993 (approved):

PR #1831, opened 2025-05-21
test(node/e2e): extend the p2p tests

### LINKED ISSUE #1462: Deterministic integration tests for the p2p discovery/gossip layer
## Description

We should start testing the `kona-node` discovery/gossip layer capabilities using the `devnet-sdk` and kurtosis.

- [x] #1823 
- [ ] #1831

theochap:
## Description

This PR extends the P2p tests for the kona-node by adding a large network test and a peer address check (to ensure that the connected nodes are all known within a kurtosis network).

With that done, we are effectively testing the ability of p2p nodes to connect to each other and check that they are part of the gossip mesh.

This PR also extends the CI tasks to add a separate large network test.

Closes #1462


clabby (commented):
This looks like a great test, but that large configuration is going to absolutely kill our CI runners. I'll reach out to platforms to see if we can get some self hosted runners for these kinds of jobs.

clabby (approved):

PR #1832, opened 2025-05-21
feat(docker): `kona-node` recipe

clabby:
## Overview

Adds an initial recipe within `docker` that launches a `kona-node` + `op-reth` instance, with an accompanying `prometheus` server to scrape metrics and `grafana` instance to host visualizations.


theochap (approved):
Looks great!

PR #1833, opened 2025-05-21
feat(node/p2p): streaming engine state through websockets

### LINKED ISSUE #1825: feat(node/rpc): add a websocket subscription http endpoint to follow unsafe/safe block gossip.
## Description

In order to increase the test coverage for the `kona-node` we should be able to follow the stream of `unsafe`/`safe` block gossips incoming from the `L1` and the other nodes. This would allow us to test that the node can effectively follow the safe head and synchronize correctly inside kurtosis.

## Proposal

To be able to effectively track the state of safe/unsafe block gossip, we could add a `RPC` websocket endpoint that allows external actors to subscribe to the blocks received by the node. We could add a parameter/flag inside the cli to make this behavior optional (we may not want to enable that in-production)

theochap:
## Description

This PR adds a couple of websocket endpoints to stream the sync state of the `kona-node`. These endpoints can be useful inside our e2e testing suite to ensure that the node is correctly syncing.

## Follow-ups

Maybe we should provide a configuration flag to make those RPC endpoints optional. We may not want to be able to subscribe to the current node state. Here is a ticket to track that #1842

Close #1825

### DIFF
--- a/crates/node/engine/src/query.rs
+++ b/crates/node/engine/src/query.rs
@@ -17,7 +17,7 @@ pub type EngineQuerySender = tokio::sync::mpsc::Sender<EngineQueries>;
 pub enum EngineQueries {
     /// Returns the rollup config.
     Config(Sender<RollupConfig>),
-    /// Returns L2 engine state information.
+    /// Returns current L2 engine state information.
     State(Sender<EngineState>),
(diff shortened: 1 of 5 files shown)


clabby (approved):
Neat!

PR #1835, opened 2025-05-22
feat(supervisor/storage): derivation provider

### LINKED ISSUE #1828: feat(supervisor/storage): derivation provider
# Description 

We need to implement the DerivationProvider for both read-only `DbTx` and read-write `DbTxMut` transaction types. This will enable safe and ergonomic access to derived block pair data within database transactions

dhyaniarun1993:
Closes #1828


crates/supervisor/storage/src/error.rs
```diff
@@ -0,0 +1,21 @@
...
+/// This enum is used across all implementations of the Storge traits.
+#[allow(dead_code)]
+#[derive(Debug, Error)]
+pub enum StorageError {
+    /// Represents a database error that occurred while interacting with storage.
+    #[error("Database error")]
+    Database(#[source] DatabaseError),
+
+    /// The expected entry was not found in the database.
+    #[error("Entry not found: {0}")]
```
emhane, line 15:
```suggestion
    #[error("entry not found: {0}")]
```
> Error messages are typically concise lowercase sentences without trailing punctuation

https://doc.rust-lang.org/std/error/trait.Error.html

crates/supervisor/storage/src/models/block.rs
```diff
@@ -28,6 +29,36 @@ pub struct BlockRef {
     pub time: u64,
 }
 
+/// Converts from [`BlockInfo`] (external API format) to [`BlockRef`] (storage
+/// format).
```
emhane, line 33: nice helpful docs

crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -0,0 +1,181 @@
...
+        Self { tx }
+    }
+}
+
+impl<TX> DerivationProvider<'_, TX>
+where
+    TX: DbTx,
+{
+    /// Helper to get [`StoredDerivedBlockPair`] by block number.
+    #[allow(dead_code)]
```
emhane, line 33:
```suggestion
    /// Helper to get [`StoredDerivedBlockPair`] by block number.
    #[expect(dead_code)]
```
expect attribute makes sure a warning is emitted by compiler when this code is not dead code, so we know when we can remove the attribute

crates/supervisor/storage/src/providers/mod.rs
```diff
@@ -0,0 +1,5 @@
+mod derivation_provider;
+
+// todo: remove this when the provider is used
+#[allow(unused_imports)]
```
emhane, line 4:
```suggestion
#[expect(unused_imports)]
```
also here expect is helpful

emhane (approved):
nitpicks

crates/supervisor/storage/src/models/block.rs
```diff
@@ -28,6 +29,36 @@ pub struct BlockRef {
...
+impl From<BlockRef> for BlockInfo {
+    fn from(block: BlockRef) -> Self {
+        Self {
+            number: block.number,
+            hash: block.hash,
+            parent_hash: block.parent_hash,
+            timestamp: block.time,
+        }
+    }
+}
```
itschaindev, line 60: Would we need this if I remove the BlockRef and use BlockInfo as you suggested in the comment on my pr?

dhyaniarun1993: This is a different struct. We introduced BlockRef specifically for use with the database. I created it to satisfy the Compact trait required by the database layer. We didn’t want to pull in the reth_db::Compact dependency into the protocol crate, so we defined a separate struct just for that purpose.

itschaindev (approved):

sadiq1971 (approved):

PR #1836, opened 2025-05-22
Supervisor handling node updates via subscription

### LINKED ISSUE #1681: feat(supervisor/core): Build l2-subscriber
# Description

Build a l2 subscriber that subscribe to the events coming from the op-node. These events will work as trigger for the supervisor

itschaindev:
Closes #1681

Things that are still left:

- Call relevant DB functions on event notification
- Check subscription with actual node
- Fallback to HTTP polling using `PullEvents` if subscription does not work


crates/supervisor/core/src/types.rs
```diff
@@ -0,0 +1,192 @@
...
+//!
+//! This module defines the data structures used for communicating between the supervisor
+//! and the op-node components in the rollup system. It includes block references,
+//! block seals, derivation events, and event notifications.
+
+use alloy_eips::BlockId;
+use alloy_primitives::{B256, U64};
+use op_alloy_consensus::OpReceiptEnvelope;
+use serde::{Deserialize, Serialize};
+
```
dhyaniarun1993, line 11:
```suggestion
// todo:: Determine appropriate locations for these structs and move them accordingly.
// todo:: Link these structs to the spec documentation after the related PR is merged.
```

crates/supervisor/core/src/syncnode/traits.rs
```diff
@@ -0,0 +1,64 @@
...
+
+/// Errors that can occur during node subscription and synchronization operations.
+#[derive(Debug, Error)]
+pub enum SubscriptionError {
+    /// Error related to JSON Web Token handling
+    #[error("JWT error: {0}")]
+    JwtError(String),
+
+    /// Authentication failures when connecting to the node
+    #[error("Authentication error: {0}")]
```
emhane, line 18: lower case error messages pls, see how std lib does it

crates/supervisor/core/src/syncnode/traits.rs
```diff
@@ -0,0 +1,64 @@
+//! Traits for synchronization node functionality.
+//!
+//! This module defines the core traits that represent the functionality of a synchronization
+//! node in the system. It includes error types and interfaces for node synchronization,
+//! control operations, and event subscription.
+
+use thiserror::Error;
+// use crate::types::{BlockRef, ManagedEvent};
```
emhane, line 8: pls remove if not used, add back later in pr that uses it

crates/supervisor/core/src/types.rs
```diff
@@ -0,0 +1,192 @@
...
+    ) -> Self {
+        Self {
+            reset,
+            unsafe_block,
+            derivation_update,
+            exhaust_l1,
+            replace_block,
+            derivation_origin_update,
+        }
+    }
```
emhane, line 191: remove in favour of `derive_more::Constructor`

crates/supervisor/core/src/types.rs
```diff
@@ -0,0 +1,192 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct BlockReplacement<T = BlockRef> {
+    /// The block that replaces the invalidated block
+    pub replacement: T,
+    /// Hash of the block being invalidated and replaced
+    pub invalidated: B256,
+}
+
+impl<T> BlockReplacement<T> {
+    /// Creates a new block replacement.
```
dhyaniarun1993, line 115:
```suggestion
    /// Creates a new [`BlockReplacement`].
```

emhane (changes requested):

crates/supervisor/core/src/types.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// The state root hash
+    pub state_root: B256,
+    /// Storage root of the message passer contract
+    pub message_passer_storage_root: B256,
+    /// The block hash
+    pub block_hash: B256,
+}
+
+impl OutputV0 {
+    /// Creates a new OutputV0 instance.
```
dhyaniarun1993, line 134:
```suggestion
    /// Creates a new [`OutputV0`].
```

crates/supervisor/core/src/types.rs
```diff
@@ -0,0 +1,192 @@
...
+
+    /// Confirms a successful block replacement operation
+    pub replace_block: Option<BlockReplacement>,
+
+    /// Indicates an update to the derivation origin
+    pub derivation_origin_update: Option<BlockRef>,
+}
+
+impl ManagedEvent {
+    /// Creates a new ManagedEvent with the specified fields.
```
dhyaniarun1993, line 174:
```suggestion
    /// Creates a new [`ManagedEvent`] with the specified fields.
```

crates/supervisor/core/src/types.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// The block's timestamp
+    pub time: U64,
+    /// Reference to the L1 origin block this L2 block is derived from
+    pub l1_origin: T,
+    /// Distance to the first block of the epoch (sequence position)
+    pub sequence_number: U64,
+}
+
+impl<T> L2BlockRef<T> {
+    /// Creates a new L2 block reference.
```
dhyaniarun1993, line 74:
```suggestion
    /// Creates a new [`L2BlockRef `].
```

crates/supervisor/core/src/types.rs
```diff
@@ -0,0 +1,192 @@
...
+    /// The block's hash
+    pub hash: B256,
+    /// The block number
+    pub number: U64,
+    /// The block's timestamp
+    pub timestamp: U64,
+}
+
+impl BlockSeal {
+    /// Creates a new block seal with the given hash, number, and timestamp.
```
dhyaniarun1993, line 25:
```suggestion
    /// Creates a new [`BlockSeal `] with the given hash, number, and timestamp.
```

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -0,0 +1,552 @@
...
+};
+use std::{path::PathBuf, sync::Arc};
+use tokio::{sync::watch, task::JoinHandle};
+use tracing::{debug, error, info, warn};
+
+/// Configuration for the managed node.
+#[derive(Debug)]
+pub struct ManagedNodeConfig {
+    /// The chain ID of the L2
+    pub chain_id: U256,
```
dhyaniarun1993, line 21: We don't know the chain id upfront. I don't think it will be part of the config.

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -0,0 +1,552 @@
...
+    /// The URL of the managed node
+    pub url: String,
+    /// The port of the managed node
+    pub port: u16,
```
dhyaniarun1993, line 25: I think we can just have a single field for the endpoint. CLI is also taking url as a single endpoint.
```suggestion
    pub endpoint: String,
```

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -0,0 +1,552 @@
...
+    /// The path to the JWT token for the managed node
+    pub jwt_path: Option<PathBuf>,
```
dhyaniarun1993, line 27: Based on cli, it will be a string

crates/supervisor/core/src/syncnode/node.rs
itschaindev: would this be part of the config then? i mean how will we know which l2's to initialize? if we have the url, we would also have the chainId, no?

dhyaniarun1993: We will initialize the node and then get the chain_id using [rpc](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/interop/managed/api.go#L58). 

Core module interacting with the managed node would need the chain_id.

emhane (approved):
lgtm

dhyaniarun1993 (approved):

PR #1857, opened 2025-05-24
test(node/sync-e2e): adding e2e safe head sync tests

theochap:
## Description

This PR adds a test to ensure that nodes sync properly the safe heads and have a consistent view of the state.
For each kona-node, this test:
- Gathers the node's safe head gossip over 10sec
- Checks that all the other nodes that have seen all these safe heads have a consistent view with the gossiped safe heads.

For now, we only have positive tests so we're lacking some coverage (what happens with rogue nodes, network failures, reorgs, ...). These cases will come once we adapt the integration testing suite to use `in-memory` mode.

### DIFF
--- a/crates/protocol/protocol/src/block.rs
+++ b/crates/protocol/protocol/src/block.rs
@@ -23,12 +23,10 @@ pub struct BlockInfo {
     /// The block hash
     pub hash: B256,
     /// The block number
-    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub number: u64,
     /// The parent block hash
     pub parent_hash: B256,
     /// The block timestamp
-    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub timestamp: u64,
 }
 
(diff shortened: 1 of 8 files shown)


crates/protocol/protocol/src/block.rs
```diff
@@ -23,12 +23,10 @@ pub struct BlockInfo {
     /// The block hash
     pub hash: B256,
     /// The block number
-    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
     pub number: u64,
     /// The parent block hash
     pub parent_hash: B256,
     /// The block timestamp
-    #[cfg_attr(feature = "serde", serde(with = "alloy_serde::quantity"))]
```
theochap, line 31: It seems that the `op-node` rpc interface actually _really expects_ `u64` instead of the alloy-serde format. See https://github.com/theochap/optimism/blob/265a69ea8fb6538bd5e15cd6d80351afe10c2798/op-service/eth/id.go#L33-L40

tests/devnets/large-kona.yaml
```diff
@@ -1,17 +1,18 @@
...
       - el_type: op-geth
         cl_type: op-node
-        count: 1
+        count: 2
       - el_type: op-reth
         cl_type: kona-node
         # Note: we use the local image for now. This allows us to run the tests in CI pipelines without publishing new docker images every time.
         cl_image: "kona-node:local"
-        count: 2
+        count: 8
```
clabby, line 15: Why such a large increase of participants?

clabby (approved):
Test looks good! Only curious about the large increase in participants on the network; Could be pretty heavy to run in CI.

clabby:
Also need to update the serialization tests for the `BlockInfo` type.

tests/devnets/large-kona.yaml
theochap: After chatting offline we settled on 4 kona-nodes for the network. I believe it will help a more realistic test configuration than 1 op-node and 2 kona-nodes.

PR #1858, opened 2025-05-24
feat(node): L2 finalization routine

### LINKED ISSUE #1693: feat(node/service): Finality Actor
### Description

Tracking issue for a new `FinalityActor`.

The `FinalityActor` is comparable to the `op-node`'s [`finalizer`](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/finality/finalizer.go).

This ticket can be closed once all subtickets are complete.

### LINKED ISSUE #1694: feat(node/service): Finality Actor Setup
### Description

Create a templated `FinalityActor` in the `kona-node-service` crate.

The finality actor should be composed of the following.
- A channel that receives finalized heads from the [`L1WatcherRpc`](https://github.com/op-rs/kona/blob/main/crates/node/service/src/actors/l1_watcher_rpc.rs#L21).
- A channel to send `L2BlockInfo` of the finalized L2 block.
- An l1 provider to validate l1 info matches.
- The cancellation token.

### LINKED ISSUE #1695: feat(node/service): Attempt to Finalize Event
### Description

Listen to the finalized L1 channel and perform the following.

1. Ensure that the new finalized head is at a higher block height than previously seen.
2. Attempt to finalize the L2
    - Get the highest L2 finality data that has the l1 origin equal to the finalized l1 block info.
    - Validate the finalized L1 by fetching the block info and verifying the hash. (see [`op-node`](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/finality/finalizer.go#L227))
    - Fetch the l1 block info for the finalized l2 block info and verify the hash. (see [`op-node`](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/finality/finalizer.go#L239))
    - Send the finalize `L2BlockInfo` out through the sender channel.

### LINKED ISSUE #1696: feat(node/service): Receive Safe L2 Block Info
### Description

In the `FinalityActor`'s main `tokio::select!` in the `execute` method, receive safe `L2BlockInfo` from the `EngineActor`. This branch is functionally equivalent to the [`op-node`'s `onDerivedSafeBlock`](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/finality/finalizer.go#L255) method.

It should append the `L2BlockInfo` to a buffer of safe l2 blocks. This is buffer is used to get the `L2BlockInfo` for the given `BlockInfo` origin (`l1_origin`) when the `FinalityActor` receives notification of a finalized l1 head (`BlockInfo`).

### LINKED ISSUE #1698: feat(node/service): Handle Idle Derivation
### Description

The `op-node`'s `finalizer` [attempts to finalize l2 blocks](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/finality/finalizer.go#L146C14-L147) if derivation goes idle based on a finality delay.

Support this logic in the `FinalityActor`.

clabby:
## Overview

Adds a path for finalizing L2 blocks, based off of the L1 block they were derived from. To accomplish this:
1. The derivation pipeline now attaches the L1 block that the L2 block was derived from, within `OpAttributesWithParent`.
2. The DA watcher actor now watches a stream of finalized L1 blocks, polled every 1m.
3. The engine actor now holds onto a map of `l1_block_number => highest_l2_block_in_epoch`, saturated when it receives attributes from the derivation actor.
    - When a new finalized L1 block is observed by the engine actor, it will find the highest L2 block whose batch data is contained within the finalized L1 chain, and finalize it (if it knows of any.)

This approach is different than suggested by the tickets, but results in a more slim outcome that takes advantage of existing actors.

### Periphery changes

- Some refactoring of the `OpAttributesWithParent` type.
- The sync start algorithm now stops safe block traversal at the finalized head, to ensure it never assigns a safe block past it.
- The `L1WatcherRpc` now uses a custom `BlockStream`, since alloy's `watch_blocks` stream doesn't dedup nor allow for observing anything other than head block updates.

### Result

Finalized blocks are correctly streaming in 😄 

Screenshot 2025-05-24 at 5 09 01 PM

### Meta

closes #1693 
closes #1694 
closes #1695 
closes #1696 
closes #1698


clabby:
## 📚 PR Stack

- https://github.com/op-rs/kona/pull/1859
- https://github.com/op-rs/kona/pull/1858
- `main`

_This comment was auto-generated_

crates/node/service/src/actors/engine.rs
```diff
@@ -108,7 +123,13 @@ impl EngineActor {
             }
         }
 
+        // Clear the queue of attributes awaiting finalization. It will be re-saturated following
+        // derivation.
+        self.awaiting_finalization.clear();
```
theochap, line 128: Question: do we really need to reset the engine task queue when the engine resets? Can't we use an optimistic approach and keep the tasks around and they will be naturally rejected if they don't build on the canonical chain anymore?

crates/node/service/src/actors/engine.rs
```diff
@@ -56,8 +58,16 @@ pub struct EngineActor {
...
+    /// A channel to receive finalized block updates.
+    finalized_block_rx: UnboundedReceiver<BlockInfo>,
     /// The cancellation token, shared between all tasks.
     cancellation: CancellationToken,
+
+    /// A map of L1 block number -> highest derived L2 block number within the L1 epoch, used to
+    /// track derived attributes awaiting finalization. When a new finalized L1 block is
+    /// received, the highest L2 block whose inputs are contained within the finalized L1 chain
+    /// is finalized.
+    awaiting_finalization: BTreeMap<u64, u64>,
```
theochap, line 70: Mini nit, but we can use type aliases to improve the readability of what objects we're mapping in this BTreeMap.

crates/node/service/src/actors/l1_watcher_rpc.rs
```diff
@@ -19,20 +23,21 @@ use tokio::{
...
     /// The L1 provider.
     l1_provider: RootProvider,
     /// The latest L1 head sent to the derivation pipeline and watcher. Can be subscribed to, in
     /// order to get the state from the external watcher.
     latest_head: tokio::sync::watch::Sender<Option<BlockInfo>>,
-    /// The outbound event sender.
+    /// The outbound L1 head block sender.
     head_sender: UnboundedSender<BlockInfo>,
+    /// The outbound L1 finalized block sender.
+    finalized_sender: UnboundedSender<BlockInfo>,
```
theochap, line 40: This is independent of this PR, but I am wondering if we should not use a mpsc with back pressure instead of an unbounded sender for this field (and the one above). Not urgent but maybe we should open a ticket to think about it.

crates/node/service/src/actors/l1_watcher_rpc.rs
```diff
@@ -65,44 +72,15 @@ impl L1WatcherRpc {
...
-        // handled downstream by receivers of the head update signal.
-        let head_block_info = BlockInfo {
-            hash: block.header.hash,
-            number: block.header.number,
-            parent_hash: block.header.parent_hash,
-            timestamp: block.header.timestamp,
-        };
-
-        Ok(head_block_info)
-    }
```
theochap, line 104: Was that dead code?

crates/node/service/src/actors/l1_watcher_rpc.rs
```diff
@@ -243,19 +222,69 @@ impl NodeActor for L1WatcherRpc {
...
+/// Note that this stream is not guaranteed to be contiguous. It may miss certain blocks, and
+/// yielded items should only be considered to be the latest block matching the given
+/// [`BlockNumberOrTag`].
+struct BlockStream<'a> {
+    /// The inner [`RootProvider`].
+    l1_provider: &'a RootProvider,
+    /// The block tag to poll for.
+    tag: BlockNumberOrTag,
+    /// The poll interval (in seconds).
+    poll_interval_secs: u64,
```
theochap, line 236: Nit: can this be a duration, and can we remove the `_secs` suffix, so that we can easily change the unit scale (if we realize we want subsecs instead)?

crates/node/service/src/actors/l1_watcher_rpc.rs
```diff
@@ -243,19 +222,69 @@ impl NodeActor for L1WatcherRpc {
...
+
+impl<'a> BlockStream<'a> {
+    /// Creates a new [`BlockStream`] instance.
+    ///
+    /// ## Panics
+    /// Panics if the passed [`BlockNumberOrTag`] is of the [`BlockNumberOrTag::Number`] variant.
+    fn new(l1_provider: &'a RootProvider, tag: BlockNumberOrTag, poll_interval_secs: u64) -> Self {
+        if matches!(tag, BlockNumberOrTag::Number(_)) {
+            panic!("Invalid BlockNumberOrTag variant - Must be a tag");
+        }
```
theochap, line 247: Not blocking for this PR, but we could avoid that if we actually had a struct for the allowed block tags and only two variants in the block number or tag enum. We can take care of that in a follow up but panicking here may be a bit risky

crates/node/service/src/actors/l1_watcher_rpc.rs
theochap: My reasoning: if the engine or the other actors that are supposed to be subscribed to these channels are too far away behind, we could reset them, which could help with the recovery.

theochap (approved):
Looks great! Most of the comments are nits or can be treated as follow ups. Going to approve to not block it. It would be nice to at least have tickets for the points that would deserve some more attention in your opinion!

crates/node/service/src/actors/l1_watcher_rpc.rs
clabby: Now it is, yeah. We used to use it, now the `BlockStream` handles it.

crates/node/service/src/actors/engine.rs
clabby: This isn't the engine task queue, but just the map of L1 blocks -> the highest L2 block in each present L1 block's epoch. But yeah, we also clear the engine task queue.

It's important to clear this, at least, since an L1 reorg could have caused the L1 block which a payload was derived within to disappear. For the tasks, we might be able to hold onto them, but it's not _too_ important. For example - we'd receive a new unsafe block anyways, and the EL would backfill, and for derived blocks (`Consolidate` tasks), we'd just re-derive them.

crates/node/service/src/actors/l1_watcher_rpc.rs
clabby: We can explore that at some point - especially keen to change this sort of thing over if we've got metrics informing that it's beneficial :D

crates/node/service/src/actors/l1_watcher_rpc.rs
clabby: Yeah that'd be nice. We're using `alloy-rpc-types`' [`BlockNumberOrTag`](https://docs.rs/alloy-rpc-types/latest/alloy_rpc_types/enum.BlockNumberOrTag.html) - maybe let's make an issue there to add a `BlockTag` enum?

crates/node/service/src/actors/engine.rs
clabby: `u64` is sort of ubiquitously used across `alloy`, `reth`, etc. for block numbers; I don't disagree, but let's stick with this for now. It'd be a pretty large refactor to add consistent type alias usage for block numbers, etc.

PR #1861, opened 2025-05-25
test(node/e2e-sync): adding e2e sync tests for unsafe/finalized sync

theochap:
## Description

This PR adds a couple more tests for unsafe head/finalized head synchronization

### DIFF
--- a/crates/node/engine/src/task_queue/tasks/insert/task.rs
+++ b/crates/node/engine/src/task_queue/tasks/insert/task.rs
@@ -4,16 +4,20 @@ use crate::{
     EngineClient, EngineForkchoiceVersion, EngineState, EngineTaskError, EngineTaskExt,
     InsertUnsafeTaskError, Metrics,
 };
+use alloy_eips::eip7685::EMPTY_REQUESTS_HASH;
 use alloy_provider::ext::EngineApi;
 use alloy_rpc_types_engine::{
-    ExecutionPayloadInputV2, ForkchoiceState, INVALID_FORK_CHOICE_STATE_ERROR, PayloadStatusEnum,
+    CancunPayloadFields, ExecutionPayloadInputV2, ForkchoiceState, INVALID_FORK_CHOICE_STATE_ERROR,
+    PayloadStatusEnum, PraguePayloadFields,
 };
 use async_trait::async_trait;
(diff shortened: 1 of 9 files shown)


clabby (approved):
Tests here look good. Not sure if the implementation is faulty, but e2e tests look to be flaking. From the logs it looks like an issue with the devnet environment.

crates/node/engine/src/task_queue/tasks/insert/task.rs
```diff
@@ -99,19 +133,14 @@ impl EngineTaskExt for InsertUnsafeTask {
...
-            .map_err(InsertUnsafeTaskError::FromBlockError)?;
         let new_unsafe_ref =
             L2BlockInfo::from_block_and_genesis(&block, &self.rollup_config.genesis)
                 .map_err(InsertUnsafeTaskError::L2BlockInfoConstruction)?;
 
+        let head_block_hash = new_unsafe_ref.block_info.hash;
+
         let fcu = ForkchoiceState {
-            head_block_hash: self.envelope.payload.block_hash(),
+            head_block_hash,
```
clabby, line 143: nit - this can be inlined, i.e.

```rust
let fcu = ForkchoiceState {
    head_block_hash: new_unsafe_ref.block_info.hash,
    ...
```

The `head_block_hash` var isn't used anywhere else.

docker/apps/kona_app_generic.dockerfile
```diff
@@ -8,7 +8,7 @@ FROM ubuntu:22.04 AS dep-setup-stage
 SHELL ["/bin/bash", "-c"]
 
 # Install deps
-RUN apt-get update && apt-get install -y --no-install-recommends \
+RUN apt-get update && apt-get install -y --no-install-recommends --fix-missing \
```
clabby, line 11: There's a PR up for this @ https://github.com/op-rs/kona/pull/1890 - can we keep this separate? re: the comment there about this skipping dependencies that can't be installed.

clabby (approved):
Nice :)

crates/node/engine/src/task_queue/tasks/insert/task.rs
theochap: Good catch!

PR #1862, opened 2025-05-25
fix(node/p2p): changed peer scoring to light as default

### LINKED ISSUE #1855: fix(node/p2p): set the default peer scoring level to light
## Description

We are disabling `p2p.scoring` by default since the p2p stable is not stable yet. We should think about reactivating this parameter later on.

PoulavBhowmick03:
Fixes #1855


theochap:
Good initiative, but before we are able to do that we need to stabilize the p2p gossip to hold on to connected peers. Now the peers are dropped pretty quickly

theochap:
Closing that PR until that's addressed

PR #1865, opened 2025-05-26
feat(supervisor/db): safety head reference storage  

### LINKED ISSUE #1750: feat(supervisor/state-manager): block safety storage
Spec can be found here: #1684 

### Block Safe References:**

* **Purpose:** Stores references to blocks heads considered "safe" based on different criteria (unsafe, locally safe, cross-chain safe).
* **Key Patterns:**
    * `{chain_index}:unsafe`
    * `{chain_index}:crossUnsafe`
    * `{chain_index}:localSafe`
    * `{chain_index}:crossSafe`
* **Value:** block number.: 

It's a normal key-value storage to track the head information of the chains

sadiq1971:
Closes #1750


crates/supervisor/storage/src/models/head_ref.rs
```diff
@@ -0,0 +1,143 @@
...
+            [0] => Ok(Self::Unsafe),
+            [1] => Ok(Self::LocalSafe),
+            [2] => Ok(Self::CrossUnsafe),
+            [3] => Ok(Self::Safe),
+            [4] => Ok(Self::Finalized),
+            [255] => Ok(Self::Invalid),
+            _ => Err(DatabaseError::Decode),
+        }
+    }
+}
```
emhane, line 57:
```suggestion
/// Key representing a particular head reference type.
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, PartialOrd, Ord, Serialize, Deserialize, TryFrom)]
#[try_from(repr)]
#[repr(u8)]
pub enum SafetyHeadRefKey {
    /// Latest unverified or unsafe head.
    Unsafe = 0,

    /// Head block considered safe via local verification.
    LocalSafe = 1,

    /// Head block considered unsafe via cross-chain sync.
    CrossUnsafe = 2,

    /// Head block considered safe.
    Safe = 3,

    /// Finalized head block.
    Finalized = 4,

    /// Invalid head reference.
    Invalid = u8::MAX,
}

/// Implementation of [`Encode`] for [`SafetyHeadRefKey`].
impl table::Encode for SafetyHeadRefKey {
    type Encoded = [u8; 1];

    fn encode(self) -> Self::Encoded {
        self as u8
    }
}

/// Implementation of [`Decode`] for [`SafetyHeadRefKey`].
impl table::Decode for SafetyHeadRefKey {
    fn decode(value: &[u8]) -> Result<Self, DatabaseError> {
        if value.is_empty() {
            return Err(DatabaseError::Decode)
        }
        
        data[0].try_into().map_err(DatabaseError::Decode)
    }
}
```
note on using `table::Encode` path, for somethings likes channels or encoding/cryptography, we are interested in knowing what type of channel or encoding or crypto is being handled without reading the module prelude. if the name of the trait was instead `DBEncode` or `EncodeMdbx`, we would just import those in the module prelude, but here, `Encode` by itself isn't enough information to be helpful in understanding the current scope. is it RLP encoding, or some other encoding? same with channels, you would use `mpsc::Sender`, because in the scope of a function it's necessary to know if the sender is a broadcasting, multicasting or unicasting to understand the logic fully.

on the contrary, whether the library is `std` or `no_std`, isn't really relevant in the function scope. that is more relevant in the module scope. neither is it relevant to know which runtime a program is using in the function scope, that is rather interesting, and definitely smthg we want to be able to read just by looking at, the module import prelude, not have to in-depth read the module's function logic to figure out. For example, there is never really a need to use full path in the code for example for `std::futures::Future` or `tokio::sync::mpsc::Sender`, `std::futures::Future` and `tokio::sync::mpsc` are imported in import prelude of module, and `Future` and `mpsc::Sender` are used through out the module. cc @itschaindev

emhane (changes requested):

emhane (approved):
nice!

PR #1866, opened 2025-05-26
feat(supervisor/core): Implement l2 controller

### LINKED ISSUE #1680: feat(supervisor/core): Build l2-controller
# Description

Build a l2 controller for supervising managed l2 nodes by sending control signals.

itschaindev:
Closes #1680.

- Broke down functions
- Added API calls to `ManagedNodeApi`
- Added `call_rpc()` to `ManagedNode` to for calling the API by passing the `ManagedNodeApi` function along with params.
- Added basic unit tests


crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -285,6 +274,86 @@ impl ManagedNode {
 
         Ok(())
     }
+
+    /// Calls a ManagedNodeApi method using an HTTP client with JWT authentication.
```
emhane, line 278:
```suggestion
    /// Calls a [`ManagedNodeApi`](ManagedNodeApiClient) method using an HTTP client with JWT authentication.
```

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -285,6 +274,86 @@ impl ManagedNode {
...
+    ///
+    /// # Arguments
+    /// * `f` - A closure that takes an HTTP client and returns a future with the RPC call
+    ///
+    /// # Returns
+    /// * `RpcResult<T>` - The result of the RPC call or an appropriate error
```
emhane, line 284:
```suggestion
    /// * [`RpcResult`]`<T>` - The result of the RPC call or an appropriate error
```

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -285,6 +274,86 @@ impl ManagedNode {
...
+            ErrorObject::from(ErrorCode::InternalError)
+        })
+    }
+
+    /// Creates authentication headers using JWT secret.
+    fn create_auth_headers(&self) -> RpcResult<HeaderMap> {
+        let jwt_secret = self.config.jwt_secret().ok_or_else(|| {
+            error!(target: "managed_node", "JWT secret not found or invalid");
+            ErrorObject::from(ErrorCode::ParseError)
+        })?;
```
emhane, line 313:
```suggestion
        let Ok(jwt_secret) = self.config.jwt_secret() else {
            error!(target: "managed_node", "JWT secret not found or invalid");
            return ErrorObject::from(ErrorCode::ParseError
        };
```
pattern matching syntax is preferred over chained calls

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -285,6 +274,86 @@ impl ManagedNode {
...
+
+        let mut headers = HeaderMap::new();
+        let auth_header =
+            format!("Bearer {}", alloy_primitives::hex::encode(jwt_secret.as_bytes()));
+
+        headers.insert(
+            "Authorization",
+            HeaderValue::from_str(&auth_header).map_err(|err| {
+                let err_msg = format!("Invalid authorization header: {}", err);
+                error!(target: "managed_node", ?err, %err_msg);
```
emhane, line 323:
```suggestion
                error!(target: "managed_node", %err, "Invalid authorization header");
```
why isn't it enough to log the error message once in display formatting?

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -285,6 +274,86 @@ impl ManagedNode {
...
+    /// * `client` - The authenticated HTTP client
+    /// * `f` - The RPC call closure
+    async fn execute_rpc_call<T, F, Fut>(
+        &self,
+        client: jsonrpsee::http_client::HttpClient,
+        f: F,
+    ) -> RpcResult<T>
+    where
+        F: FnOnce(jsonrpsee::http_client::HttpClient) -> Fut,
+        Fut: std::future::Future<Output = Result<T, jsonrpsee::core::ClientError>>,
```
emhane, line 344:
```suggestion
        F: FnOnce(HttpClient) -> Fut,
        Fut: Future<Output = Result<T, jsonrpsee::core::ClientError>>,
```
no need to specify full path here, import in import prelude of module

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -285,6 +274,86 @@ impl ManagedNode {
...
+    {
+        debug!(target: "managed_node", "Executing RPC call");
+
+        let result = f(client).await.map_err(|err| {
+            let err_msg = format!("RPC call failed: {}", err);
+            error!(target: "managed_node", ?err, %err_msg);
```
emhane, line 350: same as comment above, why do you need to log the error twice here, once with display formatting and once with debug formatting?

emhane (changes requested):

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -285,6 +274,86 @@ impl ManagedNode {
...
+        F: FnOnce(jsonrpsee::http_client::HttpClient) -> Fut,
+        Fut: std::future::Future<Output = Result<T, jsonrpsee::core::ClientError>>,
+    {
+        let client = self.create_authenticated_client().await?;
+        self.execute_rpc_call(client, f).await
+    }
+
+    /// Creates an authenticated HTTP client with JWT headers.
+    async fn create_authenticated_client(&self) -> RpcResult<jsonrpsee::http_client::HttpClient> {
+        let http_url = format!("http://{}", self.config.url);
```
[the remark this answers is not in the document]
itschaindev: That's why I initially had a separate port field in addition to url. But then I was unable to find where they were different. Will use the WsClient for now.

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -285,6 +219,123 @@ impl ManagedNode {
...
+        };
+
+        let mut headers = HeaderMap::new();
+        let auth_header =
+            format!("Bearer {}", alloy_primitives::hex::encode(jwt_secret.as_bytes()));
+
+        headers.insert(
+            "Authorization",
+            HeaderValue::from_str(&auth_header).map_err(|err| {
+                error!(target: "managed_node", "Invalid authorization header: {:?}", err);
```
emhane, line 266:
```suggestion
                error!(target: "managed_node", ?err, "Invalid authorization header");
```

emhane (changes requested):
logs need fixing

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -285,6 +219,123 @@ impl ManagedNode {
...
+    {
+        let client = self.get_ws_client().await.map_err(|err| {
+            error!(target: "managed_node", "Failed to get WebSocket client: {:?}", err);
+            ErrorObject::from(ErrorCode::InternalError)
+        })?;
+
+        debug!(target: "managed_node", "Executing RPC call");
+
+        let result = f(client).await.map_err(|err| {
+            error!(target: "managed_node", "RPC call failed: {:?}", err);
```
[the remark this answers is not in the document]
itschaindev: @emhane I'm confused between your code suggestion and comment. So it should be `?err`, i.e. Debug formatted?

dhyaniarun1993 (approved):

crates/supervisor/core/src/syncnode/node.rs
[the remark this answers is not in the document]
itschaindev: Wouldn't that be different from how they are used in all other parts of the code? E.g. in [node](https://github.com/op-rs/kona/blob/1fe5ca2609b7530441edd29859a4eb77d0fea0fd/crates/node/service/src/actors/engine/actor.rs#L115), they have used `?err`. 

I don't mind changing it to `%err`, just confirming if that's fine to defer from the followed pattern.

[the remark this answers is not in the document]
dhyaniarun1993: I think we should use `%err` in majority of the cases and  use`?err` in specific places where we need debug information.

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -64,73 +69,33 @@ pub struct ManagedNode {
...
-                    // Handle L1 exhaustion
-                }
+    /// Returns a reference to the WebSocket client, creating it if it doesn't exist.
+    // todo: support http client as well
+    pub async fn get_ws_client(&self) -> Result<Arc<WsClient>, ClientError> {
+        let mut ws_client_guard = self.ws_client.lock().await;
+        if ws_client_guard.is_none() {
+            let headers = self.create_auth_headers().map_err(|err| {
+                error!(target: "managed_node", %err, "Failed to create auth headers");
+                ClientError::Custom(format!("failed to create auth headers: {}", err))
```
emhane, line 83:
```suggestion
                ClientError::Custom(format!("failed to create auth headers: {ett}"))
```
nitpick, this syntax is preferred, pls check if there are other places in this pr diff where this can be fixed too

emhane (approved):

PR #1867, opened 2025-05-27
feat(node-service): Refactor `EngineActor`

clabby:
## Overview

Refactors the `EngineActor` a bit, thinning out the control loop and spreading out the module.


emhane (dismissed):

crates/node/service/src/actors/engine/actor.rs
```diff
@@ -25,31 +27,30 @@ use url::Url;
...
     engine_l2_safe_head_tx: WatchSender<L2BlockInfo>,
-    /// Handler for inbound queries to the engine.
-    inbound_queries: Option<tokio::sync::mpsc::Receiver<EngineQueries>>,
-    /// A channel to send a signal that syncing is complete.
+    /// A channel to send a signal that EL sync has completed.
     /// Informs the derivation actor to start.
-    sync_complete_tx: UnboundedSender<()>,
-    /// A way for the engine actor to signal back to the derivation actor
-    /// if a block building task produced an `INVALID` response.
+    sync_complete_tx: Option<OneshotSender<()>>,
```
theochap, line 48: Question: I probably asked that already but is there a situation where the `EL` engine falls out of sync so much (because the CL is down or we haven't received updates from the L1/other L2 nodes for a while) that it has to kickoff EL sync again? In that case, how is it going to be handled here? We just restart a new node process?

crates/node/service/src/actors/engine/actor.rs
```diff
@@ -116,35 +109,76 @@ impl EngineActor {
...
+                // by the engine api. Post-holocene, the payload is replaced by
+                // a "deposits-only" block and re-executed. At the same time,
+                // the channel and any remaining buffered batches are flushed.
+                warn!(target: "engine", ?err, "[HOLOCENE] Invalid payload, Flushing derivation pipeline.");
+                match self.derivation_signal_tx.send(Signal::FlushChannel) {
+                    Ok(_) => {
+                        debug!(target: "engine", "[SENT] flush signal to derivation actor")
+                    }
+                    Err(err) => {
+                        error!(target: "engine", ?err, "[ENGINE] Failed to send flush signal to the derivation actor.");
```
theochap, line 149: nit: it seems we have another convention to write logs here (by starting with a subtarget `[ENGINE]`/`[SENT]`/...) - I like the idea of having a "log type" discriminant like that. Maybe we could add a field for that and remove the `[ENGINE]`, ... from the log itself?
It could be a `type` field for instance.
If we decide to go ahead with it, we could add that to our CONTRIBUTIONS guidelines

crates/node/service/src/actors/engine/finalizer.rs
```diff
@@ -0,0 +1,77 @@
...
+
+        // If the highest safe block is found, enqueue a finalization task and drain the
+        // queue of all L1 blocks not contained in the finalized L1 chain.
+        if let Some((_, highest_safe_number)) = highest_safe {
+            let task =
+                EngineTask::Finalize(FinalizeTask::new(self.client.clone(), *highest_safe_number));
+            engine.enqueue(task);
+
+            self.awaiting_finalization.retain(|&number, _| number > new_finalized_l1.number);
+        }
```
theochap, line 73: nit: I would move the inner logic that enqueues the finalize task to the engine inside the `select!` statement to be more consistent with the other branches. I am a bit worried about taking a mutable reference to an external data structure inside the `.await` branch of a select statement.

theochap (commented):
Looks good, I left a handful of nits, but most of them can be addressed as follow ups

crates/node/service/src/actors/derivation.rs
```diff
@@ -13,6 +13,7 @@ use tokio::{
     select,
     sync::{
         mpsc::{UnboundedReceiver, UnboundedSender},
+        oneshot::Receiver as OneshotReceiver,
```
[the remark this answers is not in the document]
clabby: I agree. For now I added named aliases to the channels used in the engine actor. Let's follow up with a standard convention; I prefer shorter qualified paths such as `mpsc::{Sender/Receiver}`, but glad to discuss w/e is best.

crates/node/service/src/actors/engine/actor.rs
clabby: Yes, this can happen. This notification only unblocks derivation for the first time. Let's say that we don't receive any new batches nor unsafe blocks for a while (like 2h) - we'd end up kicking off EL sync again implicitly when the next payload gets inserted, but we don't necessarily need to block derivation while that's happening because it's already been inititated.

The big reason we have this EL sync state machine within `EngineState` is mainly to synchronize the initialization routine. We should definitely look for common issues there, though, once things are further along (i.e. just testing turning off the EL for a while, simulating p2p dropping off, etc. etc.)

PR #1840, opened 2025-05-22
test(supervisor): Configure supervisor workflow

emhane:
Closes https://github.com/op-rs/kona/issues/1715

PR #1876, opened 2025-05-27
fix(test/e2e-justfile): fixing e2e justfile

theochap:
## Description

#1840 broke the kurtosis e2e tests. This PR fixes it


clabby (approved):
g2g if CI passes.

Issue #1813, opened 2025-05-20
feat(node/docker): allow the use of a local version of the codebase to build the node's image.


theochap:
## Description

As of May 19th 2025, we can only pass a remote tag/commit-hash to build docker images for the `kona-node`. This is a bit cumbersome - it would be great to be able to build docker images of the node from local versions of it. 

## Course of action

This would likely require updating the `docker-bake.hcl` file to add a parameter that enables using the local image of the `kona-node` to generate a docker image.

theochap:
> can I take this?
> 

Nobody's working on it afaik!

JoshdfG:
cool. You can assign it to me then, thanks

JoshdfG:
Hello @theochap, I'm done with the present task, should I also make the asteric-repo.dockerfile & cannon-repo.dockerfile use the docker image?
If its not important then i'll create a PR right away.

theochap:
Wonderful! I don't think we need support for asterisc or cannon right away, so feel free to open the PR without them.
Thanks for taking care of it!

clabby:
> Hello [@theochap](https://github.com/theochap), I'm done with the present task, should I also make the asteric-repro.dockerfile & cannon-repro.dockerfile use the docker image? If its not important then i'll create a PR right away.

Just `kona_app_generic.dockerfile` needs this. We should be able to retain the existing behavior, but optionally, copy in the local codebase from the host machine and build that, rather than cloning the repo from the github remote. Should be able to configure this with a build argument, ideally!

JoshdfG:
Can you share your docker build command for final testing, to be sure I'm on the right path

clabby:
> Can you share your docker build command for final testing, to be sure I'm on the right path, sorry for the delay

Hey - thank you for your interest in this issue, this ended up being a bit blocking to some work we're doing and I've picked it up to get it over the line. Will re-assign to myself.

Glad to get you started on any of the other [good first issues](https://github.com/op-rs/kona/issues?q=is%3Aissue%20state%3Aopen%20label%3AM-good-first-issue) if you'd like!

PR #1877, opened 2025-05-27
feat(docker): Local builds + Cleanup

clabby:
## Overview

Adds support for building images for apps in the local repo with the generic docker application image. Also cleans up the documentation, and adds a `justfile` script for easily building local images.

~~This PR does not yet replace recipes such as `build-node`, `build-node-with-tag`, or `build-supervisor-with-tag`. That transition will be done in a follow-up, but must be merged into this before `main` since this is a breaking change to the dockerfile.~~ done w/ #1878

closes https://github.com/op-rs/kona/issues/1813

### DIFF
--- a/tests/Justfile
+++ b/tests/Justfile
@@ -2,14 +2,14 @@ SOURCE := source_directory()
 DEFAULT_DEVNET_PATH := source_directory() + "/devnets/simple-kona.yaml"
 DEFAULT_OP_PACKAGE_PATH := "github.com/ethpandaops/optimism-package"
 
-build-devnet BINARY COMMIT_TAG="" :
+build-devnet BINARY:
     #!/usr/bin/env bash
     if [ {{BINARY}} != "node" ] && [ {{BINARY}} != "supervisor" ]; then
         echo "Invalid binary specified. Must be either 'node' or 'supervisor'."
         exit 1
     fi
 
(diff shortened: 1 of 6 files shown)


tests/Justfile
```diff
@@ -9,12 +9,7 @@ build-devnet BINARY COMMIT_TAG="" :
         exit 1
     fi
 
-    export COMMIT_TAG={{COMMIT_TAG}}
-    if [ -z "{{COMMIT_TAG}}" ]; then
-        export COMMIT_TAG=$(git rev-parse HEAD)
-    fi
-
-    cd {{SOURCE}}/.. && just build-{{BINARY}}-with-tag $COMMIT_TAG
+    cd {{SOURCE}}/../docker/apps && just build-local "kona-{{BINARY}}" "{{COMMIT_TAG}}" "kona-{{BINARY}}:local"
```
theochap, line 12: According to this comment (https://github.com/op-rs/kona/pull/1877/files#diff-738654dd8e73056db0f2d5fbd55d78043b363d940b17c5ac64b7c197e1e08c96R40), do we still need to specify the `COMMIT_TAG` if we're building local images?

theochap: If not, let's remove that argument from both `just` recipes. That should simplify the github actions even more!

PR #1884, opened 2025-05-28
chore(workspace): Move `kona-macros` to `utilities`

clabby:
## Overview

Moves `kona-macros` out of the `node` crates and into `utilities`. It's not node-specific, and these macros can be used by other crates as well.

Also updates the `README.md`'s list of crates, which had gotten a bit out of date.


emhane (commented):
lgtm, @dhyaniarun1993 @itschaindev @sadiq1971 worth checking these macros out, perhaps some of them are useful to supervisor impl :)

emhane (approved):

PR #1885, opened 2025-05-28
chore(tests): Skip `node/p2p` e2e tests

clabby:
## Overview

Temporarily skips the `node/p2p` e2e tests, since they're unreliable and keeping CI red. Once these are fixed to be more reliable, we'll turn them back on in trunk.

cc @theochap


emhane (approved):
good call

PR #1887, opened 2025-05-28
feat(supervisor/core): managed node event channel

### LINKED ISSUE #1886: feat/(supervisor/core): Managed node event channel
# Description

Introduce an event channel to push node event upstream.

dhyaniarun1993:
Closes #1886


crates/supervisor/core/src/syncnode/event.rs
```diff
@@ -0,0 +1,28 @@
+use kona_interop::DerivedRefPair;
+use kona_protocol::BlockInfo;
+use kona_supervisor_types::BlockReplacement;
+
+/// Represents node events that managed nodes emits.
```
emhane, line 5:
```suggestion
/// Represents node events that [`ManagedNode`]s emits.
```
this can probably link to the helpful docs of the `ManagedNode` type thereby providing more context for whoever wants it, right, or is there some cyclical dep issue?

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -72,45 +77,75 @@ impl ManagedNode {
...
-    /// TODO: Call relevant DB functions to update the state
-    pub fn handle_managed_event(event_result: Option<ManagedEvent>) {
+    async fn handle_managed_event(
+        event_tx: &mpsc::Sender<NodeEvent>,
+        event_result: Option<ManagedEvent>,
+    ) {
         match event_result {
             Some(event) => {
-                debug!(target = "managed_node", event = ?event, "Handling ManagedEvent");
+                debug!(target: "managed_node", ?event, "Handling ManagedEvent");
```
emhane, line 86: nice, thanks

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -72,45 +77,75 @@ impl ManagedNode {
...
 
                 // Process each field of the event if it's present
                 if let Some(reset_id) = &event.reset {
-                    info!(target = "managed_node", reset_id = %reset_id, "Reset event received");
+                    info!(target: "managed_node", ?reset_id, "Reset event received");
```
emhane, line 90: any reason for debugging formatting? without a specific reason to debug format, use `%` operator for display format

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -72,45 +77,75 @@ impl ManagedNode {
...
                     // Handle reset action
                 }
 
                 if let Some(unsafe_block) = &event.unsafe_block {
-                    info!(target = "managed_node", ?unsafe_block, "Unsafe block event received");
-                    // Handle unsafe block
+                    info!(target: "managed_node", ?unsafe_block, "Unsafe block event received");
```
emhane, line 95: same here, any reason for using debug formatting? otherwise use display formatting as in suggestion. see docs for `tracing` library 
```suggestion
                    info!(target: "managed_node", %unsafe_block, "Unsafe block event received");
```

emhane (changes requested):
first pass

sadiq1971:
> > That is right, But As it seems similar So i thouth whether we can use use it across services.
> 
> That’s a thoughtful question. I think it's help us to decouple the events. Keeping them separate allows us to adapt to changes in the external API with minimal impact. NodeEvent is an internal type, which we can control and can evolve as needed for the service’s logic, logging, or API. Let me know what you think.
> 
> Even if the types are sort of similar, I think using our own internal event type is best practice for maintainability, flexibility, and robustness—especially when integrating with external systems.

Hmm. make sense

sadiq1971 (approved):

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -244,45 +248,75 @@ impl ManagedNode {
...
-    /// TODO: Call relevant DB functions to update the state
-    fn handle_managed_event(event_result: Option<ManagedEvent>) {
+    async fn handle_managed_event(
+        event_tx: &mpsc::Sender<NodeEvent>,
+        event_result: Option<ManagedEvent>,
+    ) {
         match event_result {
             Some(event) => {
-                debug!(target: "managed_node", ?event, "Handling ManagedEvent");
+                debug!(target: "managed_node", %event, "Handling ManagedEvent");
```
emhane, line 257:
```suggestion
                debug!(target: "managed_node", ?event, "Handling ManagedEvent");
```
here is a place where it makes most sense to just use the debug format, since these are debug logs, and the implementation that was done for display, Is basically the debug format, so we can just use debug format here and remove the display impl

crates/supervisor/types/src/types.rs
```diff
@@ -134,3 +136,33 @@ pub struct ManagedEvent {
...
+            parts.push(format!("derivation_origin_update: {origin}"));
+        }
+
+        if parts.is_empty() {
+            write!(f, "ManagedEvent {{ none }}")
+        } else {
+            write!(f, "ManagedEvent {{ {} }}", parts.join(", "))
+        }
+    }
+}
```
emhane, line 168: this impl is redundant because it's basically the same as the debug impl. display impl usually doesn't show the type name. so rather just a comma separated list of the values without and object delimiters or such.
> 
```
// Similarly, implement `Display` for `Point2D`.
impl fmt::Display for Point2D {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        // Customize so only `x` and `y` are denoted.
        write!(f, "x: {}, y: {}", self.x, self.y)
    }
}
```

see how it's done in these docs https://doc.rust-lang.org/rust-by-example/hello/print/print_display.html

dhyaniarun1993: Thanks for pointing that out. We can remove the outer ManagedEvent {{ ... }} wrapper and just display the fields directly, separated by commas.

I think Display implementation for `ManagedEvent` offers a practical advantage. It omits fields that are `None`, showing only the fields that are actually present in the event. This makes logs and output much more concise and focused, especially since `ManagedEvent` is a sparse struct where most fields are often None. 

Let me know what you think.

emhane: yes please leave out the type name and curly brackets

dhyaniarun1993: Thanks. I have pushed the changes

Issue #1761, opened 2025-05-14
Dedup `MetricsArgs`


dhyaniarun1993:
# Description

`MetricsArgs` is being used in multiple places. Move this to the `kona-cli` and utilize the it in the dependent sections.

### References
https://github.com/op-rs/kona/blob/main/bin/supervisor/src/flags/metrics.rs#L7
https://github.com/op-rs/kona/blob/main/bin/node/src/flags/metrics.rs#L13

JoshdfG:
Quick question, is the kona-cli located in utilities/cli/src/lib.rs ? or can you help with the path?

PR #1888, opened 2025-05-28
chore(node/supervisor):Move MetricArgs into kona-cli 

JoshdfG:
Closes https://github.com/op-rs/kona/issues/1761


crates/utilities/cli/src/lib.rs
```diff
@@ -22,3 +22,5 @@ mod prometheus;
 pub use prometheus::init_prometheus_server;
 
 pub mod sigsegv_handler;
+
+pub mod unified_metrics;
```
emhane, line 26: let's just call this module `metrics` or `metrics_args`

emhane (approved):
smol comment, otherwise lgtm. pending @clabby or @refcell

crates/utilities/cli/src/lib.rs
JoshdfG: okay i'll do that

bin/node/src/cli.rs
emhane, line 1: we need to call `init_unified_metrics` inside `Cli::run`, and then we can remove the metrics arg from all the `init_telemetry` methods, plus rename that method form `init_telemetry` to `init_logs`.

JoshdfG: Okay, I'm on it.

emhane (commented):
nice! lint check needs to pass, looks like an unused import statement that needs to be removed

JoshdfG:
> nice! lint check needs to pass, looks like an unused import statement that needs to be removed

Done, didn't notice early, Thanks

PR #1890, opened 2025-05-28
fix(dockerfile): fix the dependency installer

theochap:
## Description

Small PR that fixes the dockerfile installer. It seems that the new local builds need the `--fix-missing` flag to correctly install all the dependencies @clabby


clabby:
What is `apt` failing to install on your machine locally? Just tested it out and didn't have an issue without this.

A bit worried about `--fix-missing`, since it instructs `apt install` to skip any of the packages it can't install (whereas we want it to error if it fails to install anything.)

PR #1899, opened 2025-05-29
fix(node/e2e-sync): fix large sync test

theochap:
## Description

Small PR that fixes the kurtosis sync tests. Since the op-nodes don't support the websocket endpoint this test was always failing


theochap:
> nice. so this means the test will only run for kona-node and not for op-node, right?

Yeah, because the op node doesn't support our custom websocket endpoint!

PR #1910, opened 2025-05-29
fix(node/net): Added back the `pubpublish_rx `in the NetworkActor, adjusted the other sections

### LINKED ISSUE #1849: fix(node/net): add back unsafe block publisher
## Description
#1848 removes the `publish_rx` channel from the `NetworkActor`. We are never using the associated sender so the channel is always closed by default.

We should find a way to add that channel back and ensure that kona-nodes gossip their own vision of unsafe blocks to other nodes of the network.

PoulavBhowmick03:
Closes #1849

### DIFF
--- a/bin/node/src/flags/p2p.rs
+++ b/bin/node/src/flags/p2p.rs
@@ -21,7 +21,6 @@ use std::{
 };
 use tokio::time::Duration;
 use url::Url;
-use op_alloy_rpc_types_engine::OpNetworkPayloadEnvelope;
 
 /// P2P CLI Flags
 #[derive(Parser, Clone, Debug, PartialEq, Eq)]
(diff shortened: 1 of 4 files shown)


bin/node/src/flags/p2p.rs
```diff
@@ -374,6 +375,9 @@ impl P2PArgs {
         let mut gossip_address = libp2p::Multiaddr::from(self.listen_ip);
         gossip_address.push(libp2p::multiaddr::Protocol::Tcp(self.listen_tcp_port));
 
+        let (_publish_tx, publish_rx) = tokio::sync::mpsc::channel::<OpNetworkPayloadEnvelope>(100);
+        let publish_rx = Arc::new(publish_rx);
```
theochap, line 379: Since we drop the sender here, the receiver will always error out. To fix this issue, you would need to actually have a receiving end to it (probably inside the network to send gossip to other nodes). That is a non-trivial task

theochap (changes requested):
For this issue to be properly tackled we need to implement both the sender and the receiver's end of the channel.

refcell (approved):

PR #1915, opened 2025-05-29
fix(docker): add fix-missing to fix docker builds

theochap:
## Description

This PR adds a couple of parameters to the ubuntu installs inside kona's docker images to try to recover when package install fails.


clabby (approved):

clabby:
I still am a bit worried about `--fix-missing`, but let's merge this for now.

PR #1914, opened 2025-05-29
feat(node/e2e-tests): add a restart recipe for hot reloads of the kurtosis network

theochap:
## Description

This PR enables hot restarts of the kurtosis network by updating only the CL clients by swapping their docker image. This and #1896 greatly reduce the iteration time for local kurtosis development (from 5-10 mins to <1 min for reloading a kurtosis network).

Depends on #1914

PR #1916, opened 2025-05-29
feat(node/p2p): adding gossipsub event metrics

theochap:
## Description

This PR adds a new metric to track the events received from gossipsub
Depends on #1914 and #1915

### DIFF
--- a/crates/node/p2p/src/gossip/driver.rs
+++ b/crates/node/p2p/src/gossip/driver.rs
@@ -5,7 +5,7 @@ use discv5::Enr;
 use futures::stream::StreamExt;
 use libp2p::{
     Multiaddr, PeerId, Swarm, TransportError,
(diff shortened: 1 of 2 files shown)


emhane (approved):
ok, but not completely clear to me what distinction between `GOSSIP_EVENT` and `GOSSIPSUB_EVENT` is, perhaps the name of either can be improved

theochap:
> ok, but not completely clear to me what distinction between `GOSSIP_EVENT` and `GOSSIPSUB_EVENT` is, perhaps the name of either can be improved

That is pretty confusing indeed. We are supporting 3 types of protocols at the gossip layer - `GOSSIPSUB`, `IDENTIFY` and `PING`. The `GOSSIP_EVENT` counts the total number of messages that were sent by any of the protocols, whereas `GOSSIPSUB_EVENT` only counts the gossipsub messages

PR #1918, opened 2025-05-30
feat: make websocket optional

### LINKED ISSUE #1842: feat(node/rpc): make websocket endpoints optional
## Description

The PR https://github.com/op-rs/kona/pull/1833 adds a custom endpoint in `kona-node` that tracks the node's sync state through websockets. In production we may not want to expose those websocket endpoints to limit the networking activity of the node.

To address that, we can only expose the websocket endpoints optionally. This feature can be opted-in/out by using CLI flags.

varun-doshi:
Closes #1842

### DIFF
--- a/bin/node/src/commands/node.rs
+++ b/bin/node/src/commands/node.rs
@@ -3,7 +3,7 @@
 use alloy_rpc_types_engine::JwtSecret;
 use anyhow::{Result, bail};
 use backon::{ExponentialBuilder, Retryable};
-use clap::Parser;
+use clap::{ArgAction, Parser};
 use kona_engine::EngineKind;
 use kona_genesis::RollupConfig;
 use kona_node_service::{RollupNode, RollupNodeService};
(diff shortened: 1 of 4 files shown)


theochap (approved):
Looks good! Thanks for taking care of it

varun-doshi:
> Remove `ws_enabled` from the rollup config and add the field to the rollup node builder. In the implementation of the `ValidatorNodeService` trait on the `RollupConfig`, set the ws_enabled field in a new field on the `RpcLauncher`. The `RpcLauncher` can then handle the logic of whether ws is enabled or not.

@refcell can you check if this is in the right direction

crates/node/service/src/service/standard/builder.rs
```diff
@@ -160,6 +167,7 @@ impl RollupNodeBuilder {
             p2p_config: self.p2p_config,
             network_disabled: self.network_disabled,
             runtime_launcher,
+            ws_enabled: self.ws_enabled,
```
refcell, line 170: We can omit this from the `RollupNode` and set it on the `rpc_launcher` we build a little earlier in this function.

Take a look at like 143:
```rust
        let rpc_launcher = self.rpc_config.map(|c| c.as_launcher()).unwrap_or_default();
```

Since you added the `with_ws_enabled` method on the `RpcLauncher`, we can just set this here in the builder like so.

```rust
        let rpc_launcher = self.rpc_config.map(|c| c.as_launcher()).unwrap_or_default().with_ws_enabled(self.ws_enabled);

```

refcell (changes requested):
Very close, thanks for fixing up.

refcell (approved):
Nice work @varun-doshi

Issue #1845, opened 2025-05-23
Bump rust version


emhane:
Bump rust version to `1.87`

PR #1919, opened 2025-05-30
chore(msrv): Bump rust to version 1.87

emhane:
Closes https://github.com/op-rs/kona/issues/1845


clabby:
pref #1945 - performs the minimum bump to allow us to update the `reth` deps.

Issue #1925, opened 2025-06-02
feat(node/p2p): Gossipsub event metrics


clabby:
## Overview

Track the number of gossipsub events received in a gauge

https://github.com/op-rs/kona/blob/a9fa23c69993bbb54c8bdf9b1a0d282513bedc9b/crates/node/p2p/src/gossip/driver.rs#L320-L375

PR #1932, opened 2025-06-02
fix(p2p): Labels for Gossip Events

refcell:
### Description

Adds proper labeling for gossip event metrics.

Sets up #1925


refcell:
## 📚 $\text{Stack Overview}$

* #1932 👈
* `main`

clabby (changes requested):
Looks good for the most part, but let's remove the `total` type. We should be able to `sum_by(type)` instead of manually adding a `total`, and having it breaks `sum_by(type)`.

PR #1937, opened 2025-06-02
perf(engine): Batch safe-head FCUs

### LINKED ISSUE #1757: feat(node): Batch safe head FCUs
## Overview

Right now, every time we consolidate derived attributes, we send a forkchoice update. This is fairly inefficient, and it stacks up over time.

Rather than sending a forkchoice update for every single safe block promotion, we should take advantage of the `OpAttributesWithParent`'s `is_last_in_span` field. The `op-node` can internally promote its safe head, but it can wait until the last block in the span batch to apply the forkchoice update with the execution layer (unless the consolidate job decides a block needs to be built.)

clabby:
## Overview

Updates the consolidate task to only send a forkchoice update for safe head promotion if the attributes are either derived from a `SingleBatch`, or if the attributes are the last attributes within a `SpanBatch`.

closes #1757


refcell (approved):
Nice

PR #1938, opened 2025-06-02
feat(node/service): Add metric for critical derivation errors

### LINKED ISSUE #1628: feat(node/service): Record Critical Derivation Error
### Description

If the [`DerivationActor` experiences a critical error](https://github.com/op-rs/kona/blob/main/crates/node/service/src/actors/derivation.rs#L138-L141), it will bubble up the error. A critical error requires investigation and should produce an alert. Add a [gauge metric](https://docs.rs/metrics/0.24.2/metrics/macro.gauge.html) that records if a critical error occurs.

clabby:
## Overview

Adds a metric for critical derivation errors. This metric is primarily meant to be used for alerting purposes, since a critical failure is grounds for the node to exit.

closes #1628


refcell (approved):
Good stuff thank you

PR #1944, opened 2025-06-02
Added Duration to be tracked that peers stay connected to libp2p swarm

### LINKED ISSUE #1926: feat(node/p2p): Swarm peer connection duration
## Overview

Track the amount of time that peers stay connected to the libp2p swarm, preferably as a histogram.

Connection established @:
https://github.com/op-rs/kona/blob/33b9c4531ebcbe1404d98d8200c071b843b475cf/crates/node/p2p/src/gossip/driver.rs#L271-L283

Connection terminated @:
https://github.com/op-rs/kona/blob/33b9c4531ebcbe1404d98d8200c071b843b475cf/crates/node/p2p/src/gossip/driver.rs#L307-L314

PoulavBhowmick03:
Fixes #1926

### DIFF
--- a/crates/node/p2p/src/metrics/mod.rs
+++ b/crates/node/p2p/src/metrics/mod.rs
@@ -44,6 +44,10 @@ impl Metrics {
     /// Identifier for a histogram that tracks peer scores.
     pub const PEER_SCORES: &str = "kona_node_peer_scores";
 
+    /// Identifier for the gauge that tracks the duration of peer connections in seconds.
+    pub const GOSSIP_PEER_CONNECTION_DURATION_SECONDS: &str =
+        "kona_node_gossip_peer_connection_duration_seconds";
+
     /// Initializes metrics for the P2P stack.
     ///
     /// This does two things:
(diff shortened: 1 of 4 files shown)


crates/node/p2p/src/metrics/mod.rs
clabby, line 1: We'll also need to add a `metrics::describe_histogram!(...)` invocation in the `describe` fn here.

clabby:
> Good start, wondering if we should not try to update the histogram whenever a peer receives a ping from another peer. That would give a better feedback loop.

I agree. This would keep the histogram more representative of the connection duration of all peers at any given point.

refcell:
@PoulavBhowmick03 can you resolve conflicts here?

PoulavBhowmick03:
> @PoulavBhowmick03 can you resolve conflicts here?

Yes I'll get it done! Can you tell me what I need to do regarding the additional hook/function as mentioned

refcell (approved):
LGTM, pushed a few updates + rebase. Waiting for @clabby 's review before merging

PR #1946, opened 2025-06-03
feat(supervisor/core): managed derivation data flow

### LINKED ISSUE #1687: feat(supervisor/core): Integrate the managed derivation process
# Description

Supervisor as L1 data source for op-node. Integrate with the L1 watcher and L2 controller to manage the derivation process

itschaindev:
Closes #1687


crates/supervisor/core/src/syncnode/error.rs
```diff
@@ -46,3 +46,23 @@ pub enum SubscriptionError {
...
     SubscriptionNotFound,
 }
+
+/// Error handling managed event task.
+#[derive(Debug, Error, PartialEq)]
+pub enum ManagedEventTaskError {
+    /// Unable to successfully fetch next L1 block.
+    #[error("failed to get block by number")]
+    GetBlockByNumberFailed,
+    /// Current block hash and parent block hash do not match.
```
emhane, line 56:
```suggestion
    /// Current block hash and parent block hash of next block do not match.
```

crates/supervisor/core/src/syncnode/error.rs
```diff
@@ -46,3 +46,23 @@ pub enum SubscriptionError {
...
+    BlockHashMismatch,
+    /// This should never happen, new() always sets the rpc client when creating the task.
+    #[error("rpc client for managed node is not set")]
+    ManagedNodeClientEmpty,
+    /// Managed node api call failed.
+    #[error("managed node api call failed")]
+    ManagedNodeAPICallFailed,
+    /// Next block is either empty or unavailable.
+    #[error("next block is either empty or unavailable")]
+    NextBlockNotFound,
```
emhane, line 67:
```suggestion
    #[error("next block is either empty or unavailable, number: {0}")]
    NextBlockNotFound(u64),
```

emhane (approved):
really nice! thanks. some nitpicks remaining.

crates/supervisor/core/src/syncnode/task.rs
```diff
@@ -0,0 +1,363 @@
...
+
+/// [`ManagedEventTask`] sorts and processes individual events coming from a subscription.
+#[derive(Debug)]
+pub struct ManagedEventTask {
+    /// The URL of the L1 RPC endpoint to use for fetching L1 data
+    l1_rpc_url: String,
+    /// The channel to send the events to which require further processing e.g. db updates
+    event_tx: mpsc::Sender<NodeEvent>,
+    /// The WebSocket client to use for connecting to managed node (optional for testing)
+    client: Option<Arc<WsClient>>,
```
dhyaniarun1993, line 23:
```suggestion
    client: Arc<WsClient>,
```

crates/supervisor/core/src/syncnode/node.rs
```diff
@@ -158,12 +164,12 @@ impl ManagedNode {
                     }
 
                     // Listen for events from subscription
-                    event = subscription.next() => {
-                        match event {
-                            Some(event_result) => {
-                                match event_result {
+                    incoming_event = subscription.next() => {
```
dhyaniarun1993, line 167: We can potentially move the entire background listen to the task.

itschaindev: I was also confused if I should do that or not and I decided to keep it here because it would be good separation of concerns. The `ManagedNode` can handle all transport related stuff, which also includes the subscription part and letting the `ManagedNodeEventTask` focus on only handling the events. If we were to put the background listening task to `ManagedNodeEventTask`, it would also require to pass the subscription and handles after initiating them in `ManagedNode`, which would be not great from readability perspective.

PR #1947, opened 2025-06-03
fix(node/p2p): Gossipsub connection metric keys

clabby:
## Overview

Fixes the `kona_node_gossipsub_connection` metrics to have proper `type` labels, allowing for filtering by `type` in the grafana visualizations.


theochap (approved):
Just checking: this pr also updates the rust version. Is that intended?
Otherwise looks great

clabby:
> Just checking: this pr also updates the rust version. Is that intended? Otherwise looks great

Ah this is from #1945 - rebasing

PR #1948, opened 2025-06-03
feat(node/engine): Add `info` log for safe head promotion

clabby:
## Overview

Adds an `info` log for when the `ConsolidateTask` successfully promotes the safe head of the chain.


emhane (approved):
this would be nice as a metric too, wdyt?

Issue #1727, opened 2025-05-12
feat(supervisor/rpc): implement `supervisor_syncStatus`


dhyaniarun1993:
# Description

Implement supervisor_syncStatus RPC method.

Ref: https://github.com/ethereum-optimism/specs/blob/main/specs/interop/supervisor.md#supervisor_syncstatus

PR #1952, opened 2025-06-03
feat(supervisor/rpc): Add RPC method `supervisor_syncStatus`

emhane:
Ref https://github.com/op-rs/kona/issues/1727

Implements data processing of method in scope of `SupervisorRpc` server. Remaining is implementation in scope of `Supervisor` service, which requires `Supervisor` to be integrated with `ChainDbFactory`.


crates/supervisor/rpc/src/response.rs
```diff
@@ -0,0 +1,140 @@
...
+#[cfg_attr(
+    feature = "serde",
+    derive(serde::Serialize, serde::Deserialize),
+    serde(rename_all = "camelCase")
+)]
+pub struct SupervisorChainSyncStatus {
+    /// Highest [`Unsafe`] head of chain.
+    ///
+    /// [`Unsafe`]: op_alloy_consensus::interop::SafetyLevel::Unsafe
+    pub r#unsafe: BlockInfo,
```
dhyaniarun1993, line 41:
```suggestion
    pub local_unsafe: BlockInfo,
```

dhyaniarun1993 (commented):
Just 1 comment, otherwise Lgtm.

crates/supervisor/rpc/src/response.rs
dhyaniarun1993: If you look at the current state in the [golang implementation](https://github.com/ethereum-optimism/optimism/blob/develop/op-service/eth/supervisor_status.go#L17), it still says localUnsafe

emhane: true, `SafetyLevel` def is also updated in go impl. probably specs need updating. https://github.com/ethereum-optimism/optimism/blob/e6848b15f5dafb3159e993f7aa24844679a44e5b/op-supervisor/supervisor/types/types.go#L258-L266

emhane: opened issue https://github.com/ethereum-optimism/specs/issues/717

dhyaniarun1993 (approved):

Issue #1841, opened 2025-05-22
Rename network params file


emhane:
Rename file to stick to standard, file should be called _kurtosis_network_params_kona_node.yaml_ or _kurtosis_network_params_node.yaml_
https://github.com/op-rs/kona/blob/main/tests/devnets/simple-kona.yaml

emhane:
enclave name must be constructed in other way than from network params file path name since is not allowed to contain underscores

PR #1959, opened 2025-06-03
Rename network params file from simple-kona.yaml to kurtosis-network-params-node.yaml

JoshdfG:
closes: https://github.com/op-rs/kona/issues/1841


emhane (changes requested):
you need to update the path in the e2e-test workflow file as well to point to this "new" network params file. recommend searching for simple-kona in the search field of your IDE to make sure we don't miss updating any other places too (except CHANGELOG.md of course)

PR #1972, opened 2025-06-03
chore(rpc): Server Restarts

### LINKED ISSUE #1320: feat(node/service): Rpc Server Restart Policy
### Description

https://github.com/op-rs/kona/pull/1318 introduces an `RpcActor` in the `kona-node-service`.

Currently, when the rpc server stops, the cancellation token is triggered which will stop the entire `kona-node`.

This ticket is to create some sort of restart policy for the rpc server as opposed to triggering the cancellation token. Perhaps a retry to restart the server 3 times and the trigger the cancellation token if the server fails after 3 restarts. This ought to be enabled by default and can be overridden with a `--fail-fast` cli option.

refcell:
### Description

Allows the RPC server to be restarted up to 3 times before failing and killing the node.

Closes #1320


refcell:
## 📚 $\text{Stack Overview}$

* #1972 👈
* `main`

PR #1963, opened 2025-06-03
feat(supervisor/storage): derivation storage initialization

dhyaniarun1993:
Part of #1685 

Blocked by #1953

PR #1953, opened 2025-06-03
feat(supervisor/core): rollup config

dhyaniarun1993:
Closes #1950

Issue #1685, opened 2025-05-08
feat(supervisor/core): integrate supervisor initialisation


dhyaniarun1993:
# Description

This issue is responsible for initialising the op-supervisor state based on the flow described in the [doc](https://github.com/ethereum-optimism/design-docs/blob/main/protocol/supervisor-dataflow.md#supervisor-as-orchestrator-for-sync).

```
- Upon startup, establish a connection to the Owned Node
- If the Supervisor database is uninitialized, ask for initialization information from the Node
       - Initialization Info is returned
- Ask the node for its current latest head
- If the Supervisor is missing data for some of the data to the Head, request it
- The Supervisor may ask multiple Owned Nodes for data
- Repeat the check-and-request for data during runtime
```

Additional comments: I think initialisation information is being fetched using op-node's [Anchor Point API](https://github.com/ethereum-optimism/optimism/blob/v1.13.2/op-node/rollup/interop/managed/api.go#L42). This needs to be confirmed.

dhyaniarun1993:
Additional discussion can be found [here](https://discord.com/channels/1244729134312198194/1371428573822320640)

PR #1979, opened 2025-06-04
feat(supervisor/storage): log storage initialization

dhyaniarun1993:
Part of https://github.com/op-rs/kona/issues/1685

Based on https://github.com/op-rs/kona/pull/1963

Blocked by https://github.com/op-rs/kona/pull/1953 https://github.com/op-rs/kona/pull/1963


crates/supervisor/storage/src/providers/log_provider.rs
```diff
@@ -250,40 +287,114 @@ mod tests {
...
+        res
+    }
+
+    /// Helper to insert a pair in a new transaction, committing if successful.
+    fn insert_block_logs(
+        db: &DatabaseEnv,
+        block: &BlockInfo,
+        logs: Vec<Log>,
+    ) -> Result<(), StorageError> {
+        let tx = db.tx_mut().expect("Could not get mutable tx");
```
emhane, line 316: are we sure we want the node to go offline completely when this happens in production?

crates/supervisor/storage/src/providers/log_provider.rs
```diff
@@ -250,40 +287,114 @@ mod tests {
...
+        let provider = LogProvider::new(&tx);
+        let res = provider.store_block_logs(block, logs);
+        if res.is_ok() {
+            tx.commit().expect("Failed to commit transaction");
```
emhane, line 320: same here, are we sure we want the node to be killed in production? how about just panicking in debug mode with `debug_assert` and otherwise returning error?

emhane (changes requested):

emhane:
updated base branch to make the diff only be the relevant part to review + wrote this to pr description. when the base branch is merged it usually switches the base branch to main automatically.

crates/supervisor/storage/src/providers/log_provider.rs
dhyaniarun1993: This is part of unit testing

crates/supervisor/storage/src/providers/log_provider.rs
dhyaniarun1993: This is part of unit testing

PR #1984, opened 2025-06-04
feat(supervisor): Service initialisation

### LINKED ISSUE #1685: feat(supervisor/core): integrate supervisor initialisation

dhyaniarun1993:
Closes #1685 

Blocked on #1963 #1979 

Additional Bugfixes
- JWT authentication
- ManagedModeAPI

### DIFF
--- a/crates/supervisor/service/src/service.rs
+++ b/crates/supervisor/service/src/service.rs
@@ -43,12 +43,12 @@ impl Service {
         let mut supervisor =
             Supervisor::new(self.config.clone(), database_factory, CancellationToken::new());
 
-        supervisor.initialise().await.map_err(|e| {
+        supervisor.initialise().await.map_err(|err| {
             warn!(target: "supervisor_service",
-                %err
+                %err,
                 "Failed to initialise Supervisor"
             );
-            anyhow::anyhow!("failed to initialise Supervisor: {}", e)
+            anyhow::anyhow!("failed to initialise Supervisor: {}", err)
         })?;
 
         let supervisor = Arc::new(supervisor);


crates/supervisor/core/src/syncnode/error.rs
```diff
@@ -17,8 +17,22 @@ pub enum ManagedNodeError {
     Authentication(#[from] AuthenticationError),
 }
 
+impl PartialEq for ManagedNodeError {
+    fn eq(&self, other: &Self) -> bool {
+        use ManagedNodeError::*;
+        match (self, other) {
+            (Client(a), Client(b)) => format!("{}", a) == format!("{}", b),
```
emhane, line 24: isn't it simpler to just compare `a.to_string() == b.to_string()`? moreover, seems better if `a` and `b` would implement `PartialEq` as well instead of comparing strings

crates/supervisor/storage/src/providers/log_provider.rs
```diff
@@ -43,27 +43,51 @@ impl<TX> LogProvider<'_, TX>
...
+        if !latest_block.is_parent_of(block) {
+            warn!(
+                target: "supervisor_storage",
+                %latest_block,
+                incoming_block = %block,
+                "Incoming block does not follow latest stored block"
+            );
+            return Err(StorageError::ConflictError(
+                "incoming block does not follow latest stored block".into(),
+            ));
```
emhane, line 80: please make a new `StorageError` variant instead

tests/devnets/simple-supervisor.yaml
emhane, line 1: this seems out of scope. can we please submit this in a separate pr? incase we need to rollback smthg or simply for referencing these changes, it would be make more sense if CI changes were introduced in one cohesive pr. should also be possible to point that separate pr to the e2e testing issue as well.

emhane (changes requested):

crates/supervisor/core/src/syncnode/error.rs
emhane: should compare error codes here instead. however, can add a todo and fix this as debt since we already discussed replacing error variants wrapping strings as debt issue + match them up with spec rpc errors https://github.com/ethereum-optimism/specs/blob/main/specs/interop/supervisor.md#errors. additionally im wondering if we add a new spec section which describes the same errors as data availability errors without the context of rpc as well, since this seems to feel counterintuitive to think about rpc errors in the context of consensus.

sadiq1971 (approved):

crates/supervisor/rpc/src/jsonrpsee.rs
```diff
@@ -120,7 +132,7 @@ pub trait ManagedModeApi {
 
     /// Get the chain id
     #[method(name = "chainID")]
-    async fn chain_id(&self) -> RpcResult<ChainId>;
+    async fn chain_id(&self) -> RpcResult<String>;
```
itschaindev, line 135: Not sure why this is a String. In op-node's API everywhere it's ChainId: https://github.com/ethereum-optimism/optimism/blob/944d43d0465b04c43df6b0cc1caf12ff398d5cd8/op-node/rollup/interop/managed/api.go#L63

dhyaniarun1993: So the chainId is of type `uint256.Int` and `MarshalJSON` for `uint256.Int` is String.

emhane: possibly we can use `U256` here. will look into it and open a debt issue in that case.

PR #2008, opened 2025-06-04
feat(derive): Frame + Channel Metrics

### LINKED ISSUE #1985: feat(derive/frame): Frame Queue Size Metric
### Description

Add a metric that track the number of frames in the frame queue. This metric should be updated when frames are [loaded into the queue](https://github.com/op-rs/kona/blob/main/crates/protocol/derive/src/stages/frame_queue.rs#L160).

Add a metric that tracks the amount of memory taken by the frames in the queue in bytes.

### LINKED ISSUE #1992: feat(derive/channel): Channel Bank Size
### Description

Add a metric that tracks the number of channels in the channel bank. This may be updated when [`next_data`](https://github.com/op-rs/kona/blob/main/crates/protocol/derive/src/stages/channel/channel_bank.rs#L198) is called.

Add a metric that tracks the byte size of channels in memory.

refcell:
### Description

Adds metrics around frames and channels in the pipeline stages.

Closes #1985
Closes #1992

### DIFF
--- a/crates/protocol/derive/src/stages/frame_queue.rs
+++ b/crates/protocol/derive/src/stages/frame_queue.rs
@@ -109,6 +109,20 @@ where
 
     /// Loads more frames into the [FrameQueue].
     pub async fn load_frames(&mut self) -> PipelineResult<()> {
+        // Update metrics both before and after loading frames to ensure
+        // the gauge is _decremented_ if the queue is empty.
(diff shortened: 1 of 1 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #2008 👈
* #2007
* `main`

clabby (approved):

PR #2009, opened 2025-06-04
feat(derive): Max Channel Size and Timeout Metrics

### LINKED ISSUE #1994: feat(derive/channel): Channel Timeout Metrics
### Description

Add a metric that tracks when a channel [times out](https://github.com/op-rs/kona/blob/main/crates/protocol/derive/src/stages/channel/channel_bank.rs#L134-L143).

Add a metric that tracks the amount of time until the closest channel times out. This could be a gauge that is set to the earliest timestamp where the first channel will timeout. It monotonically increases. In grafana, the distance from the timeout can be calculated using `channel_timeout_timestamp - now()` which should give the number of seconds until the current closest channel times out.

### LINKED ISSUE #1998: feat(derive/channel): Max RLP Bytes per Channel Metric
### Description

Add a simple gauge that tracks the maximum rlp bytes allowed per channel. This is changed based on the fjord hardfork [here](https://github.com/op-rs/kona/blob/main/crates/protocol/derive/src/stages/channel/channel_reader.rs#L67-L71).

This gauge value can be composed with another metric that tracks channel byte size to visualize how far from the max threshold channel sizes are.

refcell:
### Description

Add a metric that track the max rlp byte size of the channel.

Adds a metric that tracks the amount of blocks until the next duration.

Closes #1994
Closes #1998


crates/protocol/derive/src/metrics/mod.rs
```diff
@@ -25,6 +25,12 @@ impl Metrics {
 
     /// Identifier for the channel buffer memory overhead gauge.
     pub const PIPELINE_CHANNEL_MEM: &str = "kona_derive_channel_mem";
+
+    /// Identifier for a gauge that tracks the number of blocks until the next channel times out.
+    pub const PIPELINE_CHANNEL_TIMEOUT: &str = "kona_derive_channel_timeout";
```
clabby, line 30: Perhaps we can use a more descriptive name? `kona_derive_blocks_until_channel_timeout` or something (verbose, but describes what's being tracked a bit better - I read this as a static value.)

crates/protocol/derive/src/stages/channel/channel_assembler.rs
```diff
@@ -98,6 +98,12 @@ where
         kona_macros::set!(gauge, crate::metrics::Metrics::PIPELINE_CHANNEL_BUFFER, count);
 
         if let Some(channel) = self.channel.as_mut() {
+            // Track the number of blocks until the channel times out.
+            let timeout = channel.open_block_number() + self.cfg.channel_timeout(origin.timestamp);
+            let margin =
+                if timeout < origin.number { 0_f64 } else { (timeout - origin.number) as f64 };
```
clabby, line 104: smol nit: we can use a `saturating_sub` here rather than the `if .. else`

crates/protocol/derive/src/metrics/mod.rs
refcell: Yea sure a bit picky but happy to increase verbosity

PR #2023, opened 2025-06-05
fix: managed node initialisation

dhyaniarun1993:


crates/supervisor/core/src/supervisor.rs
```diff
@@ -161,11 +161,14 @@ impl Supervisor {
...
             let chain_id = managed_node.chain_id().await?;
-            if !self.managed_nodes.contains_key(&chain_id) {
+            if self.managed_nodes.contains_key(&chain_id) {
                 warn!(target: "supervisor_service", "Managed node for chain {chain_id} already exists, skipping initialization");
                 continue;
             }
             self.managed_nodes.insert(chain_id, Arc::new(managed_node));
+            info!(target: "supervisor_service",
+                "Managed node for chain {chain_id} initialized successfully",
+            );
```
emhane, line 171:
```suggestion
            info!(target: "supervisor_service",
                 chain_id,
                "Managed node for chain initialized successfully",
            );
```

emhane (approved):

crates/supervisor/core/src/supervisor.rs
dhyaniarun1993: Thanks for pushing the fix

PR #2024, opened 2025-06-05
chore(supervisor): Remove deprecated supervisor api from `kona-node`

emhane:
Remove deprecated supervisor api from `kona-node` which isn't used since the same API is re-exported from `kona-rpc` in crate root


refcell (commented):
Lints need to be fixed, otherwise looks good

Issue #1929, opened 2025-06-02
p2p stability: at least 48 hours of no missing unsafe gossip


refcell:
Once this is achieved, [Phase 2: Unsafe Head Sync](https://github.com/kona/milestones/9) can be considered complete.

refcell:
💹

theochap:
For later reference, a combination of #2042 and #2025 fixed it

PR #2025, opened 2025-06-05
fix(node/p2p): fixing gossip stability by limiting the number of dials to a given peer

### LINKED ISSUE #1854: fix(node/p2p): remove infinite redial by default
## Description

We are temporarily enabling peers to be infinitely redialed by default. This should be removed once the P2P stack gets more stable. In the meantime, this parameter helps maintaining a larger network of peers.

theochap:
## Description

This PR fixes the gossip stability issues we've experience on initial connections to the `op-node`s by preventing `kona-node`s from redialing already connected peers, or peers that have been already dialed.

With that fix on, we're not experiencing the peer drops in kurtosis anymore

On a side note: this PR also increases the log levels of some gossip events.

## Explanation

My understanding of the situation:

- The libp2p library imposes a substream limit for outgoing connections (constant, equal to 5). This is to prevent DOS vectors when creating new connections
- It seems that every dial call will open a new outbound substream.
- Since we're dialing peers quite often in the discovery layer, we're causing the libp2p library to reach the substream limit
- When we reach the substream limit, the protocol gets disconnected and we're not advertising it to peers anymore. This causes the gossip connection to drop

Relevant piece of code in libp2p : https://github.com/libp2p/rust-libp2p/blob/d3e88cfc2ec944c3e6beb7117a762452cb855e38/protocols/gossipsub/src/handler.rs#L499-L511. In our case, the outbound event is consistently the ConnectionEvent::FullyNegotiatedOutbound event, which happens on every new dial.

## Development

- Close #1854
- Progress towards #1929

### DIFF
--- a/crates/node/p2p/src/gossip/driver.rs
+++ b/crates/node/p2p/src/gossip/driver.rs
@@ -312,7 +312,7 @@ impl GossipDriver {
                 self.peerstore.insert(peer_id, endpoint.get_remote_address().clone());
             }
             SwarmEvent::OutgoingConnectionError { peer_id, error, .. } => {
-                error!(target: "gossip", "Outgoing connection error: {:?}", error);
+                debug!(target: "gossip", "Outgoing connection error: {:?}", error);
                 kona_macros::inc!(
                     gauge,
                     crate::Metrics::GOSSIPSUB_CONNECTION,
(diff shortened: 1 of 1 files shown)


refcell (approved):
Also closes #1854 I believe

Issue #1993, opened 2025-06-04
feat(node/service): Use bounded channels


clabby:
## Overview

Currently, throughout the `kona-service` actors, we commonly use unbounded channels for communication.

**Example:**
https://github.com/op-rs/kona/blob/c0f5443c8dedb33d61bda5853cb70dbc9e0d84ca/crates/node/service/src/actors/engine/actor.rs#L51-L63

---

We should re-assess the use of unbounded channels in favor of bounded channels, in order to introduce backpressure into the system. In the event that,
for example, the L2 EL goes offline, we do not want the node to endlessly fill up a buffer that cannot be drained.

PR #2030, opened 2025-06-06
feat(node/service): Introduce backpressure in DA watcher channels

clabby:
## Overview

Introduces backpressure within the DA watcher channels. This change ensures that if the consumers of the DA watcher events (L1 {head/finalized} updates, unsafe block signer updates) do not process their events in a timely manner, the DA watcher doesn't endlessly fill up the channel queue, and yields until the other actors can process its pending updates.

progress on #1993

### DIFF
--- a/crates/node/service/src/service/validator.rs
+++ b/crates/node/service/src/service/validator.rs
@@ -15,10 +15,7 @@ use kona_rpc::{
     RpcLauncherError, WsRPC, WsServer,
 };
 use std::fmt::Display;
-use tokio::sync::{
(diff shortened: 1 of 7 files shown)


crates/node/service/src/service/validator.rs
```diff
@@ -16,7 +16,7 @@ use kona_rpc::{
 };
 use std::fmt::Display;
 use tokio::sync::{
-    mpsc::{self, UnboundedSender},
+    mpsc::{self},
```
refcell, line 19: gippity
```suggestion
    mpsc,
```

PR #2052, opened 2025-06-09
feat(node/p2p): improve dial metric granularity

theochap:
## Description

This PR improve the granularity of the `DIAL_PEER` metric by making sure we only increase the gauge when a peer is actually dialed. If that's not the case we emit the `DIAL_PEER_ERROR` metric

### DIFF
--- a/crates/node/p2p/src/net/driver.rs
+++ b/crates/node/p2p/src/net/driver.rs
@@ -160,7 +160,6 @@ impl Network {
                             return;
                         };
                         self.gossip.dial(enr);
-                        kona_macros::inc!(gauge, crate::Metrics::DIAL_PEER);
                     },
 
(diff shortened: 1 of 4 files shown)


crates/node/p2p/src/net/driver.rs
```diff
@@ -160,7 +160,6 @@ impl Network {
                             return;
                         };
                         self.gossip.dial(enr);
-                        kona_macros::inc!(gauge, crate::Metrics::DIAL_PEER);
```
refcell, line 163: How come we delete this?

refcell: Oh I see you moved it post-dial. Great

refcell (approved):
Good improvement 💯

PR #2055, opened 2025-06-09
feat(p2p): Connection Gater

refcell:
### Description

Refactors peer management into a connection gater.

Simplifies the `GossipDriver` connection gate logic through a trait.


crates/node/p2p/src/gossip/builder.rs
clabby, line 1: Fine for now; Maybe in the future we can make the builder generic over the `ConnectionGater` such that folks can supply their own?

crates/node/p2p/src/gossip/driver.rs
```diff
@@ -198,36 +171,20 @@ impl GossipDriver {
...
-        if self.current_dials.contains(&peer_id) {
-            debug!(target: "gossip", peer=?addr, "Already dialing peer, not dialing");
-            return;
-        }
-
-        if self.dial_threshold_reached(&addr) {
-            debug!(target: "gossip", peer=?addr, "Dial threshold reached, not dialing");
+        // Check if we're allowed to dial the address.
+        if !self.connection_gate.can_dial(&addr) {
+            warn!(target: "gossip", "[GATE] unable to dial peer");
```
clabby, line 176: tiny nit - rm `[GATE]`
```suggestion
            warn!(target: "gossip", "unable to dial peer");
```

crates/node/p2p/src/gossip/gater.rs
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    fn unblock_subnet(&mut self, _subnet: &str) {
+        // TODO
+    }
+
+    fn list_blocked_subnets(&self) -> Vec<String> {
+        // TODO
+        vec![]
+    }
```
clabby, line 149: What will these be used for out of curiosity?

clabby (approved):
I like this change; Nice encapsulation of concerns here.

crates/node/p2p/src/gossip/builder.rs
refcell: Yes exactly, I can write up a ticket in nice-to-have to fix later

crates/node/p2p/src/gossip/driver.rs
refcell: okie, I like these flags so it's ez to see in logs but fine to rm

PR #2058, opened 2025-06-09
chore(engine): Log finalized head updates

clabby:
## Overview

Adds an `info` log for finalized head updates.


refcell (approved):
Beautiful

PR #2080, opened 2025-06-10
feat(ci): improve caching of e2e jobs

theochap:
## Description

This PR aggressively caches the docker/go dependencies for e2e tests in CI. This allows to take advantage of `cargo-chef`. This seems to considerably improve CI runtimes if we're not updating the dependencies in rust/go.

Locally tested, works like a charm

### DIFF
--- a/.github/workflows/e2e_tests.yaml
+++ b/.github/workflows/e2e_tests.yaml
@@ -23,26 +23,23 @@ jobs:
         uses: jlumbroso/free-disk-space@main
         with:
           large-packages: false
+
       - uses: taiki-e/install-action@just
+
+      - uses: jdx/mise-action@v2       # installs Mise + runs `mise install`
+
       - name: Setup Go 1.24.3
         uses: actions/setup-go@v5
         with:
           # Semantic version range syntax or exact version of Go
           go-version: '1.24.3'
-      - name: Cache Go modules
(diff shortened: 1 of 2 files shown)


.github/workflows/e2e_tests.yaml
```diff
@@ -48,6 +97,54 @@ jobs:
...
           go-version: '1.24.3'
+      - name: Cache Go modules
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/go/pkg/mod
+            ~/.cache/go-build
+          key: ${{ runner.os }}-go-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go-
```
clabby, line 108: Is there a reason why you're using `actions/cache` rather than the Go setup's [built-in caching](https://github.com/actions/setup-go?tab=readme-ov-file#caching-dependency-files-and-build-outputs)?

clabby (commented):
This will blow out our cache really fast. Let's persist these caches only on the main branch?

.github/workflows/e2e_tests.yaml
theochap: Good call, set up the built in caching in go!

refcell (approved):

PR #2083, opened 2025-06-10
chore(p2p): Banned Peer Count

refcell:
### Description

Teeny PR to return the banned peer count in the peer stats rpc request.


crates/node/p2p/src/rpc/request.rs
```diff
@@ -507,8 +509,7 @@ impl P2pRpcRequest {
                 blocks_topic_v2: block_topics[1],
                 blocks_topic_v3: block_topics[2],
                 blocks_topic_v4: block_topics[3],
-                // TODO(@theochap): track the number of banned peers
-                banned: 0,
+                banned: banned_peers,
```
clabby, line 512:
```suggestion
                banned: banned_peers as u32,
```

refcell: wow surprised my linter didn't catch this...

PR #2084, opened 2025-06-10
chore(p2p): Protected Peer Stat

refcell:
### Description

Micro PR to correctly return if the peer is protected in the connection gate.

Progress on #1562


clabby (approved):
nice

PR #2112, opened 2025-06-11
feat(p2p): Track and Return Peer Ping Latencies

refcell:
### Description

Tracks peer ping latencies so we can return the latencies in the peer info response.

Progress on #1562

### DIFF
--- a/crates/node/p2p/src/rpc/request.rs
+++ b/crates/node/p2p/src/rpc/request.rs
@@ -300,6 +300,8 @@ impl P2pRpcRequest {
                 return;
             };
 
+            let pings = { pings.lock().await.clone() };
+
             let node_to_peer_id: HashMap<NodeId, PeerId> = peer_ids.into_iter().filter_map(|id|
             {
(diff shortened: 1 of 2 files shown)


crates/node/p2p/src/rpc/request.rs
```diff
@@ -227,6 +227,9 @@ impl P2pRpcRequest {
             .map(|id| (id, gossip.connection_gate.connectedness(&id)))
             .collect::<HashMap<PeerId, Connectedness>>();
 
+        // Clone the ping map
+        let pings = gossip.ping.clone();
```
clabby, line 231: Not a blocker, but this map can get _huge_ after the node has been online for a while. Let's `Arc` it and ensure that it's pruned as peers disconnect?

clabby (approved):
Good to go for now. Let's keep an eye on this to see if using the `std` mutex causes any issues.

It'll block the whole thread while waiting to acquire the lock, which could be pretty problematic. Switching over to a tokio mutex and biting the `async` bullet would be preferable if not too much trouble.

PR #2118, opened 2025-06-12
feat(node/p2p): handle peer score inside `opp2p_peers`

theochap:
## Description

This PR completes the implementation of the `opp2p_peers` RPC endpoint by exposing the `peer_score` through the endpoint.

### DIFF
--- a/crates/node/p2p/src/rpc/request.rs
+++ b/crates/node/p2p/src/rpc/request.rs
@@ -7,7 +7,7 @@ use std::{
     sync::Arc,
 };
 
-use crate::{Discv5Handler, GossipDriver};
+use crate::{Discv5Handler, GossipDriver, GossipScores, TopicScores};
 use alloy_primitives::map::foldhash::fast::RandomState;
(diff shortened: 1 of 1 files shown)


crates/node/p2p/src/rpc/request.rs
```diff
@@ -389,11 +451,25 @@ impl P2pRpcRequest {
...
-                                peer_scores: PeerScores::default(),
+                                peer_scores: PeerScores {
+                                    gossip: GossipScores {
+                                        total: score,
+                                        blocks: topic_params,
+                                        // Note: We can't compute the ip colocation factor because
+                                        // `rust-libp2p` doesn't expose that information
+                                        ip_colocation_factor: Default::default(),
+                                        // Note: We can't compute the behavioral penalty because
+                                        // `rust-libp2p` doesn't expose that information
```
theochap, line 465: Can open an issue on `rust-libp2p` if we feel that'd be helpful to expose those two fields

PR #2122, opened 2025-06-13
feat(supervisor/rpc); Implement `supervisor_syncStatus` DB read

emhane:
Closes https://github.com/op-rs/kona/issues/1727

### DIFF
--- a/crates/supervisor/core/src/supervisor.rs
+++ b/crates/supervisor/core/src/supervisor.rs
@@ -6,7 +6,9 @@ use async_trait::async_trait;
 use jsonrpsee::types::{ErrorCode, ErrorObjectOwned};
 use kona_interop::{ExecutingDescriptor, SafetyLevel};
 use kona_protocol::BlockInfo;
-use kona_supervisor_storage::{ChainDb, ChainDbFactory, DerivationStorageReader, StorageError};
+use kona_supervisor_storage::{
+    ChainDb, ChainDbFactory, DerivationStorageReader, HeadRefStorageReader, StorageError,
+};
 use kona_supervisor_types::SuperHead;
 use op_alloy_rpc_types::InvalidInboxEntry;
 use std::{collections::HashMap, sync::Arc};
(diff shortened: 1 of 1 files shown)


crates/supervisor/core/src/supervisor.rs
```diff
@@ -210,8 +212,17 @@ impl SupervisorService for Supervisor {
         self.config.dependency_set.dependencies.keys().copied()
     }
 
-    fn super_head(&self, _chain: ChainId) -> Result<SuperHead, SupervisorError> {
-        todo!("implement call to ChainDbFactory")
+    fn super_head(&self, chain: ChainId) -> Result<SuperHead, SupervisorError> {
+        let db = self.database_factory.get_db(chain)?;
+
+        let l1_source = db.get_current_l1()?;
```
emhane, line 218: why can't we just use `DerivationStorageReader::derived_to_source` passing the block id of the local unsafe block as param here @varun-doshi ?

dhyaniarun1993: I think this comment is for me.

crates/supervisor/core/src/supervisor.rs
```diff
@@ -210,8 +212,17 @@ impl SupervisorService for Supervisor {
...
+        let local_unsafe = db.get_safety_head_ref(SafetyLevel::Unsafe)?;
+        let cross_unsafe = db.get_safety_head_ref(SafetyLevel::CrossUnsafe)?;
```
dhyaniarun1993, line 220: It would be better to create a `get_super_head` method on the `chainDB` and use that here. `HeadRefStorageReader::get_super_head` can use a single `tx` to get the data from the store thereby making sure we have the consistent view of the data.

emhane: I thought so too, however we can do this as a follow up I thought in MVP 3, unless it is a non-spec'd assumption of the Go implementation which hence will lead to e2e tests using this rpc method failing

dhyaniarun1993: Yep, I a happy with this being the first iteration.

crates/supervisor/core/src/supervisor.rs
emhane: I see, that is what I suspected, thanks for clarifying. will add this to docs for next person who wonders.

dhyaniarun1993 (approved):

crates/supervisor/core/src/supervisor.rs
emhane: https://github.com/op-rs/kona/pull/2127

PR #2127, opened 2025-06-13
chore(supervisor/storage): Add note about local unsafe block to `DerivationStorageReader::derived_to_source`

emhane:
Add note to docs for `DerivationStorageReader::derived_to_source` about L1 source block of local unsafe block, possibly not yet being written to db.

### DIFF
--- a/crates/supervisor/storage/src/traits.rs
+++ b/crates/supervisor/storage/src/traits.rs
@@ -15,9 +15,8 @@ use std::fmt::Debug;
 pub trait DerivationStorageReader: Debug {
     /// Gets the source [`BlockInfo`] for a given derived block [`BlockNumHash`].
     ///
-    /// NOTE: [`LocalUnsafe`] block might or might not be part of the derivation storage. There are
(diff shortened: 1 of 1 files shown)


crates/supervisor/storage/src/traits.rs
```diff
@@ -15,12 +15,18 @@ use std::fmt::Debug;
...
+    /// [`HeadRefStorageReader::get_current_l1`].
+    ///
     /// # Arguments
     /// * `derived_block_id` - The identifier (number and hash) of the derived (L2) block.
     ///
     /// # Returns
     /// * `Ok(BlockInfo)` containing the source block information if it exists.
     /// * `Err(StorageError)` if there is an issue retrieving the source block.
+    ///
+    /// [`LocalUnsafe`]: SafetyLevel::Unsafe
```
dhyaniarun1993, line 29: Not sure what this means since `LocalUnsafe` blocks aren't part of derivation storage.

emhane: updated the comment

Issue #2136, opened 2025-06-13
fix(supervisor/rpc): `localUnsafe` ref always points to genesis


dhyaniarun1993:
# Description

Update `localUnsafe` ref during UnSafe block event

PR #2145, opened 2025-06-14
fix(supervisor/rpc): localUnsafe always points to genesis

varun-doshi:
Ref #2136


emhane (approved):
nice! pending @dhyaniarun1993

dhyaniarun1993 (changes requested):
Thanks @varun-doshi for picking this up. The `Safety head ref` should be done within the same transaction as `process_and_store_logs`. Please refer `save_derived_block_pair`.

dhyaniarun1993:
@varun-doshi Can you resolve the conflicts?

PR #2148, opened 2025-06-16
feat (supervisor/rpc): implemented `checkAccessList` RPC method 

### LINKED ISSUE #1729: feat(supervisor/rpc): implement supervisor_checkaccesslist
# Description

Implement supervisor_checkaccesslist RPC method.

Ref: https://github.com/ethereum-optimism/specs/blob/main/specs/interop/supervisor.md#supervisor_checkaccesslist

sadiq1971:
Closes #1729


crates/supervisor/core/src/supervisor.rs
```diff
@@ -215,6 +219,34 @@ impl Supervisor {
...
+    ) -> Result<(), SupervisorError> {
+        // check block exists on the derivation storage
+        self.database_factory
+            .get_db(chain_id)?
+            .derived_to_source(BlockNumHash { number: block.number, hash: block.hash })?;
+
+        let head_ref = self.database_factory.get_db(chain_id)?.get_safety_head_ref(safety)?;
+
+        if head_ref.number < block.number {
+            return Err(SupervisorError::from(SuperchainDAError::ConflictingData));
```
emhane, line 245: nice that you're making use of `SuperchainDAError` already! assuming Go impl returns the same error if this condition is met

sadiq1971: Yes.

itschaindev (commented):
A few comments, cross-checked with spec, looks good on the first iteration

crates/supervisor/types/src/access_list.rs
```diff
@@ -0,0 +1,447 @@
...
+    }
+
+    /// Checks the expiry and validity of the message
+    pub const fn validate_message_lifetime(
+        &self,
+        exec_ts: u64,
+        exec_ts_with_duration: u64,
+        expiry_window: u64,
+    ) -> Result<(), AccessListError> {
+        if self.timestamp > exec_ts {
```
dhyaniarun1993, line 90:
```suggestion
        if self.timestamp >= exec_ts {
```

crates/supervisor/types/src/access_list.rs
```diff
@@ -0,0 +1,447 @@
...
+            return Err(AccessListError::MalformedEntry);
```
dhyaniarun1993, line 91: Please create a new Error `InvalidTimestampInvariant` for this.

crates/supervisor/core/src/supervisor.rs
```diff
@@ -256,13 +288,68 @@ impl SupervisorService for Supervisor {
...
+            // Ref: https://github.com/ethereum-optimism/specs/blob/main/specs/interop/derivation.md#invariants
+
+            // TODO: support 32 bytes chain id and convert to u64 via dependency set to be usable
+            // across services
+            let initiating_chain_id =
+                u64::from_be_bytes(access.chain_id[24..32].try_into().unwrap());
+
+            // TODO: Extend the `ExecutingDescriptor` to accept chain_id.
+            // And set executing_chain_id as initiating_chain_id only for backward compat.
+            let executing_chain_id = initiating_chain_id;
```
dhyaniarun1993, line 313: Can we remove this for now since it's not part of the spec? 
We can keep the todo to discuss the same with the team.

sadiq1971: It was missing from the spec. A PR is already there to fix this
https://github.com/op-rs/kona/pull/2173

crates/supervisor/core/src/config/rollup_config_set.rs
```diff
@@ -78,6 +78,16 @@ impl RollupConfig {
...
     }
+
+    /// Returns `true` if the timestamp is strictly after the interop activation block.
+    ///
+    /// Interop activates at `interop_time`. This function checks whether the current
+    /// block timestamp is *after* that activation, skipping the activation block itself.
+    ///
+    /// Returns `false` if `interop_time` is not configured.
+    pub fn is_post_interop(&self, timestamp: u64) -> bool {
+        self.interop_time.is_some_and(|t| timestamp.saturating_sub(self.block_time) >= t)
```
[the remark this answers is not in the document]
sadiq1971: Nope. if you follow the doc of this method - post interop time is the time after interop activation block.
Now interop time = t, interop activation block time is `t+block_time`. So, we need to check if the timestamp is greater than `t+block_time`.

crates/supervisor/types/src/access_list.rs
```diff
@@ -0,0 +1,447 @@
...
+///
+/// # Returns
+///
+/// A vector of fully parsed [`Access`] items if all entries are valid.
+///
+/// # Errors
+///
+/// Returns [`AccessListError`] if entries are out-of-order, malformed, or incomplete.
+pub fn parse_access_list(entries: Vec<B256>) -> Result<Vec<Access>, AccessListError> {
+    let mut list = Vec::with_capacity(entries.len() / 2);
```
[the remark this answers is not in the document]
sadiq1971: The entries are what we are getting from RPC. Now, to build `Access`, we need at least two entries. There might be at most 3 entries to build the `Access` object, but usually it will be 2. So we are setting the capacity closest to the number of entries we might get, so the vector doesn't get reallocated multiple times.

Issue #1083, opened 2025-02-18
[Feature] Add remaining predeploy addresses


emhane:
### Component

protocol

### Describe the feature you would like

Add remaining predeploy addresses as constants
https://github.com/op-rs/maili/blob/main/crates/protocol/src/predeploys.rs

### Additional context

https://specs.optimism.io/protocol/predeploys.html

thinking these constants make more sense in `maili-genesis`, though moving these can also be handled as separate issue

PR #2149, opened 2025-06-16
feat: add predeploy_address constants in maili-genesis now kona-genesis

JoshdfG:
Closes: #1083 add remainig predeploy_addresses and move into maili-genesis
You said, move the constants to maili-genesis.rs.
@emhane [See this file] (https://github.com/JoshdfG/kona/blob/main/crates/protocol/protocol/src/predeploys.rs)


crates/protocol/genesis/src/genesis.rs
refcell, line 1: Looks good, but can we move this into it's own module? `predeploys.rs` perhaps?

JoshdfG: Okay then, that's what I was trying to clarify

JoshdfG:
> link to specs in docs for each address

Done

Issue #2107, opened 2025-06-11
feat(node): Supervisor Actor


refcell:
### Description

Introduce a new supervisor actor to kona's rollup node service.

This new actor will contain all the channels needed to communicate with other actors in the node.

It should contain a trait-abstracted component that implements the methods defined in the ["Managed Node" specs](https://specs.optimism.io/interop/managed-mode.html).

By defining these methods as a trait, a supervisor implementation could programmatically plug right into the kona node directly. Alternatively, an implementation of the trait can expose an RPC server that allows the golang supervisor to "manage" the `kona-node`.

PR #2161, opened 2025-06-16
feat(node): Supervisor Actor Setup

refcell:
### Description

Sets up the supervisor actor which will spawn and handle communication to and from the supervisor as an external or in-process service.

Progress on #2107.

### DIFF
--- a/crates/node/service/src/actors/supervisor.rs
+++ b/crates/node/service/src/actors/supervisor.rs
@@ -3,6 +3,7 @@
 use crate::NodeActor;
 use async_trait::async_trait;
 use futures::{Stream, StreamExt};
+use kona_interop::{ControlEvent, ManagedEvent};
 use thiserror::Error;
 use tokio::sync::mpsc;
 use tokio_util::sync::CancellationToken;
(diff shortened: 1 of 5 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #2161 👈
* `main`

crates/node/service/src/actors/supervisor.rs
```diff
@@ -0,0 +1,106 @@
...
+                // Check if the cancellation token has been cancelled.
+                _ = self.cancellation.cancelled() => {
+                    warn!(target: "supervisor", "Supervisor actor cancelled");
+                    return Ok(());
+                },
+                // Wait for an event from the kona node.
+                Some(event) = self.node_events.recv() => {
+                    // Process the event and send it to the supervisor.
+                    if let Err(err) = self.supervisor_ext.send_event(event).await {
+                        error!(target: "supervisor", ?err, "Failed to send event to supervisor");
```
theochap, line 78: In what case can we get an error inside `send_event`? Maybe we should return after - if not, can we recover from that error? This case seems similar to the branch below (where we return an error)

refcell: This should only error if it's fatal.

PR #2163, opened 2025-06-16
chore(node/rpc): cleanup rpc config flags and launcher

theochap:
## Description

This PR does some logic cleanups for RPC config. In particular:

- Simplifies the data structures (merge `RpcConfig` and `RpcLauncher`)
- Code cleanups

### DIFF
--- a/crates/node/rpc/src/launcher.rs
+++ b/crates/node/rpc/src/launcher.rs
@@ -47,6 +47,22 @@ impl RpcLauncher {
         Self { config, module: RpcModule::new(()) }
     }
 
+    /// Creates a new [`RpcLauncher`] that is disabled.
+    pub fn new_disabled() -> Self {
+        Self {
+            config: RpcConfig {
+                disabled: true,
+                no_restart: false,
+                // Use a dummy socket address. The RPC server is disabled, so it will not be used.
+                socket: SocketAddr::from(([127, 0, 0, 1], 8080)),
+                enable_admin: false,
+                admin_persistence: None,
+                ws_enabled: false,
+            },
+            module: RpcModule::new(()),
+        }
+    }
+
     /// Returns whether WebSocket RPC endpoint is enabled
     pub const fn ws_enabled(&self) -> bool {
         self.config.ws_enabled
--- a/crates/node/service/src/service/standard/builder.rs
+++ b/crates/node/service/src/service/standard/builder.rs
@@ -18,7 +18,7 @@ use url::Url;
 use kona_genesis::RollupConfig;
 use kona_p2p::Config;
 use kona_providers_alloy::OnlineBeaconClient;
-use kona_rpc::{RpcConfig, SupervisorRpcConfig};
+use kona_rpc::{RpcConfig, RpcLauncher, SupervisorRpcConfig};
 
 /// The [`RollupNodeBuilder`] is used to construct a [`RollupNode`] service.
 #[derive(Debug, Default)]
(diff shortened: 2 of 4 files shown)


bin/node/src/commands/net.rs
```diff
@@ -54,10 +54,21 @@ impl NetCommand {
...
+
+            info!(target: "net", "Starting RPC server on {:?}", launcher.socket());
+
+            let handle = launcher.launch().await?;
+
+            Some(handle)
+        } else {
+            info!(target: "net", "RPC server disabled");
+            None
+        };
```
refcell, line 71: This feels more complex to me than the previous interface. Is the goal of this to hoist the rpc enabled/disabled branch logic?

crates/node/rpc/src/config.rs
```diff
@@ -1,22 +1,17 @@
...
-    /// If the RPC is enabled.
-    pub enabled: bool,
     /// Prevent the rpc server from being restarted.
     pub no_restart: bool,
-    /// The RPC listening address.
-    pub listen_addr: IpAddr,
-    /// The RPC listening port.
-    pub listen_port: u16,
+    /// The RPC socket address.
+    pub socket: SocketAddr,
```
refcell, line 14: I like this change, no need to wait to construct the socket address.

crates/node/rpc/src/launcher.rs
```diff
@@ -35,115 +33,101 @@ pub struct HealthzResponse {
...
-        Self {
-            disabled: false,
-            no_restart: false,
-            socket: Some(socket),
-            module: None,
-            ws_enabled: false,
-        }
-    }
+    /// The RPC configuration associated with the [`RpcLauncher`].
+    pub(crate) config: RpcConfig,
```
refcell, line 39: Makes sense to just inline the config. I'm a proponent of this change.

crates/node/service/src/service/standard/builder.rs
```diff
@@ -132,7 +132,7 @@ impl RollupNodeBuilder {
         let rpc_client = RpcClient::new(http_hyper, false);
         let l2_provider = RootProvider::<Optimism>::new(rpc_client);
 
-        let rpc_launcher = self.rpc_config.map(|c| c.as_launcher()).unwrap_or_default();
+        let rpc_launcher = self.rpc_config.map(|c| c.as_launcher());
```
refcell, line 135: We'll need to add the `with_healthz` call on the launcher here now since it's optional. We always want the node to expose the `/healthz` endpoint if the rpc is launched so we should call it here.

refcell (commented):
A lot of nice changes here. I have a slight preference to keeping the `enabled` bool flag in the config. This allows the user interface for the launcher to be much cleaner. The user doesn't need to branch on an optional launcher; they can just set whether the rpc is enabled and always merge modules.

crates/node/service/src/service/validator.rs
```diff
@@ -166,18 +166,20 @@ pub trait ValidatorNodeService {
...
+            launcher.merge(rollup_rpc.into_rpc())?;
+            if launcher.ws_enabled() {
+                launcher
+                    .merge(WsRPC::new(engine_query_sender).into_rpc())
+                    .map_err(Self::Error::from)?;
+            }
+            Some(RpcActor::new(launcher, cancellation.clone()))
+        } else {
+            None
+        };
```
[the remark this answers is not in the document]
theochap: Sounds good, will revert that change then!

refcell (approved):

PR #2168, opened 2025-06-17
feat(node): Supervisor Rpc Server Flags

### LINKED ISSUE #2164: feat(node): Supervisor Rpc Server Flags
### Description

When run as a separate process, the supervisor - be it the golang or rust implementation - will need to connect to the l2 consensus over a websocket connection. This is detailed in the ["Managed Mode" OP Stack Specs](https://specs.optimism.io/interop/managed-mode.html#control-flow-between-supervisor-and-managed-node).

Since the websocket server serves managed events to the supervisor and responds to requests from the supervisor, it belongs as an actor on the `kona-node`.

This websocket server is distinct from the node's rpc server since it will need to perform JWT authentication.
The `jsonrpsee` github repository provides a [good example](https://github.com/paritytech/jsonrpsee/blob/master/examples/examples/jsonrpsee_as_service.rs#L107-L113) for how to setup auth.

This ticket is to add supervisor cli flags to the `kona-node` that allow the user to specify the ip address + port for the supervisor websocket server. It should also allow the user to specify either a jwt secret or a path of the file that contains the jwt secret. All of these flags should have sensible defaults.

Also add a `--supervisor.rpc-enabled` flag that is _disabled_ by default. Meaning, the supervisor rpc is not enabled by default.

varun-doshi:
Fixes #2164

### DIFF
--- a/bin/node/src/flags/supervisor.rs
+++ b/bin/node/src/flags/supervisor.rs
@@ -1,0 +1,238 @@
+//! Supervisor RPC CLI Flags
+//!
+//! These are based on supervisor flags from the [`op-node`][op-node] CLI.
+//!
+//! [op-node]: https://github.com/ethereum-optimism/optimism/blob/develop/op-node/flags/flags.go#L233-L265
+
+use std::{fs, net::IpAddr, path::PathBuf};
+
+use anyhow::{Ok, anyhow};
+use clap::Parser;
+
+/// Supervisor CLI Flags
+#[derive(Parser, Clone, Debug, PartialEq, Eq)]
+pub struct SupervisorArgs {
+    /// Enable Supervisor Websocket
+    #[arg(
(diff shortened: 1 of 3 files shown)


bin/node/src/flags/supervisor.rs
```diff
@@ -0,0 +1,238 @@
...
+pub struct SupervisorArgs {
+    /// Enable Supervisor Websocket
+    #[arg(
+        long = "supervisor.rpc-enabled",
+        default_value = "false",
+        env = "KONA_NODE_SUPERVISOR_RPC_ENABLED"
+    )]
+    pub rpc_enabled: bool,
+
+    /// IP to bind Sequencer Websocket to.
```
refcell, line 23:
```suggestion
    /// IP to bind Supervisor Websocket RPC server to.
```

refcell (approved):

PR #2173, opened 2025-06-17
feat(supervisor/interop): added a `chain_id` field to ExecutingDescriptor

### LINKED ISSUE #2170: fix (supervisor/interop): `ExecutingDescriptor` should include chain_id field
The Rust `ExecutingDescriptor` type, defined [here](https://github.com/op-rs/kona/blob/4971c0de7ae1be0fa602f1c927fa6c022adaf203/crates/protocol/interop/src/message.rs#L79) currently lacks a `chain_id` field. However, the [Go implementation](https://github.com/ethereum-optimism/optimism/blob/88e70f8f9f805a845171e51ee879c974362e6c54/op-supervisor/supervisor/types/types.go#L307) of `ExecutingDescriptor` includes `ChainID`, and the [spec](https://github.com/ethereum-optimism/specs/blob/main/specs/interop/supervisor.md#executingdescriptor) implies its presence or necessity for accurate execution context.

Currently in Rust, we infer the `chain_id` from the executing messages for backward compatibility. However, this limits flexibility and clarity when the chain ID must be explicitly known or validated during runtime.

PoulavBhowmick03:
Fixes #2170

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4503,6 +4503,7 @@ dependencies = [
  "alloy-eips",
  "alloy-primitives",
  "alloy-rlp",
+ "alloy-serde",
  "alloy-sol-types",
  "arbitrary",
  "async-trait",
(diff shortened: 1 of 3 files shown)


emhane:
please also add a test to check the serialisation is as expected, ref
https://github.com/alloy-rs/alloy/blob/c714dd3018608a6933d7c26adb0d46ceb3cda509/crates/serde/src/quantity.rs#L351-L365

PR #2175, opened 2025-06-17
chore(node/net): simplify the network interface.

theochap:
## Description

This PR simplifies and cleans-up the network interface. In particular it:

- Removes error variants for the network builders by adding type-safety (making config fields mandatory instead of optional)
- Removes duplicate fields

### DIFF
--- a/crates/node/p2p/src/net/builder.rs
+++ b/crates/node/p2p/src/net/builder.rs
@@ -195,9 +195,21 @@ mod tests {
     use libp2p::gossipsub::IdentTopic;
     use std::net::{IpAddr, Ipv4Addr, SocketAddr};
 
-    fn network_builder(signer: Option<Address>) -> NetworkBuilder {
+    #[derive(Debug)]
+    struct NetworkBuilderParams {
+        rollup_config: RollupConfig,
+        signer: Address,
+    }
+
+    impl Default for NetworkBuilderParams {
+        fn default() -> Self {
+            Self { rollup_config: RollupConfig::default(), signer: Address::random() }
+        }
+    }
+
+    fn network_builder(params: NetworkBuilderParams) -> NetworkBuilder {
         let keypair = Keypair::generate_secp256k1();
-        let signer = signer.unwrap_or(Address::random());
+        let signer = params.signer;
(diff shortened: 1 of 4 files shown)


crates/node/p2p/src/gossip/builder.rs
```diff
@@ -8,54 +8,56 @@ use libp2p::{
...
 };
 
 /// A builder for the [`GossipDriver`].
-#[derive(Debug, Default)]
+#[derive(Debug)]
 pub struct GossipDriverBuilder {
-    /// The idle connection timeout as a [`Duration`].
-    timeout: Option<Duration>,
+    /// The [`RollupConfig`] for the network.
+    rollup_config: RollupConfig,
```
refcell, line 21: Having an owned `RollupConfig` feels like tech debt.
It looks like the `RollupConfig` is passed into the `BlockHandler` constructor so it can switch block topics based on hardfork timestamps.
Let me think more about how we can best decouple this logic.
Perhaps we'll want to move the `BlockHandler` into a new `kona-topics` crate that holds the block topic + handler logic so the large `RollupConfig` doesn't need to be passed around like this in memory.

crates/node/p2p/src/net/builder.rs
```diff
@@ -36,36 +32,51 @@ pub struct NetworkBuilder {
...
+        .with_discovery_randomize(config.discovery_randomize)
+        .with_bootstore(config.bootstore)
+        .with_bootnodes(config.bootnodes)
+        .with_discovery_interval(config.discovery_interval)
+        .with_gossip_config(config.gossip_config)
+        .with_peer_scoring(config.scoring)
+        .with_peer_monitoring(config.monitor_peers)
+        .with_topic_scoring(config.topic_scoring)
+        .with_gater_config(config.gater_config)
+        .with_local_signer(config.local_signer)
```
refcell, line 52: It's an aesthetic thing but I find this to be a bit of an unappealing user interface personally. Not a big deal though.

refcell: Totally correct! I'm fine merging this and going this route

PR #2180, opened 2025-06-18
feat(supervisor/rpc): Implement `supervisor_superRootAtTimestamp`

### LINKED ISSUE #1726: feat(supervisor/rpc): implement supervisor_superrootattimestamp
# Description

Implement supervisor_superrootattimestamp RPC method.

Ref: https://github.com/ethereum-optimism/specs/blob/main/specs/interop/supervisor.md#supervisor_superrootattimestamp

itschaindev:
Closes #1726

### DIFF
--- a/crates/protocol/interop/src/root.rs
+++ b/crates/protocol/interop/src/root.rs
@@ -4,6 +4,7 @@
 
 use crate::{SUPER_ROOT_VERSION, SuperRootError, SuperRootResult};
 use alloc::vec::Vec;
+use alloy_eips::BlockNumHash;
 use alloy_primitives::{B256, Bytes, U256, keccak256};
 use alloy_rlp::{Buf, BufMut};
 
@@ -109,10 +110,14 @@ pub struct ChainRootInfo {
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 pub struct SuperRootResponse {
+    /// The Highest L1 Block that is cross-safe among all chains.
+    pub cross_safe_derived_from: BlockNumHash,
     /// The timestamp of the super root.
     pub timestamp: u64,
     /// The super root hash.
     pub super_root: B256,
+    /// The version of the super root.
+    pub version: u8,
     /// The chain root info for each chain in the dependency set.
     /// It represents the state of the chain at or before the timestamp.
     pub chains: Vec<ChainRootInfo>,
--- a/crates/protocol/interop/src/root.rs
+++ b/crates/protocol/interop/src/root.rs
@@ -109,7 +109,7 @@ pub struct ChainRootInfo {
 #[derive(Debug, Clone, Eq, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
-pub struct SuperRootResponse {
+pub struct SuperRootOutput {
     /// The Highest L1 Block that is cross-safe among all chains.
     pub cross_safe_derived_from: BlockNumHash,
     /// The timestamp of the super root.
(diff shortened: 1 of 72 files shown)


crates/protocol/interop/src/root.rs
```diff
@@ -109,10 +110,14 @@ pub struct ChainRootInfo {
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
 pub struct SuperRootResponse {
+    /// The Highest L1 Block that is cross-safe among all chains.
```
emhane, line 113: since this type is only used in supervisor crate, it makes more sense to put it in _crates/supervisor/rpc/src/response.rs_ no?

itschaindev: I don't know if there were any plans on using this type in interop crate in the future. They might still need it. Better to leave it there since we are anyways using several types from that crate in supervisor.

sadiq1971:
> @dhyaniarun1993 @sadiq1971 I'm introducing a new trait for calling API on managed node. Do you think we can club this and `ReceiptProvider`? We can also add the remaining API call methods to this trait.
> 
> Also, if you want I can separate the types in different files to make it consistent.
> 
> Let me know what you guys think.

No issues introducing the new trait. On club - I think it is still better to have trait based on uses - for example ReceiptProvider being used in log_indexer and if we club we are going to send a bunch of other method that is not required there. But also a lot of traits might looks messy. We can think about it later as well

crates/supervisor/core/src/supervisor.rs
```diff
@@ -61,6 +67,15 @@ pub trait SupervisorService: Debug + Send + Sync {
...
     fn finalized_l1(&self) -> Result<BlockInfo, SupervisorError>;
 
+    /// Returns the [`SuperRootResponse`] at a specified timestamp, which represents the global
+    /// state across all monitored chains.
+    ///
+    /// [`SuperRootResponse`]: kona_interop::SuperRootResponse
+    async fn super_root_at_timestamp(
+        &self,
+        timestamp: u64,
+    ) -> Result<SuperRootResponse, SupervisorError>;
```
emhane, line 77: it seems counterintuitive that the rpc response is assembled in the service. it would be more adequate if this returned only a tuple of super roots and chain infos. that way, the service handles processing which requires communication with db or managed node, but nothing more. the rpc server then handles only communication with the service, and any further processing of the data it returns into the rpc response format.

crates/supervisor/core/src/supervisor.rs
```diff
@@ -222,6 +237,59 @@ impl SupervisorService for Supervisor {
...
+    async fn super_root_at_timestamp(
+        &self,
+        timestamp: u64,
+    ) -> Result<SuperRootResponse, SupervisorError> {
+        let mut chain_ids = self.config.dependency_set.dependencies.keys().collect::<Vec<_>>();
+        // Sorting chain ids for deterministic super root hash
+        chain_ids.sort();
+
+        let mut chain_infos = Vec::<ChainRootInfo>::with_capacity(chain_ids.len());
+        let mut super_root_chains = Vec::<OutputRootWithChain>::with_capacity(chain_ids.len());
```
emhane, line 249: nice

emhane (changes requested):

dhyaniarun1993 (approved):

crates/supervisor/types/src/types.rs
```diff
@@ -30,42 +29,9 @@ impl BlockSeal {
...
-    ) -> Self {
-        Self { hash, number, parent_hash, timestamp, l1_origin, sequence_number }
-    }
-}
-
 /// Output data for version 0 of the protocol.
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(rename_all = "camelCase")]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Default)]
+#[serde(rename_all = "PascalCase")]
```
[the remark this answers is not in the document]
dhyaniarun1993: As of now it's `PascalCase`.
There is discord thread going on for this.

PR #2187, opened 2025-06-18
fix(interop): remove timestamps from the depset

dhyaniarun1993:
This PR 
- removes the timestamps from the depset since we are now using rollup config for it.
- this is also inline with the the response of the rpc method.

### DIFF
--- a/crates/protocol/interop/src/depset.rs
+++ b/crates/protocol/interop/src/depset.rs
@@ -6,20 +6,7 @@ use kona_registry::HashMap;
 #[derive(Debug, Clone, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 #[cfg_attr(feature = "serde", serde(rename_all = "camelCase"))]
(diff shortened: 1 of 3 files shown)


crates/protocol/interop/src/depset.rs
```diff
@@ -6,20 +6,7 @@ use kona_registry::HashMap;
...
-    /// This is the minimum timestamp of the inclusion of an executing message that can be
-    /// considered valid.
-    pub activation_time: u64,
-
-    /// Lower bound timestamp (in seconds) for data storage and accessibility.
-    /// This is the minimum timestamp of an initiating message that is required to be
-    /// accessible to other chains. This is set to 0 when all data since genesis is executable.
-    pub history_min_time: u64,
-}
+pub struct ChainDependency {}
```
emhane, line 9:
```suggestion
pub struct ChainDependency;
```
preferred syntax for ZST

emhane (changes requested):

emhane (approved):

itschaindev (approved):

PR #2198, opened 2025-06-18
chore(node/rpc): unify BaseFeeConfig to follow op-geth

theochap:
## Description

This PR unifies the `BaseFeeConfig` format between `op-geth` and the `kona-node`. 

`op-geth` format: https://github.com/ethereum-optimism/op-geth/blob/6005dd53e1b50fe5a3f59764e3e2056a639eff2f/params/config.go#L509-L514 as they're using `u64` instead.

We were following `op-alloy`'s format until now: https://github.com/alloy-rs/alloy/blob/0031ec3dd2ac6da303056374b86c1b4a8995bc34/crates/eips/src/eip1559/basefee.rs#L15-L26


refcell (approved):
This LGTM

PR #2212, opened 2025-06-19
feat(supervisor): implemented safety checker job

### LINKED ISSUE #2185: feat(supervisor): Implement `CrossChainSafetyProviderJob`
Create a job for a particular chain that will periodically try to promote the next candidate block for cross-unsafe and cross-safe.
* As long as there is a valid candidate block, it keeps promoting.
* When it finishes promoting blocks, wait a duration less than the block generation time of that chain.

sadiq1971:
Closes #2185 
Depends On #2200


crates/supervisor/core/src/safety_checker/task.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        let candidate = self.provider.get_block(self.chain_id, current_head.number + 1)?;
+
+        Ok(candidate)
+    }
+
+    // Returns the safety level that defines the upper promotion boundary for the current level.
+    //
+    // For example:
+    // - CrossUnsafe promotions are bounded by LocalUnsafe.
```
emhane, line 126:
```suggestion
    /// - [`CrossUnsafe`](SafetyLevel::CrossUnsafe) promotions are bounded by LocalUnsafe.
```
don't think there is any problem
to turn all the comments on the private methods into docs instead. see comment about risky of variable name changing uncaught, same applies to type names.

crates/supervisor/core/src/safety_checker/task.rs
```diff
@@ -0,0 +1,255 @@
...
+    // - CrossSafe promotions are bounded by LocalSafe.
+    const fn promotion_upper_bound(&self) -> Result<SafetyLevel, CrossSafetyError> {
+        match self.target_level {
+            SafetyLevel::CrossUnsafe => Ok(SafetyLevel::LocalUnsafe),
+            SafetyLevel::CrossSafe => Ok(SafetyLevel::LocalSafe),
+            _ => Err(CrossSafetyError::UnsupportedTargetLevel(self.target_level)),
+        }
+    }
+}
```
emhane, line 135: the target level doesn't change as the program runs, so this can be done once as part of the constructor, and stored in a new files of the type `promotion_upper_bound`

crates/supervisor/core/src/safety_checker/task.rs
```diff
@@ -0,0 +1,255 @@
...
+
+        Ok(candidate)
+    }
+
+    // Returns the safety level that defines the upper promotion boundary for the current level.
+    //
+    // For example:
+    // - CrossUnsafe promotions are bounded by LocalUnsafe.
+    // - CrossSafe promotions are bounded by LocalSafe.
+    const fn promotion_upper_bound(&self) -> Result<SafetyLevel, CrossSafetyError> {
```
emhane, line 128: wondering if a more helpful name for this function is `target_level_lower_bound`?

in terms of safety level, local-safe is lower than cross-safe. the name upper bounds applies to block height, but block height isn't what is handled in the scope of this function

crates/supervisor/core/src/safety_checker/task.rs
```diff
@@ -0,0 +1,255 @@
...
+        self.provider.update_safety_head_ref(self.chain_id, self.target_level, &candidate)?;
+
+        Ok(candidate)
+    }
+
+    // Finds the next block that is eligible for promotion at the current target level.
+    fn find_next_promotable_block(&self) -> Result<BlockInfo, CrossSafetyError> {
+        let promotion_boundary = self.promotion_upper_bound()?;
+
+        let current_head = self.provider.get_safety_head_ref(self.chain_id, self.target_level)?;
```
emhane, line 111: will any other threads be promoting blocks for the same chain in parallel? otherwise we can optimise this by storing the current block number in a field of this type, and updating it when a block is promoted. that way we only need one db/cache read (upper head) each time this thread is polled. would this optimisation be reorg safe?

crates/supervisor/core/src/safety_checker/task.rs
```diff
@@ -0,0 +1,255 @@
...
+
+    // Returns the safety level that defines the upper promotion boundary for the current level.
+    //
+    // For example:
+    // - CrossUnsafe promotions are bounded by LocalUnsafe.
+    // - CrossSafe promotions are bounded by LocalSafe.
+    const fn promotion_upper_bound(&self) -> Result<SafetyLevel, CrossSafetyError> {
+        match self.target_level {
+            SafetyLevel::CrossUnsafe => Ok(SafetyLevel::LocalUnsafe),
+            SafetyLevel::CrossSafe => Ok(SafetyLevel::LocalSafe),
```
emhane, line 131:
```suggestion
            SafetyLevel::CrossSafe | SafetyLevel::Finalized => Ok(SafetyLevel::LocalSafe),
```
how come finalized isn't supported?

emhane (changes requested):
second pass

crates/supervisor/core/src/safety_checker/task.rs
sadiq1971: I did it this way only considering the re-org. I think we surely will revisit this again while implementing re-org and will try to optimize. That's why I tried to avoid maintaining the current state for the block number.

crates/supervisor/core/src/safety_checker/task.rs
sadiq1971: Make sense

crates/supervisor/core/src/safety_checker/task.rs
```diff
@@ -0,0 +1,255 @@
...
+    use super::*;
+    use alloy_primitives::{B256, ChainId};
+    use kona_supervisor_storage::{CrossChainSafetyProvider, StorageError};
+    use kona_supervisor_types::Log;
+    use mockall::mock;
+    use op_alloy_consensus::interop::SafetyLevel;
+
+    mock! {
+        #[derive(Debug)]
+        pub Provider {}
```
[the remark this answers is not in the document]
sadiq1971: The mock macro doesn't accept that.

PR #2220, opened 2025-06-20
feat(supervisor/core): check for inconsistency and trigger reset

### LINKED ISSUE #2156: feat(supervisor/core): check consistency of the node against supervisor state
# Description

Processing events in chain processor is async. Hence we need to periodically check the consistency of the op-node state with the supervisor. I think we can start doing this consistency check at the `l1_exhaust_event`.

itschaindev:
Closes #2156


crates/supervisor/core/src/syncnode/error.rs
```diff
@@ -99,4 +99,15 @@ pub enum ManagedEventTaskError {
     /// Next block is either empty or unavailable.
     #[error("next block is either empty or unavailable, number: {0}")]
     NextBlockNotFound(u64),
+    /// Block not found in database.
+    #[error("block not found in database, number: {0}")]
+    BlockNotFound(u64),
+    /// Incoming derived block does not match stored block.
+    #[error("incoming derived block does not match stored block")]
```
emhane, line 106:
```suggestion
    #[error("incoming derived block number does not match stored block, incoming: {incoming}, stored: {stored}")]
```

crates/supervisor/core/src/syncnode/task.rs
```diff
@@ -276,6 +271,43 @@ where
...
+
+        // check if the derived block is already stored and is consistent with the incoming derived
+        // block
+        let derived_block = derived_ref_pair.derived;
+
+        match self.db_provider.get_block(derived_block.number) {
+            Ok(stored_block) => {
+                if stored_block != derived_block {
+                    error!(target: "managed_event_task", "Incoming derived block does not match stored block");
+                    self.handle_reset("incoming derived block does not match stored block").await;
```
emhane, line 296: if there is not yet an issue open for handling refactor of `handle_reset` to take an enum instead of a string literal, pls open one

emhane (changes requested):

crates/supervisor/core/src/syncnode/task.rs
```diff
@@ -276,6 +271,43 @@ where
...
+                current: derived_ref_pair.source.hash,
+                parent: next_l1_block.header.parent_hash,
+            })?
+        }
+
+        // check if the derived block is already stored and is consistent with the incoming derived
+        // block
+        let derived_block = derived_ref_pair.derived;
+
+        match self.db_provider.get_block(derived_block.number) {
```
dhyaniarun1993, line 292: This looks incorrect. Here, we should fetch the latest derived block from the storage.

crates/supervisor/core/src/syncnode/task.rs
itschaindev: Done. #2237

PR #2231, opened 2025-06-20
chore(node/service): Squash service traits

clabby:
## Overview

Squashes the `ValidatorNodeService`, `SequencerNodeService`, and `RollupNodeService` traits into a single trait: `RollupNodeService.

The additional traits are not very helpful for separate configurations, since the sequencer actor can be optionally configured at runtime.


refcell (approved):
Happy with this

PR #2232, opened 2025-06-20
feat(protocol/derive): Managed Traversal Stage

### LINKED ISSUE #2227: feat(protocol/derive): Managed L1 Traversal Stage
### Description

> [!NOTE]
>
> Builds on #2226 

#### Background

When the l2 consensus node is run in "managed mode", the node does not autonomously progress the chain. Instead, it waits for the supervisor to tell it what to do via the managed rpc server.

One critical part of the supervisor's job is to instruct the node when the l1 chain advances. In order to rejigger the `op-node` to listen for l1 updates instead of autonomously traversing the l1 chain, a new [`L1TraversalManaged` stage is used by the pipeline](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/rollup/derive/pipeline.go#L105-L109).

The pipeline has a [new `ProvideL1Traversal` event](https://github.com/ethereum-optimism/optimism/blob/e777507ce980aae7bce9b7f2ff99510212dd7995/op-node/rollup/derive/deriver.go#L197C7-L199) which provides the next l1 block to the managed traversal stage.

### Output

Since kona's pipeline is architected using stages like the `op-node`, a similar approach of creating a new `L1Traversal` stage works nicely. The new stage extends the ability for the pipeline's L1 view to be managed through events.

This ticket is to create a new `ManagedTraversal` stage that mirrors the `op-node`'s [`L1TraversalManaged`](https://github.com/ethereum-optimism/optimism/blob/e777507ce980aae7bce9b7f2ff99510212dd7995/op-node/rollup/derive/l1_traversal_managed.go) stage. It should properly handle receiving the new `Signal::ProvideBlock` variant introduced in #2226.

A future ticket will handle wiring this stage up with the derivation pipeline.

refcell:
### Description

Implements the `ManagedTraversal` stage of the derivation pipeline.

Closes #2227


crates/protocol/derive/src/stages/traversal/managed.rs
```diff
@@ -0,0 +1,342 @@
...
+}
+
+#[async_trait]
+impl<F: ChainProvider + Send> SignalReceiver for ManagedTraversal<F> {
+    async fn signal(&mut self, signal: Signal) -> PipelineResult<()> {
+        match signal {
+            Signal::Reset(ResetSignal { l1_origin, system_config, .. }) |
+            Signal::Activation(ActivationSignal { l1_origin, system_config, .. }) => {
+                self.update_origin(l1_origin);
+                self.system_config = system_config.expect("System config must be provided.");
```
theochap, line 144: If the `system_config` is mandatory (which it does seem to be the case here given that we panic), why are we passing it as an option for the signal input? Wouldn't it be safer to force the system config to be provided as part of the signal so that we don't need to panic here?

crates/protocol/derive/src/stages/traversal/polling.rs
theochap, line 1: Not related to this PR, but line 90 we don't use safe arithmetic. Can that be an issue?

theochap: Here
```
        let next_l1_origin =
            self.data_source.block_info_by_number(block.number + 1).await.map_err(Into::into)?;
```

crates/protocol/derive/src/stages/traversal/provider.rs
```diff
@@ -0,0 +1,187 @@
...
+#[async_trait]
+impl<P: ChainProvider + Send + Sync> SignalReceiver for TraversalProvider<P> {
+    /// Pass a signal to the active traversal stage.
+    async fn signal(&mut self, signal: Signal) -> PipelineResult<()> {
+        self.update_mode_from_origin();
+        if self.is_managed {
+            self.managed_traversal.signal(signal).await
+        } else {
+            self.l1_traversal.signal(signal).await
+        }
```
theochap, line 105: All those `if ... else` statements would be replaced by pattern matching if we were to use the enum approach, which would be cleaner (and more idiomatic) in my opinion (but that's opinionated so feel free to push back)

crates/protocol/derive/src/stages/traversal/provider.rs
```diff
@@ -0,0 +1,187 @@
...
+///
+/// This allows the pipeline to seamlessly switch between normal and managed operation.
+#[derive(Debug)]
+pub struct TraversalProvider<P: ChainProvider + Send + Sync> {
+    /// The autonomous L1 traversal stage (used before interop/managed mode).
+    pub l1_traversal: L1Traversal<P>,
+    /// The managed L1 traversal stage (used after interop/managed mode).
+    pub managed_traversal: ManagedTraversal<P>,
+    /// If true, use managed traversal; otherwise, use autonomous traversal.
+    pub is_managed: bool,
```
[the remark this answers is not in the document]
refcell: What would that look like concretely? Having an inner enum type with managed and poling as the variants?

I think I'll change this to more closely follow the channel stage mux where it actually instantiates the stages inside the mux. Then at least all of our muxes are consistent. We can update them later to use enums if we like.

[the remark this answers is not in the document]
theochap: Sounds good, and yes that's what I had in mind. It's a matter of personal taste, but I think in rust at least I usually prefer to use enums instead of boolean fields inside structs when possible. This usually leads to cleaner/safer code in general

crates/protocol/derive/src/stages/traversal/managed.rs
refcell: This is out of the scope of this PR - may be able to update, but this is the current handling in other stages as well.

theochap (approved):

refcell:
Replaced by #2287

PR #2235, opened 2025-06-20
feat(node/service): Supervisor Engine Resets

### LINKED ISSUE #2229: feat(node/service): Supervisor Engine Resets
### Description

When interop is enabled, the supervisor handles engine resets by sending a reset signal. It picks up that the derivation pipeline was reset through a `ManagedEvent`. See [the specs](https://specs.optimism.io/interop/managed-mode.html) for more details.

The `kona-node` currently always sends a reset signal to the engine actor when the derivation actor receives a reset step result from the derivation pipeline. This ticket is to change the derivation actor to only [send a reset signal to the engine actor](https://github.com/op-rs/kona/blob/main/crates/node/service/src/actors/derivation.rs#L182-L187) when interop is **not** active.

varun-doshi:
Fixes #2229

### DIFF
--- a/crates/node/service/src/actors/derivation.rs
+++ b/crates/node/service/src/actors/derivation.rs
@@ -178,7 +178,8 @@ where
 
                                     kona_macros::inc!(counter, Metrics::L1_REORG_COUNT);
                                 }
-                                // send the `reset` signal to the engine actor only when interop is not active.
+                                // send the `reset` signal to the engine actor only when interop is
+                                // not active.
                                 if !self
                                     .pipeline
                                     .rollup_config()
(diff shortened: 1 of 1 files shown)


crates/node/service/src/actors/derivation.rs
```diff
@@ -179,10 +179,16 @@ where
...
+                                if self
+                                    .pipeline
+                                    .rollup_config()
+                                    .is_interop_active(l2_safe_head.block_info.timestamp)
+                                {
+                                    self.reset_request_tx.send(()).await.map_err(|e| {
+                                        error!(target: "derivation", ?e, "Failed to send reset request");
+                                        DerivationError::Sender(Box::new(e))
+                                    })?;
+                                }
```
refcell, line 191: This should be the other way around - if interop is **not** active we want to send the reset request to the engine.

varun-doshi: Ahh my bad, fixed

refcell:
Nice job @varun-doshi

PR #2238, opened 2025-06-23
feat: chaindb metrics

dhyaniarun1993:


emhane (commented):
nice!

crates/supervisor/storage/src/chaindb.rs
```diff
@@ -62,171 +90,209 @@ impl ChainDb {
...
-                // this could happen during initial sync
-                warn!(
-                    target: "supervisor_storage",
-                    l1_finilized_block_number = l1_finalized.number,
-                    safe_source_block_number = safe_block_pair.source.number,
-                    "L1 finalized block is greater than safe block",
-                );
-                return sp.update_safety_head_ref(SafetyLevel::Finalized, &safe);
-            }
+        self.observe_call("update_finalized_head_ref", || {
```
sadiq1971, line 93: I was thinking, can we apply a macro without updating the code itself?
I was thinking like this way—
```
#[observe_metric("update_finalized_head_ref")]
 pub(super) fn update_finalized_head_ref.....
```

emhane: that's a great idea, suggest we open a debt issue to take care of that

dhyaniarun1993: Yep, that's a great idea to implement proc macro for this. Will create issue.

itschaindev (approved):
lgtm

PR #2240, opened 2025-06-23
feat(node/service): L1 origin selector

clabby:
## Overview

Adds an initial implementation of an L1 origin selector, for the sequencer to use when preparing payload attributes for extending the unsafe chain. [Spec](https://specs.optimism.io/protocol/derivation.html#overview).

This implementation is still subject to change, as it has not yet been plumbed into the sequencer itself, nor does it receive update events.


crates/node/service/src/actors/sequencer/origin_selector.rs
```diff
@@ -0,0 +1,148 @@
...
+
+        // If the next L1 origin is not set, fetch the next block after the current L1 origin.
+        if self.next.is_none() {
+            let next_block = self
+                .l1
+                .get_block_by_number(next_block_number.into())
+                .await?
+                .ok_or(L1OriginSelectorError::BlockNotFound(next_block_number.into()))?
+                .into_consensus()
+                .into();
```
refcell, line 85: This is duplicated in `select_origins` which makes it a good candidate for refactoring into a method. Not a blocker but slimming down these methods would be nice.

crates/node/service/src/actors/sequencer/origin_selector.rs
```diff
@@ -0,0 +1,148 @@
...
+    /// block's timestamp in relation to the current L1 origin's timestamp. If the next L2
+    /// block's timestamp is greater than the L2 unsafe head's L1 origin timestamp, the L1
+    /// origin is the block following the current L1 origin.
+    pub async fn next_l1_origin(
+        &mut self,
+        unsafe_head: L2BlockInfo,
+    ) -> Result<BlockInfo, L1OriginSelectorError> {
+        self.select_origins(&unsafe_head).await?;
+
+        let (current, mut next) = (self.current, self.next);
```
refcell, line 53: what in the gippity

refcell (approved):

crates/node/service/src/actors/sequencer/origin_selector.rs
clabby: `Copy` - don't want to mutate directly

PR #2241, opened 2025-06-23
feat(node/service): Sequencer actor

clabby:
## Overview

Adds an initial version of the sequencer actor.


crates/node/service/src/actors/sequencer/actor.rs
```diff
@@ -0,0 +1,232 @@
...
+        if self.cfg.is_first_isthmus_block(attributes.payload_attributes.timestamp) {
+            info!(target: "sequencer", "Sequencing isthmus upgrade block");
+            attributes.no_tx_pool = Some(true);
+        }
+
+        // Do not include transactions in the first Interop block.
+        if self.cfg.is_first_interop_block(attributes.payload_attributes.timestamp) {
+            info!(target: "sequencer", "Sequencing interop upgrade block");
+            attributes.no_tx_pool = Some(true);
+        }
```
refcell, line 141: Is there a way we can make this more future-proof so we don't have to come back here and update?

Perhaps we could create a `FirstBlock` variant that is returned by a new `first_block(timestamp)` method on the rollup config. Then we could match against the variants like `FirstBlock::Interop`. Then, when a new hardfork is added to the `HardforkConfig`, the `FirstBlock` must be updated or a compile error occurs which will require this to be updated due to the `match`, so it isn't silently forgotten. Something like that.

crates/node/service/src/actors/derivation.rs
```diff
@@ -310,7 +310,6 @@ where
     P: Pipeline + SignalReceiver,
 {
     /// Creates a new instance of the [DerivationActor].
-    #[allow(clippy::too_many_arguments)]
```
refcell, line 313: nice

refcell (approved):
Happy to merge as is, we can follow up with fixes

PR #2250, opened 2025-06-24
fix(supervisor/core): consistency check

dhyaniarun1993:


crates/supervisor/core/src/config/rollup_config_set.rs
```diff
@@ -71,9 +53,10 @@ impl RollupConfig {
     }
 
     /// Creates a new [`RollupConfig`] with the given genesis and block time.
-    pub fn new_from_rollup_config(config: kona_genesis::RollupConfig, l1_time: u64) -> Self {
+    pub fn new_from_rollup_config(config: kona_genesis::RollupConfig, l1_block: BlockInfo) -> Self {
+        // todo: cross check the l1_block with the config
```
theochap, line 57: Just checking, are we going to address that as part of this sprint? Otherwise, we might as well open an issue for it to not loose track of it

theochap (commented):
Looks good. Feel free to merge once you've opened the relevant issues

crates/supervisor/core/src/config/rollup_config_set.rs
dhyaniarun1993: https://github.com/op-rs/kona/issues/2254

theochap (approved):

Issue #2176, opened 2025-06-17
[TRACKING] chore(node): codebase cleanups


theochap:
## Description

This is a tracking issue for some codebase cleanups / tech-debt removal inside the kona-node.

## Items

- #2163 
- #2175 
- #2266
- #2271 
- #2284
- #2302 
- #2303 
- #2308
- #2310
- #2315

theochap:
The `RollupNodeService` is in a good enough state to close this one. We can open a new issue to spike further refactoring improvements.

PR #2266, opened 2025-06-24
chore(node/service): remove process method from NodeActor trait

theochap:
## Description

This PR removes the `process` method from `NodeActor` trait to simplify the structure of the `NodeActor`

Progress to #2176

### DIFF
--- a/crates/node/service/src/actors/derivation.rs
+++ b/crates/node/service/src/actors/derivation.rs
@@ -209,64 +209,6 @@ where
             }
         }
     }
-}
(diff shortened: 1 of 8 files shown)


clabby (approved):
Good start. I think a nice follow-up would be to think about how these `Inbound<X>Event` enums actually come into play here. Right now they're really just creating an unnecessary (but minimal) memory overhead, for example in the `EngineActor` where we're Boxing big values to avoid the big enum.

PR #2269, opened 2025-06-24
chore(ci): Propagate features to `dev-dependencies`

clabby:
## Overview

Propagates features within `dev-dependencies`.


crates/protocol/derive/Cargo.toml
```diff
@@ -56,13 +56,14 @@ op-alloy-consensus = { workspace = true, features = ["k256"] }
 
 [features]
 default = []
-metrics = [ "dep:metrics" ]
+metrics = [ "dep:metrics", "kona-derive/metrics" ]
```
clabby, line 59: hmm. I believe we do actually want to exclude the dev dependencies @refcell.

We use this trick where we add the `self` crate as a `dev-dependency` to enable `test-utils` (#1442), and that's incompatible with automatic feature propagation.

We could always add extra ignored features to the zepter config, but might as well disable propagation for dev-deps?

refcell: I'd agree with you here. Disable for dev-deps. We don't need to ensure feature propagation to dev dependencies

PR #2270, opened 2025-06-24
feat(derive): Managed Traversal Stage

### LINKED ISSUE #2227: feat(protocol/derive): Managed L1 Traversal Stage

refcell:
### Description

Splits out the new `ManagedTraversal` stage from #2232.

Closes #2227.

### DIFF
--- a/crates/protocol/derive/src/stages/traversal/polling.rs
+++ b/crates/protocol/derive/src/stages/traversal/polling.rs
@@ -162,7 +162,7 @@ impl<F: ChainProvider + Send> SignalReceiver for PollingTraversal<F> {
             }
             Signal::ProvideBlock(_) => {
                 /* Not supported in this stage. */
-                warn!(target: "traversal", "ProvideBlock signal not supported in L1Traversal stage.");
+                warn!(target: "traversal", "ProvideBlock signal not supported in PollingTraversal stage.");
                 return Err(PipelineError::UnsupportedSignal.temp());
             }
             _ => {}


clabby (approved):
Stage itself LGTM here. Good looks on the new test helper.

A few links in the book go out of date with the `L1Traversal` stage's type rename, but can be addressed in a followup.

PR #2271, opened 2025-06-24
chore(node/service): refactor node actor trait. take context out of the actor

theochap:
## Description

First attempt to decouple channel handling from the individual actors. This PR merely refactors the `NodeActor` trait to add a `Context` associated type that gets passed to the `start` method in the trait.

I think from here we can make the structure simpler by:
- Refactoring the `process` private methods into separate methods. That should remove the extra arguments for the `EngineActor`
- Abstracting the actor building process behind the `new` methods of the individual actors. In particular, split up the context using inbound/outbound associated types and define channels inside the actor's `new` methods.

### DIFF
--- a/crates/node/service/src/actors/derivation.rs
+++ b/crates/node/service/src/actors/derivation.rs
@@ -1,6 +1,6 @@
 //! [NodeActor] implementation for the derivation sub-routine.
 
-use crate::{Metrics, NodeActor};
+use crate::{Metrics, NodeActor, actors::ActorContext};
 use async_trait::async_trait;
 use kona_derive::{
     ActivationSignal, Pipeline, PipelineError, PipelineErrorKind, ResetError, ResetSignal, Signal,
@@ -12,7 +12,7 @@ use tokio::{
     select,
     sync::{mpsc, oneshot, watch},
 };
-use tokio_util::sync::CancellationToken;
+use tokio_util::sync::{CancellationToken, WaitForCancellationFuture};
 
 /// The [NodeActor] for the derivation sub-routine.
 ///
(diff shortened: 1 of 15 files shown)


clabby (commented):
I'm generally a fan of this pattern 😄 

What do you think about adding builders for the `ActorContext` implementations, and instead passing the actor contexts directly into `new` (rather than all the channels?)

Right now, the `new` functions are still very bloated.

crates/node/service/src/actors/engine/actor.rs
```diff
@@ -296,45 +360,45 @@ impl NodeActor for EngineActor {
...
-                reset = self.reset_request_rx.recv() => {
+                reset = reset_request_rx.recv() => {
                     if reset.is_none() {
                         error!(target: "engine", "Reset request receiver closed unexpectedly");
-                        self.cancellation.cancel();
+                        cancellation.cancel();
                         return Err(EngineError::ChannelClosed);
                     }
-                    self.process(InboundEngineMessage::ResetRequest).await?;
+                    self.process(InboundEngineMessage::ResetRequest, &derivation_signal_tx, &engine_l2_safe_head_tx, &cancellation).await?;
```
clabby, line 369: same ^

clabby (approved):
Approving to unblock #2284

crates/node/service/src/actors/derivation.rs
```diff
@@ -328,15 +358,15 @@ where
                         return Ok(());
                     }
 
-                    self.process(InboundDerivationMessage::NewDataAvailable).await?;
+                    self.process(InboundDerivationMessage::NewDataAvailable, &mut engine_l2_safe_head, &el_sync_complete_rx, &attributes_out, &reset_request_tx).await?;
                 }
-                _ = self.engine_l2_safe_head.changed() => {
-                    self.process(InboundDerivationMessage::SafeHeadUpdated).await?;
+                _ = engine_l2_safe_head.changed() => {
+                    self.process(InboundDerivationMessage::SafeHeadUpdated, &mut engine_l2_safe_head, &el_sync_complete_rx, &attributes_out, &reset_request_tx).await?;
```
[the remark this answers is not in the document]
theochap: Will address that as a follow up of #2284

crates/node/service/src/actors/engine/actor.rs
theochap: Will address that as a follow up of https://github.com/op-rs/kona/pull/2284

PR #2284, opened 2025-06-25
chore(node/service): refactor and simplify the `RollupNodeService` trait

theochap:
## Description

This PR builds up on the changes introduced by #2271 to simplify/unify further the structure of the node service crates. In particular this PR:

- Adds two associated types to the `NodeActor` trait. `OutboundData` and `State` which are used to specify a new build method for the actor that aims to replace single calls of the `new` methods. The build method has the following signature: 
```
fn build(initial_state: Self::State) -> (Self::OutboundData, Self);
```

- Takes out the output channels from the `ActorContext` back into the `Actor` struct. This allows to drastically reduce the number of arguments to provide to the `build` methods and remove the need to define channels inside the implementation of the `RollupNodeService` trait. Instead this is done inside the `NodeActor` themselves
- Add an associated type to the `RollupNodeService` for each actor and call the build methods from the associated types. That will make it much easier to have a modular architecture where we can easily swap out components if needed

## Follow-ups

Most of the heavy lifting is done in this PR, but I would like to do some quick follow-ups to refactor/simplify/improve the inner structure of actors even further.
We should also think about which types we want to expose to the public api and try to cut down the number of public exports in our libraries.

### DIFF
--- a/crates/node/service/src/actors/derivation.rs
+++ b/crates/node/service/src/actors/derivation.rs
@@ -1,6 +1,6 @@
 //! [NodeActor] implementation for the derivation sub-routine.
 
-use crate::{Metrics, NodeActor, actors::ActorContext};
+use crate::{Metrics, NodeActor, actors::CancellableContext};
 use async_trait::async_trait;
 use kona_derive::{
     ActivationSignal, Pipeline, PipelineError, PipelineErrorKind, ResetError, ResetSignal, Signal,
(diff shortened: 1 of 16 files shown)


crates/node/service/src/actors/l1_watcher_rpc.rs
```diff
@@ -29,26 +29,43 @@ use tokio_util::sync::{CancellationToken, WaitForCancellationFuture};
 /// An L1 chain watcher that checks for L1 block updates over RPC.
 #[derive(Debug)]
 pub struct L1WatcherRpc {
+    config: L1WatcherRpcState,
```
clabby, line 32: `config` -> `state`?

crates/node/service/src/actors/traits.rs
```diff
@@ -16,12 +16,21 @@ pub trait ActorContext: Send {
...
-pub trait NodeActor {
+pub trait NodeActor: Send + 'static {
     /// The error type for the actor.
     type Error: std::fmt::Debug;
     /// The communication context used by the actor.
+    /// These are the channels that the actor will use to receive messages from other actors.
     type Context: ActorContext;
+    /// The outbound communication channels used by the actor.
+    /// These are the channels that the actor will use to send messages to other actors.
+    type OutboundData: Sized;
```
clabby, line 27: Small nit - `Context` isn't really clear, given that it's now purposed to just be the inbound channels. Maybe we rename it to `InboundData`, and rename `ActorContext` to `CancellableActor` or something?

crates/node/service/src/actors/derivation.rs
```diff
@@ -358,15 +384,15 @@ where
...
                 }
                 _ = engine_l2_safe_head.changed() => {
-                    self.process(InboundDerivationMessage::SafeHeadUpdated, &mut engine_l2_safe_head, &el_sync_complete_rx, &attributes_out, &reset_request_tx).await?;
+                    self.state.process(InboundDerivationMessage::SafeHeadUpdated, &mut engine_l2_safe_head, &el_sync_complete_rx, &self.attributes_out, &self.reset_request_tx).await?;
                 }
                 _ = &mut el_sync_complete_rx, if !el_sync_complete_rx.is_terminated() => {
                     info!(target: "derivation", "Engine finished syncing, starting derivation.");
                     // Optimistically process the first message.
-                    self.process(InboundDerivationMessage::NewDataAvailable, &mut engine_l2_safe_head, &el_sync_complete_rx, &attributes_out, &reset_request_tx).await?;
+                    self.state.process(InboundDerivationMessage::NewDataAvailable, &mut engine_l2_safe_head, &el_sync_complete_rx, &self.attributes_out, &self.reset_request_tx).await?;
```
clabby, line 395: These are still a bit over-loaded. Not necessary for this PR, but maybe we split up `process` and remove the `InboundDerivationMessage` enum.

crates/node/service/src/actors/engine/actor.rs
```diff
@@ -366,39 +378,39 @@ impl NodeActor for EngineActor {
                         cancellation.cancel();
                         return Err(EngineError::ChannelClosed);
                     }
-                    self.process(InboundEngineMessage::ResetRequest, &derivation_signal_tx, &engine_l2_safe_head_tx, &cancellation).await?;
+                    self.state.process(InboundEngineMessage::ResetRequest, &self.derivation_signal_tx, &self.engine_l2_safe_head_tx, &mut finalizer, &cancellation).await?;
```
clabby, line 381: Same story - a bit over-loaded. Not necessary for this PR, but maybe we split up process and remove the`InboundEngineMessage` enum.

clabby (approved):
Really like this, nice work.

Follow-up items that'd be really nice are:
1. Removing the `*Launcher` patterns entirely
2. Removing `process` methods + `Inbound<Actor>Message` enums, splitting up action dispatch.

crates/node/service/src/actors/derivation.rs
theochap: Totally agree, will do that in a follow-up

crates/node/service/src/actors/engine/actor.rs
theochap: Totally agree, will do that in a follow-up

PR #2294, opened 2025-06-25
chore(node/tests): deprecate devnet-sdk and reactivate RPC endpoint tests

theochap:
## Description

This PR deprecates the devnet-sdk and reactivates the RPC endpoint tests of the kona-node.


refcell:
Looks like a p2p test failed in ci

PR #2303, opened 2025-06-26
chore(node/service): move actor component building process inside the actors

theochap:
## Description

This PR does some structural services to the `RollupNodeService` trait, in particular:
- It ensures we're only passing static configurations/builders to the `NodeActor`s and offsets the building process to the `NodeActor` itself. 
- Each actor now has an associated builder. This unifies the terminology between `*Builder`s and `*Launcher`s which seemed to refer to the same concept.
- This allows to simplify the `RollupNodeService` trait by having:
    - less constraints on the configurations accepted by the `NodeActors`.
    - having simpler associated methods which are basically just getters over some fields of the `RollupNode`. No need for async associated methods anymore or error handling.
    - The only `.await` we're using is to spin up the actors once they're initiated.

### DIFF
--- a/crates/node/service/src/service/core.rs
+++ b/crates/node/service/src/service/core.rs
@@ -1,16 +1,16 @@
 //! The core [`RollupNodeService`] trait
-
 use crate::{
     DerivationContext, EngineContext, L1WatcherRpcContext, L2Finalizer, NetworkContext, NodeActor,
     NodeMode, RpcContext, RuntimeContext, SupervisorActorContext, SupervisorExt,
     actors::{
         DerivationOutboundChannels, EngineOutboundData, L1WatcherRpcOutboundChannels,
-        NetworkOutboundData, PipelineBuilder, RuntimeOutboundData, SupervisorOutboundData,
+        NetworkOutboundData, PipelineBuilder, RuntimeOutboundData, SequencerActorState,
+        SequencerContext, SequencerOutboundData, SupervisorOutboundData,
     },
     service::spawn_and_wait,
 };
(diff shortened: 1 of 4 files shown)


crates/node/service/src/service/core.rs
```diff
@@ -96,82 +79,60 @@ pub trait RollupNodeService {
...
-    /// Returns the [`RuntimeState`] for the node.
-    fn runtime(&self) -> Option<&RuntimeState>;
-
-    /// Returns the [`EngineLauncher`]
-    fn engine(&self) -> EngineLauncher;
-
-    /// Returns the [`RpcLauncher`] for the node.
-    fn rpc(&self) -> RpcLauncher;
+    /// Returns the [`RpcBuilder`] for the node.
+    fn rpc(&self) -> RpcBuilder;
```
refcell, line 118: We suffix all the other builder methods with `_builder`, let's add that here for consistency.

crates/node/service/src/service/standard/builder.rs
```diff
@@ -168,18 +163,18 @@ impl RollupNodeBuilder {
...
             l1_provider,
             l1_beacon,
             l2_provider,
-            engine_launcher,
-            rpc_launcher,
+            engine_builder: engine_launcher,
+            rpc_builder: rpc_launcher,
             p2p_config,
-            runtime_launcher,
+            runtime_builder: runtime_launcher,
```
refcell, line 174: Let's rename all of these from launchers to builders.

refcell (approved):
LGTM, few small nits but not generally blocking

PR #2308, opened 2025-06-26
chore(node/rpc): unify RPC actor with the other actors

theochap:
## Description

This PR refactors the RPC actor in light of #2303 to unify its interactions with the other actors. Simplifies a little bit the way we handle RPC arguments by removing the `RpcBuilder` and move the `RpcModule` handling inside the `RpcActor`


crates/node/service/src/service/standard/error.rs
theochap, line 1: Those variants are not needed anymore. Note that we may be able to do some error handling by propagating the errors from the actors. We can try and do it as a follow-up.

bin/node/src/commands/net.rs
```diff
@@ -90,9 +93,9 @@ impl NetCommand {
                         Err(e) => debug!(target: "net", "Failed to receive unsafe payload: {:?}", e),
                     }
                 }
-                _ = interval.tick() => {
+                _ = interval.tick(), if tx.is_some() => {
                     let (otx, mut orx) = tokio::sync::oneshot::channel();
-                    if let Err(e) = tx.send(P2pRpcRequest::PeerCount(otx)).await {
+                    if let Err(e) = tx.as_ref().unwrap().send(P2pRpcRequest::PeerCount(otx)).await {
```
refcell, line 98: Can we match this in the select to be `, Some(sender) = &tx => {` or does that not play nicely?

If not, let's at least move this up a few lines and avoid using `unwrap()`:

```rust
let Some(sender) = tx else {
    unreachable!("tx must be some");
};
```

crates/node/p2p/src/net/driver.rs
```diff
@@ -235,7 +232,7 @@ impl Network {
                             warn!(err = ?send_err, "Impossible to send a request to the discovery handler. The channel connection is dropped.");
                         }
                     },
-                    req = rpc.recv() => {
+                    req = rpc.as_mut().unwrap().recv(), if rpc.is_some() => {
```
refcell, line 235: Same here

refcell, line 235: I think we can use an [`OptionFuture`](https://docs.rs/futures/latest/futures/future/struct.OptionFuture.html) here instead of unwrapping.

crates/node/service/src/actors/rpc.rs
```diff
@@ -14,7 +23,7 @@ pub enum RpcActorError {
     RegisterHealthz(#[from] RegisterMethodError),
     /// Failed to launch the RPC server.
     #[error("Failed to launch the RPC server")]
```
refcell, line 25: Can we do `#[error(transparent)]` on the `std::io::Error` here?

crates/node/service/src/actors/rpc.rs
```diff
@@ -27,13 +36,35 @@ pub enum RpcActorError {
...
+    /// The network rpc sender.
+    network: mpsc::Sender<P2pRpcRequest>,
+    /// The l1 watcher queries sender.
+    l1_watcher_queries: mpsc::Sender<L1WatcherQueries>,
+    /// The engine query sender.
+    engine_query: mpsc::Sender<EngineQueries>,
 }
 
 impl RpcActor {
     /// Constructs a new [`RpcActor`] given the [`RpcBuilder`] and [`CancellationToken`].
```
refcell, line 49: Comment needs an update here, can remove the `CancellationToken` bit.

crates/node/p2p/src/net/driver.rs
```diff
@@ -235,7 +232,9 @@ impl Network {
                             warn!(err = ?send_err, "Impossible to send a request to the discovery handler. The channel connection is dropped.");
                         }
                     },
-                    req = rpc.recv() => {
+                    // Note that we're using `if rpc.is_some()` to ensure that the branch does not always immediately resolve
+                    // to avoid CPU contention.
```
refcell, line 236: Nice note

refcell (approved):
Nice!

PR #2322, opened 2025-06-28
feat(node/service): Propagated errors raised during NodeActor::start

### LINKED ISSUE #2311: feat(node/service): propagate errors raised during `NodeActor::start` up to the binary level
## Description

We currently are silently exiting with emitting error logs inside the `RollupNodeService` `spawn_and_await` macro. See:

```
        // Check if the actor is present, and spawn it if it is.
        $(
            if let Some((actor, context)) = $actor {
                task_handles.spawn(async move {
                    if let Err(e) = actor.start(context).await {
                        return Err(format!("{e:?}"));
                    }
                    Ok(())
                });
            }
        )*

        while let Some(result) = task_handles.join_next().await {
            match result {
                Ok(Ok(())) => { /* Actor completed successfully */ }
                Ok(Err(e)) => {
                    tracing::error!(target: "rollup_node", "Critical error in sub-routine: {e}");
                    // Cancel all tasks and gracefully shutdown.
                    $cancellation.cancel();
                }
                Err(e) => {
                    tracing::error!(target: "rollup_node", "Task join error: {e}");
                    // Cancel all tasks and gracefully shutdown.
                    $cancellation.cancel();
                }
            }
        }
```

Instead, we should propagate the errors to the caller and let them decide how to handle it. This would help with error handling and testing coverage.

PoulavBhowmick03:
Fixes #2311

### DIFF
--- a/bin/node/src/commands/node.rs
+++ b/bin/node/src/commands/node.rs
@@ -232,7 +232,11 @@ impl NodeCommand {
             .with_supervisor_rpc_config(supervisor_rpc_config.unwrap_or_default())
             .build()
             .start()
-            .await;
+            .await
+            .map_err(|e| {
+                error!(target: "rollup_node", "Failed to start rollup node service: {e}");
+                anyhow::anyhow!("{}", e)
+            })?;
 
         Ok(())
     }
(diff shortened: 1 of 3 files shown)


theochap (approved):
Good start! I will approve this one. As a follow-up it would be great to have a more structured error type to return from `start`. Potentially an enum we can macro-generate from the actor's list or a `dyn Box<...>` error type.

PR #2331, opened 2025-07-01
fix (supervisor/core): correctly interpreting `derivationUpdate` and `derivationOriginUpdate`

### LINKED ISSUE #2086: feat(supervisor/core): Analyse `supervisor_allSafeDerivedAt` correctness
# Description

Based on the conversation [here](https://discord.com/channels/1244729134312198194/1381976003185742005), make sure the `supervisor_allSafeDerivedAt` is behaving as expected especially for the empty L1 block.

itschaindev:
Closes #2086 

The PR:
- Introduces new table `BlockTraversal` to store all L1 blocks, even if there are no L2 blocks derived from it.
- Add respective trait methods.
- Modified `latest_derived_block_pair` to send the latest derived block from `DerivedBlocks` and latest source block from `BlockTraversal`.
- Exclusively dealing with `derivationUpdate` and `derivationOriginUpdate` i.e. updating latest derived on `derivationUpdate` and updating latest source on `derivationOriginUpdate`.
- The e2e test `TestFinalizedHeadAdvancing` passes now. 

TODO: (in another PR)
- Remove `current_l1` and its usage. No longer required.
- ~~For now, `check_node_consistency()` is commented in this PR. #2330 takes care of that.~~

~~Currently in draft: Tests are missing.~~


crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -233,23 +287,10 @@ where
...
-                    error!(
-                      target: "supervisor_storage",
-                      incoming_derived_block_pair = %incoming_pair,
-                      %err,
-                      "Failed to get derived block numbers for source block"
-                    );
-                    return Err(err);
-                }
-            };
+        let mut block = self.save_source_block(incoming_pair.source)?;
```
dhyaniarun1993, line 290: I think we shouldn't save the source block here. While saving the derived block, we should assume that `save_source_block` should have been called.

crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -263,18 +304,69 @@ where
...
+
+    /// Saves a source block to the database.
+    /// If the source block already exists, it does nothing.
+    /// If the source block does not exist, it creates a new [`SourceBlockTraversal`] and saves it
+    /// to the database.
+    pub(crate) fn save_source_block(
+        &self,
+        source: BlockInfo,
+    ) -> Result<SourceBlockTraversal, StorageError> {
+        match self.get_block_traversal(source.number) {
```
dhyaniarun1993, line 328: We don't need to perform 2 fetch here `get_block_traversal` and `latest_source_block`. 

We can just fetch `latest_source_block` and do the following
- check if it is same return Ok
- if incoming block is child of `latest_source_block`, then save the incoming
- else error

dhyaniarun1993 (approved):

PR #2342, opened 2025-07-06
feat(bin/node): Change `sequencer.enabled` -> `mode`

### LINKED ISSUE #2307: feat(bin/node): Change `sequencer.enabled` -> `NodeMode`
## Overview

Currently, we use the `sequencer.enabled` flag to determine which mode the node is operating under. Instead, we should remove this flag, and have a flag that directly takes in `NodeMode`.

Once this is changed, we need to add `with_mode` to the builder:
https://github.com/op-rs/kona/blob/2cf7a4f5547b09c6cd8d67307b66791fe944e6e0/bin/node/src/commands/node.rs#L214-L227

clabby:
## Overview

Changes the `sequencer.enabled` flag to a more generic `mode` flag that can be used to specify different modes of operation for the node. Also wires this flag up to the `NodeBuilder`.

closes #2307


crates/node/service/src/service/mode.rs
```diff
@@ -14,6 +14,21 @@ pub enum NodeMode {
...
+        match s.to_lowercase().as_str() {
+            "validator" => Ok(Self::Validator),
+            "sequencer" => Ok(Self::Sequencer),
+            _ => Err(format!(
+                "Invalid node mode: '{}'. Valid options are: 'validator', 'sequencer'",
+                s
+            )),
+        }
+    }
+}
```
refcell, line 30: This should be a derive more [`FromStr`](https://docs.rs/derive_more/latest/derive_more/derive.FromStr.html)

refcell (changes requested):
Nice

refcell (approved):

PR #2344, opened 2025-07-06
chore(workspace): Bump dependencies

clabby:
## Overview

Bumps the workspace's dependencies.

### DIFF
--- a/crates/node/p2p/src/gossip/driver.rs
+++ b/crates/node/p2p/src/gossip/driver.rs
@@ -224,7 +224,7 @@ where
 
     fn handle_gossip_event(&mut self, event: Event) -> Option<OpNetworkPayloadEnvelope> {
         match event {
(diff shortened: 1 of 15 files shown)


crates/node/p2p/src/gossip/driver.rs
```diff
@@ -224,7 +224,7 @@ where
 
     fn handle_gossip_event(&mut self, event: Event) -> Option<OpNetworkPayloadEnvelope> {
         match event {
-            Event::Gossipsub(e) => return self.handle_gossipsub_event(e),
+            Event::Gossipsub(e) => return self.handle_gossipsub_event(*e),
```
refcell, line 227: We could just pass in a reference here and change the parameter type, but for the sake of keeping this bump pr lean, this is fine

refcell: Ah I see it's boxed

refcell (approved):
Very nice

PR #2360, opened 2025-07-07
feat(node/sequencer): move node mode from rollup node service trait to engine actor.

theochap:
## Description

Small PR to move the node mode parameter from the `RollupNodeService` trait to the `EngineActor`. This also improves the safety of the engine select statement branch for incoming block building requests by being able to catch communication drops from the sequencer engine.

### DIFF
--- a/crates/node/service/src/service/core.rs
+++ b/crates/node/service/src/service/core.rs
@@ -128,12 +128,10 @@ pub trait RollupNodeService {
         let cancellation = CancellationToken::new();
 
         // Create the DA watcher actor.
-        let da_watcher_builder = self.da_watcher_builder();
         let (L1WatcherRpcInboundChannels { inbound_queries: da_watcher_rpc }, da_watcher) =
-            Self::DataAvailabilityWatcher::build(da_watcher_builder);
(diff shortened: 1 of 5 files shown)


crates/node/service/src/service/core.rs
```diff
@@ -230,7 +227,7 @@ pub trait RollupNodeService {
                 Some((da_watcher, da_watcher_context)),
                 Some((derivation, derivation_context)),
                 Some((engine, engine_context)),
-                self.mode().is_sequencer().then_some((sequencer, sequencer_context))
+                sequencer_context.map(|context| (sequencer, context)),
```
clabby, line 230: This is a bit implicit / hard to read, vs the old version

clabby (changes requested):
I like where you're going with this, specifically the changes in the `EngineActor`. Rather than removing `NodeMode` from the `RollupNodeService` trait, can we instead retain it to allow for more readable branches in `RollupNodeService`?

crates/node/service/src/service/core.rs
theochap: Reverted this change!

clabby (approved):

PR #2364, opened 2025-07-07
chore(node/rpc): remove rpc disabled field in rpc builder

theochap:
## Description

A simple PR that removes the `rpc_disabled` field inside the `RpcBuilder`. This field was redundant since the `RpcBuilder` is now provided as an `Option` inside the `RollupNodeService`


clabby (approved):
much nicer. thx!

Issue #2272, opened 2025-06-25
test(supervisor): preinterop support


dhyaniarun1993:
# Description

Test the pre-interop support and fix the bug that comes along the way.

PR #2368, opened 2025-07-08
feat(supervisor): pre-interop db support

### LINKED ISSUE #2351: feat(supervisor/core): initialise database post interop

dhyaniarun1993:
Closes #2351 
Part of #2272


crates/supervisor/storage/src/providers/log_provider.rs
```diff
@@ -68,6 +68,24 @@ where
...
+            }
+            error!(
+                target: "supervisor_storage",
+                %stored_block,
+                incoming_block = %block,
+                "Incoming log block is not consistent with the stored log block",
+            );
+            return Err(StorageError::ConflictError(
+                "incoming log block is not consistent with the stored log block".to_string(),
+            ))
```
itschaindev, line 86: Should we trigger a reset in this case?

crates/supervisor/storage/src/chaindb.rs
```diff
@@ -97,16 +89,30 @@ impl DerivationStorageReader for ChainDb {
     }
 
     fn latest_derivation_state(&self) -> Result<DerivedRefPair, StorageError> {
-        self.observe_call("latest_derived_block_pair", || {
+        self.observe_call("latest_derivation_state", || {
```
itschaindev, line 92: There are a couple of other places in the code where we should change this and `save_derived_block_pair` to new names for consistency.

crates/supervisor/storage/src/providers/log_provider.rs
dhyaniarun1993: Done

itschaindev: I'm unable to find where it is triggered. Can you please point out?

itschaindev (approved):

PR #2376, opened 2025-07-08
chore(node/net): integrate the network driver inside the network actor

theochap:
## Description

This PR refactors the network driver and moves it inside the network actor. This reduces the levels of indirection and allow the network actor to comply more with the new builder refactor of the node services by avoiding to define channels inside the driver.

Changes:

- Removes the `net` module inside `kona-p2p`. Moves the relevant files to the `kona-node-services` crate.
- Removes the `broadcast` module inside the network actor. This can now be directly handled using a mpsc channel inside the network actor to implement the notion of queue. Outer interactions with the network actor happen through the `NetworkActor` context (see for instance https://github.com/op-rs/kona/compare/main...theo/send-payloads-to-network#diff-b4e04fd02334c1ae45e956ad21312e81efacba5a7c1d477b176711dbae458c32R87-R126 or https://github.com/op-rs/kona/compare/main...theo/send-payloads-to-network#diff-00c31e9f9de711af31ebbd3b81f610ea12ea100a6a8d0db6a2ef86f52f298849L80-R75)
- Divides the network actor module with the following pattern to propagate the lower level abstractions (resp `Discv5Driver`/`Discv5Handler`/`GossipDriver`):
```
NodeBuilder ->build NodeDriver ->(async) start NodeHandler
```


crates/node/service/src/actors/network/mod.rs
```diff
@@ -0,0 +1,13 @@
...
+mod builder;
+mod driver;
+mod error;
+mod handler;
+
+pub use actor::{NetworkActor, NetworkActorError, NetworkContext, NetworkInboundData};
+pub use builder::NetworkBuilder;
+pub use driver::{NetworkDriver, NetworkDriverError};
+pub use error::NetworkBuilderError;
+pub use handler::NetworkHandler;
```
refcell, line 13: Can we re-arrange this so each module is above its public `use` to keep it consistent with the rest of kona?

crates/node/service/src/actors/network/handler.rs
```diff
@@ -0,0 +1,240 @@
...
+            })
+            .await
+        {
+            warn!(err = ?send_err, "Impossible to send a request to the discovery handler. The channel connection is dropped.");
+        }
+    }
+}
+
+#[async_trait]
+impl NodeActor for NetworkActor {
```
refcell, line 115: Shouldn't this be in the `actor.rs` module? Any reason to have it live in `handler`?

crates/node/service/src/actors/network/handler.rs
```diff
@@ -0,0 +1,240 @@
...
+    /// The gossip driver.
+    pub gossip: GossipDriver<ConnectionGater>,
+    /// The discovery handler.
+    pub discovery: Discv5Handler,
+    /// The receiver for the ENRs.
+    pub enr_receiver: mpsc::Receiver<Enr>,
+    /// The sender for the unsafe block signer.
+    pub unsafe_block_signer_sender: watch::Sender<Address>,
+    /// The peer score inspector. Is used to ban peers that are below a given threshold.
+    pub peer_score_inspector: tokio::time::Interval,
```
refcell, line 31: Is this used?

crates/node/p2p/src/lib.rs
```diff
@@ -13,8 +13,8 @@ extern crate tracing;
 mod metrics;
 pub use metrics::Metrics;
 
-mod net;
-pub use net::{Broadcast, Config, Network, NetworkBuilder, NetworkBuilderError};
+mod config;
+pub use config::Config;
```
refcell, line 17: Does it make sense to have the config live anymore in `kona_p2p` if it's only used by the `NetworkDriver` which is now moved into the `kona-node-service` actors module?

crates/node/service/src/actors/network/actor.rs
```diff
@@ -55,17 +59,26 @@ pub struct NetworkInboundData {
...
+        let actor = Self { builder: driver, signer: signer_rx, rpc: rpc_rx, publish_rx };
+        let outbound_data =
+            NetworkInboundData { signer: signer_tx, rpc: rpc_tx, unsafe_blocks: publish_tx };
         (outbound_data, actor)
     }
+
+    /// Returns the [`NetworkBuilder`] that can be used to construct the [`NetworkActor`].
+    pub fn builder(config: Config) -> NetworkBuilder {
+        NetworkBuilder::from(config)
+    }
```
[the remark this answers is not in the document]
theochap: Sounds good! I think I just copied that one over, but I agree that this is not needed anymore

crates/node/service/src/actors/network/handler.rs
theochap: No specific reason, I just wanted to avoid crowding the `actor.rs` module

crates/node/service/src/actors/network/driver.rs
```diff
@@ -0,0 +1,52 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum NetworkDriverError {
+    /// An error occurred starting the libp2p Swarm.
+    #[error("error starting libp2p Swarm")]
+    GossipStartError(#[from] TransportError<std::io::Error>),
+}
+
+impl NetworkDriver {
+    /// The frequency at which to inspect peer scores to ban poorly performing peers.
+    pub(super) const PEER_SCORE_INSPECT_FREQUENCY: Duration = Duration::from_secs(1);
```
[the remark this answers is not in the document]
theochap: Can open an issue for it, I think we can configure it indeed!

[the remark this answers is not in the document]
theochap: Oh actually it is already defined as a static constant here https://github.com/op-rs/kona/blob/489b3777a3019a913e808276583ef45b40d9459b/crates/node/p2p/src/gossip/config.rs#L54-L56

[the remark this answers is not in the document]
theochap: How about we keep it constant for now then?

PR #2378, opened 2025-07-08
feat(node/net): rewire unsafe block sender from engine to network

### LINKED ISSUE #2244: feat(node/service): Send built payloads to `NetworkActor`
## Overview

The `NetworkActor` already has support for signing and publishing payloads. When the sequencer receives a built payload back from the engine, it must forward it to the `NetworkActor` for it to be propagated to the rest of the network.

https://github.com/op-rs/kona/blob/a8274156f4c211745ff64b06a25b21ac374c7446/crates/node/p2p/src/net/driver.rs#L142-L171

theochap:
## Description

Rewire the sequencer unsafe payload sender from the engine to the network. 

Previously, the sequencer was sending the unsafe blocks it received from the network `BuildTask` back to the network to call `InsertUnsafeTask`. Since the `BuildTask` inserts the newly built payloads to the engine state (like what the `InsertUnsafeTask` would do), this resulted in a no-op.

Instead, the sequencer should send the payloads to the network to gossip them to the p2p mesh.

Close #2244


crates/node/service/src/actors/network/actor.rs
```diff
@@ -62,7 +62,7 @@ pub struct NetworkInboundData {
     /// Handler for RPC Requests sent to the network actor.
     pub rpc: mpsc::Sender<P2pRpcRequest>,
     /// A channel to send unsafe blocks to the network actor.
```
clabby, line 64: Maybe adjust this comment to note that this channel is for use by the sequencer / `admin` RPC only

PR #2385, opened 2025-07-09
feat(supervisor/core): preinterop node support

### LINKED ISSUE #2409: feat(supervisor/core): Managed node `interop_blockRefByNumber` removed
# Description

In order to support preinterop, we need to depends on the latest `op-node` release. In the latest `op-node`, rpc method `interop_blockRefByNumber` have been removed and we need to switch to the `interop_l2BlockRefByNumber` method for fetching the block related information.

### LINKED ISSUE #2411: feat(supervisor/core): intergarte managed node `resetPreInterop` method
# Description

`resetPreInterop` method has been added to the managed node. This method is used to reset the node before the pre-interop state.

dhyaniarun1993:
Closes #2409 
Closes #2411 

Part of #2402


crates/supervisor/core/src/syncnode/resetter.rs
```diff
@@ -32,6 +32,11 @@ where
         let SuperHead { local_unsafe, cross_unsafe, local_safe, cross_safe, finalized, .. } =
             match self.get_latest_valid_super_head().await {
                 Ok(block) => block,
+                // todo: require refactor and corner case handling
```
itschaindev, line 35: This seems a bit too harsh to roll-back till pre-interop for a couple of unsafe block changes. I think we need to abstract away the feature of finding common ancestor and use it whenever we want to roll-back.

crates/supervisor/core/src/supervisor.rs
```diff
@@ -143,9 +143,13 @@ impl Supervisor {
         for (chain_id, config) in self.config.rollup_config_set.rollups.iter() {
             // Initialise the database for each chain.
             let db = self.database_factory.get_or_create_db(*chain_id)?;
-            let anchor = config.genesis.get_anchor();
-            db.initialise_log_storage(anchor.derived)?;
-            db.initialise_derivation_storage(anchor)?;
+            let interop_time = config.interop_time;
+            let derived_pair = config.genesis.get_derived_pair();
+            if config.is_interop(derived_pair.derived.timestamp) {
```
itschaindev, line 148: How are we initializing if not preinterop?

dhyaniarun1993: In that case, we depend on the unsafe and safe block events.

crates/supervisor/core/src/syncnode/resetter.rs
dhyaniarun1993: Here we are resetting the node to preinterop state only when the supervisor database isn't initialised(we haven't indexed any information yet).

PR #2388, opened 2025-07-09
feat(node/engine): refactor engine tasks. fix sequencer startup logic

### LINKED ISSUE #2372: feat(node/service): Sequencer startup routine
## Overview

The verifier node startup routine, at a high level, is as follows:
1. Receive first unsafe payload from the sequencer
2. Forward the unsafe payload to the `EngineActor`, [enqueue an `InsertUnsafeTask`](https://github.com/op-rs/kona/blob/8daa1ccf91e745d9df1a997f53bcaa076d2e4e2b/crates/node/service/src/actors/engine/actor.rs#L439-L451).
3. Allow EL sync to kick off, continuously insert payloads as they're received from the sequencer.
4. When EL sync finishes, [reset the engine (initializes the engine state)](https://github.com/op-rs/kona/blob/8daa1ccf91e745d9df1a997f53bcaa076d2e4e2b/crates/node/service/src/actors/engine/actor.rs#L307-L327).

The sequencer startup routine must be different, as it will be extending the chain itself, and cannot rely on the first foreign payload insertion as the route to kick off EL sync and the initial engine reset.

---

Additionally, the only codepath that can update EL sync's state is in the `InsertUnsafeTask`:

https://github.com/op-rs/kona/blob/8daa1ccf91e745d9df1a997f53bcaa076d2e4e2b/crates/node/engine/src/task_queue/tasks/insert/task.rs#L53-L68

This is required to kickstart derivation, over in:

https://github.com/op-rs/kona/blob/8daa1ccf91e745d9df1a997f53bcaa076d2e4e2b/crates/node/service/src/actors/engine/actor.rs#L307-L327

With the sequencer, we will never kickstart EL sync, since we only extend the existing chain. It will need to kickstart derivation after the first block has been built, sending the `sync_complete` notification to the derivation actor.

---

The output of this ticket should be a complete definition of the sequencer's startup routine, accounting for initializing the engine state with a reset and kicking off derivation, as well as its implementation.

theochap:
## Description

This PR refactors the core logic of the engine task handler. In particular, it moves duplicate code from the build and the insert unsafe tasks to the forkchoice task.

Changes:
- Add an associated `Output` type for the `EngineTaskExt` trait which allows the implementors of this trait to be composed. In particular, the `ForkchoiceTask` may returns the payload id that was inserted in the EL, which can be used by the `BuildTask` to consolidate the payload into the engine state.
- Allows the `ForkchoiceTask` to directly insert newly built payload attributes into the EL (by adding an `Option<OpAttributeWithParent>` argument to the `ForkchoiceTask`). This allows to refactors both calls to the `FCU` inside the `BuildTask` using the `ForkchoiceTask` and avoids manually handling the forkchoice updates
- Moves the `PayloadStatus` verification logic to the `ForkchoiceTask` to avoid having to perform the same logic in both the `BuildTask` and the `InsertUnsafeTask`, which allows to fix the sequencer kickstart issue described in #2372 

Close #2372 

## Potential follow-ups

- Remove the clones inside the engine tasks
- Refactor the `new_payload` logic into an engine task to remove duplicate code in the `InsertUnsafe` and `Build`Tasks

### DIFF
--- a/crates/node/engine/src/task_queue/tasks/build/task.rs
+++ b/crates/node/engine/src/task_queue/tasks/build/task.rs
@@ -2,12 +2,12 @@
 
 use super::BuildTaskError;
 use crate::{
-    EngineClient, EngineForkchoiceVersion, EngineGetPayloadVersion, EngineState, EngineTaskError,
-    EngineTaskExt, ForkchoiceTask, Metrics,
+    EngineClient, EngineGetPayloadVersion, EngineState, EngineTaskError, EngineTaskExt,
+    ForkchoiceTask, Metrics, state::EngineSyncStateUpdate,
 };
 use alloy_provider::ext::EngineApi;
 use alloy_rpc_types_engine::{
-    ExecutionPayloadFieldV2, ExecutionPayloadInputV2, ForkchoiceState, PayloadId, PayloadStatusEnum,
+    ExecutionPayloadFieldV2, ExecutionPayloadInputV2, PayloadId, PayloadStatusEnum,
 };
 use alloy_transport::RpcError;
 use async_trait::async_trait;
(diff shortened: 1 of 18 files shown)


crates/node/engine/src/task_queue/tasks/build/task.rs
```diff
@@ -322,16 +234,24 @@ impl EngineTaskExt for BuildTask {
             .into());
         }
 
-        // Send the forkchoice update through the input, with the current engine state and the
-        // payload attributes for the block building job.
-        let mut forkchoice = state.create_forkchoice_state();
-        forkchoice.head_block_hash = self.attributes.parent.block_info.hash;
```
clabby, line 328: This is sneaky, but we cannot remove this. The sequencer has the right to reorg the chain at will; We must build all blocks on top of the attributes' parent hash.

By offloading the creation of the `ForkchoiceState` to `ForkchoiceTask`, we don't get this capability.

crates/node/engine/src/task_queue/tasks/forkchoice/task.rs
```diff
@@ -1,36 +1,90 @@
...
+        client: Arc<EngineClient>,
+        rollup: Arc<RollupConfig>,
+        payload_attributes: Option<OpAttributesWithParent>,
+    ) -> Self {
+        Self { client, rollup, envelope: payload_attributes }
+    }
+
+    /// Checks the response of the `engine_forkchoiceUpdated` call, and updates the sync status if
+    /// necessary.
+    fn check_forkchoice_updated_status(
```
clabby, line 40: This check is strictly different than the checks we performed on synchronizing the local engine state beforehand. i.e., we'd never return a temporary error if the engine was syncing.

clabby (commented):
I think we're packing a bit too much into the `ForkchoiceTask` here; It now has two responsibilities:
1. Begin a block building job
2. Synchronize the rollup node's view of the forkchoice with the EL

each of which have slight semantic differences in the error handling, response validation checks, etc.

Though I am a fan of the attempt to modularize here. What do you think about separating the concerns here, and renaming `ForkchoiceTask` (as-is) to `SynchronizeTask`, and then creating a new task for kicking off block building?

crates/node/engine/src/task_queue/tasks/forkchoice/task.rs
```diff
@@ -49,24 +103,64 @@ impl EngineTaskExt for ForkchoiceTask {
         // Send the forkchoice update through the input.
         let forkchoice = state.create_forkchoice_state();
```
clabby, line 104: re: https://github.com/op-rs/kona/pull/2388/files#r2195674080

note that here, it expects that we'll start building on top of the existing unsafe head, rather than the one declared in the payload attributes we've either derived or formed within the sequencer. This can cause trouble.

crates/node/engine/src/task_queue/tasks/insert/task.rs
```diff
@@ -137,56 +117,23 @@ impl EngineTaskExt for InsertUnsafeTask {
...
         state.set_cross_unsafe_head(new_unsafe_ref);
         state.set_unsafe_head(new_unsafe_ref);
-        state.forkchoice_update_needed = false;
+
+        // Send a FCU to canonicalize the imported block.
+        ForkchoiceTask::new(Arc::clone(&self.client), self.rollup_config.clone(), None)
+            .execute(state)
+            .await?;
+
+        let total_duration = time_start.elapsed();
```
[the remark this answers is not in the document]
theochap: Yeah that's a good point, although I thought that in that case (if we have an inconsistent state view between the EL and the CL), the engine will reset (after returning `InvalidForkchoiceState`)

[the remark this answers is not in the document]
clabby: I think you're right here, actually - disregard. It should be fine to do this _so long as_ the EL eventually knows about it.

crates/node/engine/src/task_queue/tasks/build/task.rs
clabby: This is also codified in the execution layer: https://github.com/paradigmxyz/reth/blob/HEAD/crates/engine/tree/src/tree/mod.rs#L893-L905

crates/node/engine/src/task_queue/tasks/forkchoice/task.rs
```diff
@@ -1,72 +1,179 @@
...
 /// The [`ForkchoiceTask`] executes an `engine_forkchoiceUpdated` call with the current
 /// [`EngineState`]'s forkchoice, and no payload attributes.
 #[derive(Debug, Clone)]
 pub struct ForkchoiceTask {
     /// The engine client.
     pub client: Arc<EngineClient>,
+    /// The rollup config.
+    pub rollup: Arc<RollupConfig>,
+    /// Optional payload attributes to be used for the forkchoice update.
+    envelope: Option<OpAttributesWithParent>,
```
clabby, line 25: Any reason this is the only private field?

clabby (approved):

PR #2392, opened 2025-07-10
feat(node/p2p): Improved validation so invalid blocks are not marked seen

### LINKED ISSUE #2361: fix(node/p2p): Invalid Blocks Marked as Seen
### Description

In the `kona-node`, L2 blocks are received via libp2p gossip in the form of the `OpNetworkPayloadEnvelope`.

Once the raw libp2p message is decompressed and decoded into the `OpNetworkPayloadEnvelope` type, it is validated following the OP Stack block validation specs.

Currently, Kona has a silent inconsistency where it will [mark a block as seen](https://github.com/op-rs/kona/blob/main/crates/node/p2p/src/gossip/block_validity.rs#L166-L169) even if its signature is invalid.

For reference, the `op-node` handles block validation [here](https://github.com/ethereum-optimism/optimism/blob/develop/op-node/p2p/gossip.go#L404-L418).

### Task

Only mark the block as seen _if_ its signature is valid.

PoulavBhowmick03:
Fixes #2361

### DIFF
--- a/crates/node/p2p/src/gossip/block_validity.rs
+++ b/crates/node/p2p/src/gossip/block_validity.rs
@@ -143,6 +143,13 @@ impl BlockHandler {
         // CHECK: The payload is valid for the specific version of this block.
         Self::validate_version_specific_payload(envelope)?;
 
+        // Validate signature first
+        let msg = envelope.payload_hash.signature_message(self.rollup_config.l1_chain_id);
+        let _msg_signer = envelope
+            .signature
+            .recover_address_from_msg(&msg)
+            .map_err(|_| BlockInvalidError::Signature)?;
+
         if let Some(seen_hashes_at_height) =
             self.seen_hashes.get_mut(&envelope.payload.block_number())
         {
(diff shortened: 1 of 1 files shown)


crates/node/p2p/src/gossip/block_validity.rs
```diff
@@ -143,6 +143,13 @@ impl BlockHandler {
         // CHECK: The payload is valid for the specific version of this block.
         Self::validate_version_specific_payload(envelope)?;
 
+        // Validate signature first
+        let msg = envelope.payload_hash.signature_message(self.rollup_config.l1_chain_id);
+        let _msg_signer = envelope
+            .signature
+            .recover_address_from_msg(&msg)
+            .map_err(|_| BlockInvalidError::Signature)?;
+
```
refcell, line 152: We actually don't want this - we just want to move lines 179 - 191 here, above the seen_hashes check. That way, a block isn't marked as seen even if it has an invalid signature.

PoulavBhowmick03: got it. making the changes.
also, i was wondering, if there were any discussion groups for kona currently? would love to join and be a part of it @refcell

PR #2394, opened 2025-07-11
feat(bin/node): L2 Chain ID Abstraction

### LINKED ISSUE #2316: feat(bin/node): L2 Chain ID Abstraction
### Description

Reth [provides a simply ux](https://reth.rs/run/opstack#running-op-reth) for node operators to specify the chain for which they want to run a node. Instead of passing in a fallible chain ID number, reth uses alloy's [`NamedChain`](https://github.com/alloy-rs/chains/blob/main/src/named.rs) type which handles parsing chain ids into a strongly-typed chain identifier.

The `kona-node` currently let's the user specify the L2 chain id using the `--l2-chain-id` cli flag, which only accepts the numerical chain id.

This ticket is to update the `kona-node` to use the `NamedChain` type from `alloy-chains`.

varun-doshi:
Fixes #2316


crates/protocol/genesis/Cargo.toml
```diff
@@ -22,6 +22,7 @@ alloy-consensus.workspace = true
 alloy-eips.workspace = true
 alloy-hardforks.workspace = true
 alloy-op-hardforks.workspace = true
+alloy-chains = { workspace = true, features = ["serde", "arbitrary"] }
```
clabby, line 25: Let's break this out into the `arbitrary` and `serde` features. Shouldn't be enabling them by default.

refcell (changes requested):
Few small changes. Otherwise, looks great

PR #2395, opened 2025-07-11
test (supervisor/rpc): `supervisor_finalizedL1` and `supervisor_superRootAtTimestamp`

### LINKED ISSUE #2383: test (supervisor/rpc): e2e for `supervisor_finalizedL1` and `supervisor_dependencySetV1`

itschaindev:
Closes #2383 , #2384 

- Skipped testing for `supervisor_syncStatus` because that is being tested checking the chain progress.
- Skipped `supervisor_dependencySetV1` because we need `QueryAPI` in Go implementation to implement the end-point so that we can test it using devstack.

#### Changes

- The serialize and deserialize implemented for `ChainID` is now generic for `u64` because I needed it at other places as well.
- Due to the way Go unmarshalls the response, I needed to add the response objects `SuperRootOutputRpc` and `ChainRootInfoRpc`, which are essentially the same as `SuperRootOutput` and `ChainRootInfo` used before.  The fields `timestamp`, `version` and `chainID` needed to be compatible with `hexutil.Uint64` and hexutil.Bytes`. Hence all the modifications.

Tested locally, all tests including e2e tests are passing as expected.

### DIFF
--- a/crates/supervisor/rpc/src/response.rs
+++ b/crates/supervisor/rpc/src/response.rs
@@ -92,7 +92,7 @@ impl From<SuperHead> for SupervisorChainSyncStatus {
 pub struct ChainRootInfoRpc {
     /// The chain ID.
     #[serde(rename = "chainID", with = "alloy_serde::quantity")]
-    pub chain_id: u64,
+    pub chain_id: ChainId,
     /// The canonical output root of the latest canonical block at a particular timestamp.
     pub canonical: B256,
     /// The pending output root.


crates/supervisor/rpc/src/response.rs
```diff
@@ -80,10 +81,78 @@ impl From<SuperHead> for SupervisorChainSyncStatus {
...
+    pub version: u8,
+    /// The chain root info for each chain in the dependency set.
+    /// It represents the state of the chain at or before the timestamp.
+    pub chains: Vec<ChainRootInfoRpc>,
+}
+
+/// Serializes a [u8] as a hex string. Ensure that the hex string has an even length.
+///
+/// This is used to serialize the [`SuperRootOutputRpc`]'s version field as a hex string.
+fn serialize_u8_as_hex<S>(value: &u8, serializer: S) -> Result<S::Ok, S::Error>
```
sadiq1971, line 130: check whether the serde has some implemented features already.

itschaindev: I tried looking for serde features but couldn't find any for the required purpose. There is a function available though [serialize_hex_string_no_prefix](https://docs.rs/alloy-serde/latest/alloy_serde/fn.serialize_hex_string_no_prefix.html) that does what we want. Even with that we would need to add the 0x manually. I simplified the serialization of the u8 byte with my new commit. Let me know what you think.

crates/supervisor/rpc/src/response.rs
```diff
@@ -80,10 +81,66 @@ impl From<SuperHead> for SupervisorChainSyncStatus {
...
+///
+/// Required by
+/// [`super_root_at_timestamp`](crate::jsonrpsee::SupervisorApiServer::super_root_at_timestamp) RPC
+/// for marshalling and unmarshalling in GO implementation. Required for e2e tests.
+#[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ChainRootInfoRpc {
+    /// The chain ID.
+    #[serde(rename = "chainID", with = "alloy_serde::quantity")]
+    pub chain_id: u64,
```
sadiq1971, line 95: We can use the `ChainId` type for u64

sadiq1971 (approved):

PR #2404, opened 2025-07-14
feat(node/service): `L1OriginSelector` advancement

### LINKED ISSUE #2245: feat(node/service): L1 origin selector
## Overview

When the sequencer is extending the chain, it must rely on a service that selects the L1 origin for the next block, based off of the timestamp of the next L2 block.

Origin selection follows the [derivation spec](https://specs.optimism.io/protocol/derivation.html), advancing via sequencing epochs.

[Reference implementation](https://github.com/ethereum-optimism/optimism/blob/49a0b76c989fe38f619c063e487d40dfb8e8b3f8/op-node/rollup/sequencing/origin_selector.go#L4)

clabby:
## Overview

Adds in the logic for the `L1OriginSelector` to optimistically advance the origin based on the latest L1 block number, while allowing for transient failures to be retried.

Also adds testing for the selector logic.

closes #2245


crates/node/service/src/actors/sequencer/actor.rs
```diff
@@ -161,7 +161,17 @@ impl<AB: AttributesBuilder> SequencerActorState<AB> {
...
+            Ok(l1_origin) => l1_origin,
+            Err(err) => {
+                warn!(
+                    target: "sequencer",
+                    ?err,
+                    "Temporary error occurred while selecting next L1 origin. Re-attempting on next tick."
+                );
+                return Ok(())
+            }
+        };
```
clabby, line 174: Note that we don't really make a difference between what errors returned by the L1 origin selector are temporary vs. which aren't. The sequencer should always assume that transient RPC errors will recover.

crates/node/service/src/actors/sequencer/origin_selector.rs
```diff
@@ -96,16 +81,17 @@ impl L1OriginSelector {
...
         {
-            // If the next L2 block's timestamp is less than the next L1 origin's timestamp,
-            // return the current L1 origin.
+            // If the next L1 origin is ahead of the next L2 block's timestamp, return the current
+            // origin.
             return Ok(current);
         }
 
-        next.ok_or(L1OriginSelectorError::BlockNotFound(next_block_number.into()))
+        self.next.ok_or(L1OriginSelectorError::NotEnoughData)
```
theochap, line 94: nit: maybe we can add a little bit more information to that error, like what is the current next timestamp and what is the sum of `unsafe_head.block_info.timestamp + self.cfg.block_time`

crates/node/service/src/actors/sequencer/origin_selector.rs
```diff
@@ -142,7 +146,278 @@ pub enum L1OriginSelectorError {
...
+
+        // The expected L1 origin block is the one corresponding to the epoch of the current L2
+        // block. Assuming the next L1 origin block is not available from the eyes of the
+        // provider (_and_ it is not past the sequencer drift), the current L1 origin block
+        // will be re-used.
+        let expected_epoch =
+            if next_l1_origin_available { current_epoch + 1 } else { current_epoch };
+        assert_eq!(next.hash, B256::with_last_byte(expected_epoch as u8));
+        assert_eq!(next.number, expected_epoch);
+    }
```
theochap, line 344: These tests are _really_ great and I think we could get pretty high returns on having a few data structures like the `MockOriginSelectorProvider` in a common test crate to reuse to test the remaining actors

theochap (approved):
Looks great!

PR #2405, opened 2025-07-14
feat(node/sequencer): groundwork for op-conductor

theochap:
## Description

This PR lays the groundwork of the implementation of the `op-conductor`. It introduces skeleton code to support most of the `admin` api methods.

Meaningful changes:
- Creates admin api RPC server/client. Hook it to the `RPCActor`, the `SequencerActor` and the `NodeActor`
- Rename `NetworkRpc` to `P2pRpc`
- Make it possible to start/stop the sequencer inside the `SequencerActor`. Does a small refactor of the actor's internal logic.

## Follow-ups

- Plug the op-conductor into the CLI interface
- Tests


crates/node/service/src/actors/sequencer/actor.rs
```diff
@@ -54,12 +70,13 @@ impl From<SequencerBuilder>
...
         let cfg = seq_builder.cfg.clone();
         let l1_provider = seq_builder.l1_provider.clone();
+        let conductor = seq_builder.conductor.clone();
 
         let builder = seq_builder.build();
 
         let origin_selector = L1OriginSelector::new(cfg.clone(), l1_provider);
 
-        Self { cfg, builder, origin_selector }
+        Self { cfg, builder, origin_selector, conductor, is_active: true, is_recovery_mode: false }
```
clabby, line 79: I don't think we should have the sequencer be active at startup, yeah? Shouldn't it be disabled in order to sync, and then the conductor activates it?

clabby (approved):
Overall great start. Couple follow-ups, but nothing blocking this getting in other than the sequencer always being enabled at startup.

crates/node/service/src/actors/sequencer/actor.rs
theochap: I think that by default the sequencer should be enabled in sequencer mode. Once we have hooked up the conductor to the CLI we should ensure that when the `sequencer.stopped = true` argument is set, we deactivate the sequencer. Cf optimism-package https://github.com/theochap/optimism-package/blob/13703bb2e8e9e6f8ee04786bb8e6358f6a3827f5/src/cl/op-node/launcher.star#L268-L277

Issue #2387, opened 2025-07-09
fix(node/engine): audit for clones inside the `InsertUnsafeTask`.


theochap:
## Description

Inside the `InsertUnsafeTask` we're currently cloning the unsafe `envelope.payload` quite a few times - which means copying the whole block (transactions included).

See here: https://github.com/op-rs/kona/blob/42839d15ccff882580472900e808869ccee87114/crates/node/engine/src/task_queue/tasks/insert/task.rs#L80-L124

We should instead make sure to reduce the number of clones we're doing there by owning the data structures instead. This should improve performances since the `InsertUnsafeTask` should be executed quite frequently (whenever an unsafe block is gossiped, so potentially every second or so)

theochap:
I am going to take this @cy00r since I am working on an engine task refactor atm

PR #2412, opened 2025-07-15
feat(node/engine): remove temporary error variant. avoid cloning payloads when possible

### LINKED ISSUE #2403: feat(node/engine): remove temporary error type. Handle temporary errors inside the `execute` tasks themselves
## Description

This is a design proposal document to improve the safety/performance related to engine task handling.

We are currently returning 4 different error types when an engine task errors: `Critical`, `Reset`, `Flush` and `Temporary`. The `Temporary` error indicates that a given engine task should be retried indefinitely. This causes a few issues:

- A temporary error may block the engine forever (or for a prohibitively long time) if the temporary error never resolves. 
- The way we retry temporary errors should actually depends on various factors - we may want drop the error after a few attempts (simple FCU sync), or we may want to raise a critical errors (insert block from the sequencer actor), or raise a reset (or drop?) (received an unsafe block from another sequencer).
- Whenever we need to send some state information to the EL, we'll need to clone that structure and send it to the execution layer. This causes that any insert unsafe or build task will require copying _whole payload_ which may be _very_ expensive (copying a whole block/a few blocks).

A different solution would be to:
- Replace the `Temporary` error by a `Drop` error. When `Drop` is raised the task is dropped and not retried
- The tasks should retry operations themselves _inside the `execute` method_ when needed. It hides the complex retry logic from the caller and simplifies the interface (the caller only need to care about the final result, not the intermediary state).

## Misc

Closing this issue should fix #2387

theochap:
## Description

This PR replaces the `Temporary` error variant with a `Drop` variant inside the `Engine`. It unifies the way we propagate errors inside the engine, and in particular it postpones the error handling until they get propagated to the actor level. This makes the data flows in the engine more explicit

Close #2403 #2387

### DIFF
--- a/crates/node/engine/src/task_queue/tasks/build/error.rs
+++ b/crates/node/engine/src/task_queue/tasks/build/error.rs
@@ -14,15 +14,9 @@ use tokio::sync::mpsc;
 /// An error that occurs when running the [crate::ForkchoiceTask].
 #[derive(Debug, Error)]
 pub enum BuildTaskError {
-    /// The forkchoice update is not needed.
-    #[error("No forkchoice update needed")]
-    NoForkchoiceUpdateNeeded,
-    /// The engine is syncing.
-    #[error("Attempting to update forkchoice state while EL syncing")]
-    EngineSyncing,
-    /// Missing payload ID.
-    #[error("Missing payload ID")]
-    MissingPayloadId,
+    /// Failed to insert the payload into the engine.
+    #[error("Failed to insert the payload into the engine")]
+    InsertPayloadFailed,
     /// The initial forkchoice update call to the engine api failed.
     #[error(transparent)]
     ForkchoiceUpdateFailed(#[from] ForkchoiceTaskError),
@@ -64,12 +58,10 @@ impl EngineTaskError for BuildTaskError {
         match self {
             Self::ForkchoiceUpdateFailed(inner) => inner.severity(),
             Self::PayloadInsertionFailed(inner) => inner.severity(),
-            Self::NoForkchoiceUpdateNeeded => EngineTaskErrorSeverity::Temporary,
-            Self::EngineSyncing => EngineTaskErrorSeverity::Temporary,
-            Self::GetPayloadFailed(_) => EngineTaskErrorSeverity::Temporary,
-            Self::NewPayloadFailed(_) => EngineTaskErrorSeverity::Temporary,
+            Self::GetPayloadFailed(_) => EngineTaskErrorSeverity::Reset,
+            Self::NewPayloadFailed(_) => EngineTaskErrorSeverity::Reset,
             Self::HoloceneInvalidFlush => EngineTaskErrorSeverity::Flush,
-            Self::MissingPayloadId => EngineTaskErrorSeverity::Critical,
+            Self::InsertPayloadFailed => EngineTaskErrorSeverity::Critical,
             Self::UnexpectedPayloadStatus(_) => EngineTaskErrorSeverity::Critical,
             Self::DepositOnlyPayloadReattemptFailed => EngineTaskErrorSeverity::Critical,
             Self::DepositOnlyPayloadFailed => EngineTaskErrorSeverity::Critical,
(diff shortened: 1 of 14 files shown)


crates/node/engine/src/task_queue/tasks/build/error.rs
```diff
@@ -14,15 +14,9 @@ use tokio::sync::mpsc;
 /// An error that occurs when running the [crate::ForkchoiceTask].
 #[derive(Debug, Error)]
 pub enum BuildTaskError {
-    /// The forkchoice update is not needed.
-    #[error("No forkchoice update needed")]
-    NoForkchoiceUpdateNeeded,
-    /// The engine is syncing.
-    #[error("Attempting to update forkchoice state while EL syncing")]
-    EngineSyncing,
```
theochap, line 22: These two variants were not used

crates/node/engine/src/task_queue/tasks/build/error.rs
```diff
@@ -14,15 +14,9 @@ use tokio::sync::mpsc;
...
-    /// Missing payload ID.
-    #[error("Missing payload ID")]
-    MissingPayloadId,
+    /// Failed to insert the payload into the engine.
+    #[error("Failed to insert the payload into the engine")]
+    InsertPayloadFailed,
```
theochap, line 19: I renamed the `MissingPayloadId` to `InsertPayloadFailed` to also encapsulate the case where we exhausted the number of retries to insert the payload into the engine

crates/node/engine/src/task_queue/tasks/build/error.rs
refcell: Nice, love deleting code

crates/node/engine/src/task_queue/tasks/build/task.rs
```diff
@@ -54,14 +54,11 @@ impl BuildTask {
     /// - `engine_getPayloadV3` is used for payloads with a timestamp after the Ecotone fork.
     /// - `engine_getPayloadV4` is used for payloads with a timestamp after the Isthmus fork.
     async fn fetch_payload(
-        &self,
         cfg: &RollupConfig,
         engine: &EngineClient,
         payload_id: PayloadId,
-        payload_attrs: OpAttributesWithParent,
+        payload_timestamp: u64,
```
refcell, line 60: This is a nice change

crates/node/engine/src/task_queue/tasks/insert/task.rs
```diff
@@ -45,9 +47,65 @@ impl InsertTask {
...
+        payload: OpExecutionPayload,
+        parent_beacon_block_root: FixedBytes<32>,
+    ) -> Result<OpBlock, InsertTaskError> {
+        match payload {
+            payload @ OpExecutionPayload::V1(_) => {
+                payload.try_into_block().map_err(InsertTaskError::FromBlockError)
+            }
+            payload @ OpExecutionPayload::V2(_) => {
+                payload.try_into_block().map_err(InsertTaskError::FromBlockError)
+            }
```
refcell, line 64: Are we able to merge these two branches like so?

```suggestion
            payload @ OpExecutionPayload::V1(_) | payload @ OpExecutionPayload::V2(_) => {
                payload.try_into_block().map_err(InsertTaskError::FromBlockError)
            }
```

crates/node/service/src/actors/engine/actor.rs
```diff
@@ -268,43 +268,40 @@ impl EngineActorState {
...
                     }
                 }
+                EngineTaskErrorSeverity::Drop => {
+                    debug!(target: "engine", ?err, "Temporary error occurred while draining engine tasks. Continuing...");
+                }
             }
         }
 
+        trace!(target: "engine", "Engine tasks drained successfully");
+
```
refcell, line 304: nit: rm

```suggestion
```

theochap:
> I'm not sure if I entirely agree with removing the temporary error variant, but won't block if y'all feel this is the right move.
> 
> For me it's a bit harder to track each individual engine task's handling of temporary errors; Having the high-level retry in the task queue was really nice for this. Now, we have to handle those retries inline for every place that a temporary error can occur.

Raising good points here, closing this PR because it is too contentious of a change

PR #2415, opened 2025-07-15
feat(node/service): Reset engine prior to block building

### LINKED ISSUE #2413: feat(node/service): Initial engine reset in sequencer mode
## Overview

Prior to sequencing any blocks, the engine must be reset in order to initialize the CL's view of the current unsafe chain. Without this initial reset, the sequencer does not know what chain to build blocks on top of.

To do this, the `SequencerActor` needs to request and wait for the completion of an engine reset before entering the actor loop via `reset_request_tx`.

clabby:
## Overview

Sends an initial engine reset request and waits for it to conclude at the startup of the `SequencerActor`. This initializes the `EngineState`, and gives the sequencer a chain to begin building blocks on top of.

closes #2413

### DIFF
--- a/crates/node/engine/src/task_queue/core.rs
+++ b/crates/node/engine/src/task_queue/core.rs
@@ -2,8 +2,8 @@
 
 use super::EngineTaskExt;
 use crate::{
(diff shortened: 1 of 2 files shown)


theochap (approved):
Looks good!

PR #2416, opened 2025-07-16
fix(protocol): Serialization compatibility for `RollupConfig`

clabby:
## Overview

Fixes serialization compatibility with the Optimism monorepo's `RollupConfig` struct.

The base fee config deserialized correctly because kona-serde can handle both raw numbers or hex strings, but during serialization, it always renders a hex string: https://github.com/op-rs/kona/blob/main/crates/utilities/serde/src/quantity.rs#L26-L31

The system config fields previously deserialized correctly because a hex string of 32 bytes is a valid`U256`, but when we serialize, the `U256` ends up truncated, and anything != 32 bytes isn't a valid `B256`. https://github.com/ethereum-optimism/optimism/blob/63c6833c9bb8a707484732327302ee901d7426f9/op-service/eth/types.go#L505-L511


theochap (approved):
Thanks for taking care of it

PR #2421, opened 2025-07-16
fix(node/engine): Consolidate task transient safe chain updates

clabby:
## Overview

Fixes the transient safe chain updates introduced in https://github.com/op-rs/kona/pull/1937. These were lost in a recent refactor to the engine tasks.

### DIFF
--- a/crates/node/engine/src/task_queue/tasks/consolidate/task.rs
+++ b/crates/node/engine/src/task_queue/tasks/consolidate/task.rs
@@ -91,12 +91,21 @@ impl ConsolidateTask {
                 Ok(block_info) if !self.attributes.is_last_in_span => {
                     let total_duration = global_start.elapsed();
 
-                    info!(target: "engine",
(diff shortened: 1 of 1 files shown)


crates/node/engine/src/task_queue/tasks/consolidate/task.rs
```diff
@@ -91,12 +91,21 @@ impl ConsolidateTask {
...
+                        ?block_fetch_duration,
+                        "Skipping FCU for non-last in span block"
+                    );
+
+                    // Apply a transient update to the safe head.
+                    state.sync_state.apply_update(EngineSyncStateUpdate {
+                        safe_head: Some(block_info),
+                        local_safe_head: Some(block_info),
+                        ..Default::default()
+                    });
```
theochap, line 108: I think this will not update the `sync_state` because `.apply_update` just consumes self and returns a new version. It doesn't mutate the underlying struct

PR #2425, opened 2025-07-17
feat(node/engine): fix unsafe payload signature. add large kona sequencer config

theochap:
## Description

Fix incorrect unsafe payload signature. Add a new large kona sequencer kurtosis configuration. Closes the sequencer MVP


clabby (approved):
Looks great after `op-alloy` update, needs lint.

PR #2444, opened 2025-07-18
feat(supervisor/storage): implemented rewinder for log storage

### LINKED ISSUE #1837: feat(supervisor/storage): log provider unwinding
# Description

Support unwinding for the log provider to unwind the logs and block.

sadiq1971:
Closes #1837


itschaindev (approved):
lgtm. i'll create another ticket in re-org tracker to trigger this so that we don't miss it.

crates/supervisor/storage/src/providers/log_provider.rs
```diff
@@ -124,6 +124,19 @@ where
         }
         Ok(())
     }
+
+    pub(crate) fn rewind_from(&self, block_number: u64) -> Result<(), StorageError> {
```
dhyaniarun1993, line 128: I guess `rewind_to` name makes more sense considering we are dealing with the append only logging.

dhyaniarun1993 (commented):
Also, I think it would be better to perform rewinding based blockID instead of just block number.

sadiq1971:
> Also, I think it would be better to perform rewinding based blockID instead of just block number.

I don't think we will need hash matching required here in any case

sadiq1971:
> > Also, I think it would be better to perform rewinding based blockID instead of just block number.
> 
> I don't think we will need hash matching required here in any case

But let's do step by step. Will do the change whenever we need it

PR #2460, opened 2025-07-19
feat(node/service): support sequencer recovery mode

### LINKED ISSUE #2446: feat(node/service): Support sequencer recovery mode
## Overview

Sequencer recovery mode is currently configurable, but doesn't actually do anything in the `L1OriginSelector`. We'll need to add support for this.

varun-doshi:
Fixes #2446

### DIFF
--- a/crates/node/service/src/actors/sequencer/origin_selector.rs
+++ b/crates/node/service/src/actors/sequencer/origin_selector.rs
@@ -48,8 +48,9 @@ impl<P: L1OriginSelectorProvider> L1OriginSelector<P> {
     pub async fn next_l1_origin(
         &mut self,
         unsafe_head: L2BlockInfo,
+        is_recovery_mode: bool,
     ) -> Result<BlockInfo, L1OriginSelectorError> {
-        self.select_origins(&unsafe_head).await?;
+        self.select_origins(&unsafe_head, is_recovery_mode).await?;
 
         // Start building on the next L1 origin block if the next L2 block's timestamp is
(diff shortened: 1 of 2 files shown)


varun-doshi:
@clabby can you please confirm if this is in the right direction?

crates/node/service/src/actors/sequencer/origin_selector.rs
```diff
@@ -98,7 +99,19 @@ impl<P: L1OriginSelectorProvider> L1OriginSelector<P> {
...
+        if in_recovery_mode {
+            let current_origin = self.l1.get_block_by_hash(unsafe_head.l1_origin.hash).await?;
+            if let Some(current_origin) = current_origin {
+                self.current = Some(current_origin);
+                let next_origin = self.l1.get_block_by_number(current_origin.number + 1).await?;
+                if let Some(next_origin) = next_origin {
+                    self.next = Some(next_origin);
+                }
+            }
+        }
```
clabby, line 113:
```suggestion
        if in_recovery_mode {
            self.current = self.l1.get_block_by_hash(unsafe_head.l1_origin.hash).await?;
            self.next = self.l1.get_block_by_number(current_origin.number + 1).await?;
            return Ok(());
        }
```

clabby (commented):
This is looking good.

We also need to prevent the inclusion of user-space transactions if recovery mode is active. Over in the sequencer actor's `build_block` function, can you add a branch that sets `no_tx_pool = true` if recovery mode is active?

clabby (approved):
Thanks!

PR #2482, opened 2025-07-21
feat(meta/logs): allow logging to file, silencing stdout and more configuration

theochap:
## Description

Changes to the logging infra for kona binaries:

- Unify all the logging CLI interface behind `LogArgs`
- Add file based logging with rotation
- Make logging output to stdout optional with the quiet flag

### DIFF
--- a/bin/host/src/bin/host.rs
+++ b/bin/host/src/bin/host.rs
@@ -25,7 +25,7 @@ primary thread.
 pub struct HostCli {
     /// Logging arguments.
     #[command(flatten)]
-    pub v: LogArgs,
+    pub log_args: LogArgs,
     /// Host mode
     #[command(subcommand)]
     pub mode: HostMode,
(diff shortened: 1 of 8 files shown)


bin/host/src/bin/host.rs
```diff
@@ -23,14 +23,9 @@ primary thread.
...
-    /// If set to 0, no logs are printed.
-    /// By default, the verbosity level is set to 3 (info level).
-    #[arg(long, short, default_value = "3", action = ArgAction::Count)]
-    pub v: u8,
-    /// The format of the logs. One of: full, json, pretty, compact.
-    #[arg(long = "logs.format", short = 'f', default_value = "full")]
-    pub logs_format: LogFormat,
+    /// Logging arguments.
+    #[command(flatten)]
+    pub v: LogArgs,
```
clabby, line 28: slight nit - since `v` is no longer just the verbosity command (and instead the flattened `LogArgs`,) we can change this variable name to something more descriptive.

crates/utilities/cli/src/log.rs
```diff
@@ -1,40 +1,157 @@
...
-    /// Verbosity level (0-5).
-    /// If set to 0, no logs are printed.
+    /// Verbosity level (1-5).
     /// By default, the verbosity level is set to 3 (info level).
+    ///
+    /// This verbosity level is shared by both stdout and file logging (if enabled).
     #[arg(
-        long,
-        short,
+        long = "logs.level",
```
clabby, line 19: What's the purpose of `logs.level`? It's an `ArgAction::Count` var; Think we should likely keep it as just `v`. With the long variant, you can't do `--logs.level=3` because of the `action`, it'd be `--logs.levellogs.levellogs.level`.

clabby (approved):
I like this change overall. Couple small nits.

PR #2486, opened 2025-07-21
feat(node/service): Commit unsafe payloads to `op-conductor`

clabby:
## Overview

Adds functionality for the `kona-node` to commit unsafe payloads that the sequencer builds to the conductor, if present.


crates/node/service/src/actors/sequencer/actor.rs
```diff
@@ -314,6 +314,14 @@ impl<AB: AttributesBuilder> SequencerActorState<AB> {
         }
 
         let payload = self.try_wait_for_payload(ctx, payload_rx).await?;
+
+        // If the conductor is available, commit the payload to it.
+        if let Some(conductor) = &self.conductor {
+            if let Err(err) = conductor.commit_unsafe_payload(&payload).await {
+                error!(target: "sequencer", ?err, "Failed to commit unsafe payload to conductor");
```
theochap, line 321: We should probably explicitly return an error there - unless we think we can recover from that situation (which would be unlikely because it seems that not having this method crashes the conductor when updating the leader)

clabby: We should; The `op-node` treats this as a temporary error.

theochap (approved):

clabby:
Needs release of `op-alloy` w/ https://github.com/alloy-rs/op-alloy/pull/578

PR #2498, opened 2025-07-22
feat(supervisor/core): Handle unsafe reorg

### LINKED ISSUE #2048: feat(supervisor/core): Handle unwinding on hash mismatch on Safe block event
# Description

When we get the safe block event, there is a chance that it mismatches with the unsafe block seen earlier. In case of mismatch, we need to unwind the logs and block. Full flow handling needs to be discussed and developed.

Not sure when this will happen. If we are handling L2 Unsafe reorg, this should not happen. But in case this happens, we can use the similar logic that we did in `resetter/latest_valid_local_safe`. This logic should also be pulled into `reorg` module.

sadiq1971:
Closes #2048


itschaindev (approved):
lgtm

crates/supervisor/core/src/rewinder/chain.rs
```diff
@@ -0,0 +1,94 @@
...
+            self.db.get_block(derived_pair.derived.number).inspect_err(|err| {
+                error!(
+                    target: "rewinder",
+                    chain = %self.chain_id,
+                    block_number = derived_pair.derived.number,
+                    %err,
+                    "Error retrieving conflicting block for reorg"
+                );
+            })?;
+
```
dhyaniarun1993, line 49: I think we should cross check if the block is conflicting

crates/supervisor/storage/src/chaindb.rs
```diff
@@ -114,6 +114,12 @@ impl DerivationStorageWriter for ChainDb {
     fn save_derived_block(&self, incoming_pair: DerivedRefPair) -> Result<(), StorageError> {
         self.observe_call("save_derived_block", || {
             self.env.update(|ctx| {
+                DerivationProvider::new(ctx).save_derived_block(incoming_pair)?;
```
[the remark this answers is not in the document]
sadiq1971: I have removed the storage changes.

crates/supervisor/core/src/rewinder/chain.rs
sadiq1971: added the check.

dhyaniarun1993 (approved):

PR #2509, opened 2025-07-22
test(node/rpc): test rollup config rpc endpoint. fix rollup config metrics

theochap:
## Description

This PR:
- Tests `RollupConfig` rpc endpoint
- Fixes the rollup config metrics in grafana if the chain id is not part of the scr
- Adds a test to ensure the other nodes of the network are lagging behind the sequencer unsafe head

### DIFF
--- a/bin/node/src/commands/net.rs
+++ b/bin/node/src/commands/net.rs
@@ -9,6 +9,7 @@ use kona_node_service::{
     NetworkActor, NetworkBuilder, NetworkContext, NetworkInboundData, NodeActor,
 };
 use kona_p2p::P2pRpcRequest;
+use kona_registry::scr_rollup_config_by_alloy_ident;
 use kona_rpc::{OpP2PApiServer, P2pRpc, RpcBuilder};
 use tokio_util::sync::CancellationToken;
 use tracing::{info, warn};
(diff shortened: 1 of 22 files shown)


tests/node/mixed_preset.go
```diff
@@ -134,7 +133,6 @@ func NewMixedOpKona(t devtest.T) *MixedOpKonaPreset {
 		L2CLOpNodes:   L2CLNodes(opCLNodes, orch),
 		L2ELKonaNodes: L2ELNodes(konaELNodes),
 		L2CLKonaNodes: L2CLNodes(konaCLNodes, orch),
-		TestSequencer: dsl.NewTestSequencer(system.TestSequencer(match.Assume(t, match.FirstTestSequencer))),
```
theochap, line 137: It seems the op-devstack doesn't support detecting the sequencer nodes from inside kurtosis. I handrolled a matcher for that

crates/protocol/registry/src/lib.rs
```diff
@@ -36,16 +36,21 @@ lazy_static::lazy_static! {
...
-pub fn rollup_config_by_alloy_ident(chain: &alloy_chains::Chain) -> Option<&RollupConfig> {
+pub fn scr_rollup_config_by_alloy_ident(chain: &alloy_chains::Chain) -> Option<&RollupConfig> {
     ROLLUP_CONFIGS.get(&chain.id())
 }
 
+/// Returns a [RollupConfig] by its chain id.
+pub fn scr_rollup_config_by_chain_id(chain_id: &u64) -> Option<&RollupConfig> {
+    ROLLUP_CONFIGS.get(chain_id)
+}
+
```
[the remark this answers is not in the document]
theochap: Done

bin/node/src/flags/globals.rs
```diff
@@ -26,57 +25,12 @@ pub struct GlobalArgs {
     /// Embed the override flags globally to provide override values adjacent to the configs.
```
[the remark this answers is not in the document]
theochap: Done

bin/node/src/commands/net.rs
```diff
@@ -59,14 +60,15 @@ impl NetCommand {
 
         let rpc_config = Option::<RpcBuilder>::from(self.rpc);
 
+        let alloy_chain: alloy_chains::Chain = args.l2_chain_id.into();
```
refcell, line 63: That's fine - i was hoping the l2 chain id would be updated to `Chain` itself but can be a follow on pr

bin/node/src/commands/node.rs
```diff
@@ -200,6 +201,12 @@ impl NodeCommand {
     /// Run the Node subcommand.
     pub async fn run(self, args: &GlobalArgs) -> anyhow::Result<()> {
         let cfg = self.get_l2_config(args)?;
+
+        // If metrics are enabled, initialize the global cli metrics.
+        if args.metrics.enabled {
+            init_rollup_config_metrics(&cfg);
+        }
```
refcell, line 208: Think this could be done inline since there's no return - not a big deal tho

```rust
        args.metrics.enabled.then(|| init_rollup_config_metrics(&cfg));
```

PR #2519, opened 2025-07-22
feat(node/engine): spike a dev rpc api to get inner engine state.

theochap:
## Description

Spiked a dev rpc api to allow to have access to the inner engine state (task queue length) to help with kurtosis testing.

### DIFF
--- a/crates/node/rpc/src/jsonrpsee.rs
+++ b/crates/node/rpc/src/jsonrpsee.rs
@@ -10,6 +10,7 @@ use jsonrpsee::{
     proc_macros::rpc,
 };
 use kona_genesis::RollupConfig;
+use kona_interop::{ExecutingDescriptor, SafetyLevel};
 use kona_p2p::{PeerCount, PeerDump, PeerInfo, PeerStats};
 use kona_protocol::SyncStatus;
 use op_alloy_rpc_types_engine::OpExecutionPayloadEnvelope;


clabby (commented):
Is it possible to do this with metrics, rather than the invasive change? I was thinking we should be able to just fire off some metrics for engine queue size and observe them in the test.

theochap:
> Is it possible to do this with metrics, rather than the invasive change + new RPC API? I was thinking we should be able to just fire off some metrics for engine queue size and observe them in the test.

How would you observe the metrics from inside kurtosis? I haven't tried but that may be possible to do.
We just have to make sure that the metrics are reliably fired to avoid flakiness. Just to keep in mind if we ever decide to do metrics optimization in the future - we might encounter issues with our test code and that's a bit of a footgun

theochap:
> > Is it possible to do this with metrics, rather than the invasive change + new RPC API? I was thinking we should be able to just fire off some metrics for engine queue size and observe them in the test.
> 
> How would you observe the metrics from inside kurtosis? I haven't tried but that may be possible to do. We just have to make sure that the metrics are reliably fired to avoid flakiness. Just to keep in mind if we ever decide to do metrics optimization in the future - we might encounter issues with our test code and that's a bit of a footgun

I also like to have a dev RPC api because that makes it easier for client devs to test it with the devstack and such without having to use the metrics hack.
Won't die on that hill though

refcell (approved):

PR #2524, opened 2025-07-22
fix(node/engine): audit engine for bugs. rename forkchoice task -> synchronize task. move block building logic to build task

theochap:
## Description

Attempt at cleaning-up the engine and restoring the behavior from pre-refactor https://github.com/op-rs/kona/pull/2388/files.
In particular:
- Split up the forkchoice task into separated build and synchronize logic. The synchronize logic is now closer to the pre #2388 state
- Restore the build FCU method in the build task. Ensure that the side-effects match the pre-#2388 state
- Audit the other tasks and ensure the composition (if any), follow the side-effects from #2388
- Remove the `Forkchoice` task from the `EngineTask`s.


crates/node/engine/src/task_queue/core.rs
```diff
@@ -76,7 +77,8 @@ impl Engine {
         let start =
             find_starting_forkchoice(&config, client.l1_provider(), client.l2_provider()).await?;
 
-        if let Err(err) = ForkchoiceTask::new(
+        // Retry to synchronize the engine until we succeeds or a critical error occurs.
+        while let Err(err) = SynchronizeTask::new(
```
theochap, line 81: Motivation, the synchronize task can only fail if:
- RPC error (temporary error) -> we should retry to reset
- finalized ahead of unsafe (critical error) -> should never happen if `find_starting_forkchoice` is correct
- unexpected payload status (temporary error) -> this **should** never happen when we don't provide payload attributes. The synchronize task should **only** return `Valid` or `EngineSyncing` status
- invalid forkchoice state -> try to reset again

If we don't have a `while` loop here:
- we will keep adding tasks with the former engine state
- this will cause the engine to reset because the tasks will return `invalid forkchoice state`

crates/node/engine/src/task_queue/tasks/build/error.rs
```diff
@@ -11,33 +11,44 @@ use op_alloy_rpc_types_engine::OpExecutionPayloadEnvelope;
...
+    /// The inserted payload status is unexpected.
+    #[error("The inserted payload status is unexpected: {0}")]
+    UnexpectedPayloadStatus(PayloadStatusEnum),
+    /// The payload ID is missing.
+    #[error("The inserted payload ID is missing")]
+    MissingPayloadId,
     /// The engine is syncing.
-    #[error("Attempting to update forkchoice state while EL syncing")]
+    #[error("The engine is syncing")]
     EngineSyncing,
```
theochap, line 34: Those variants are the same as before #2388 https://github.com/op-rs/kona/pull/2388/files

crates/node/engine/src/task_queue/tasks/build/error.rs
```diff
@@ -64,13 +75,26 @@ impl EngineTaskError for BuildTaskError {
...
+            }
+            Self::EngineBuildError(EngineBuildError::UnexpectedPayloadStatus(_)) => {
+                EngineTaskErrorSeverity::Temporary
+            }
+            Self::EngineBuildError(EngineBuildError::MissingPayloadId) => {
+                EngineTaskErrorSeverity::Temporary
+            }
+            Self::EngineBuildError(EngineBuildError::EngineSyncing) => {
+                EngineTaskErrorSeverity::Temporary
+            }
```
theochap, line 95: https://github.com/op-rs/kona/pull/2388/files#diff-700a0568e8e46647bbf1fbb4830b66e5edc3820e0d322136cccc9e528f333d14L62-L81

theochap: Note here that:
- `ForkchoiceUpdateFailed` (pre #2388) -> `AttributesInsertionFailed`. Pre-#2388, the SECOND failing FCU caused a `ForkchoiceError` which _did not appear_ inside the `BuildError` variants. See https://github.com/op-rs/kona/pull/2388/files#diff-47bc5e44b8483f9c829c8829f56e6f5b6d9211888f53f7afe36b44c185515ccbL359.
- `UnexpectedPayloadStatus` split between `InvalidPayload` and `UnexpectedPayloadStatus`.

crates/node/engine/src/task_queue/tasks/build/error.rs
```diff
@@ -64,13 +75,26 @@ impl EngineTaskError for BuildTaskError {
...
+            Self::EngineBuildError(EngineBuildError::AttributesInsertionFailed(_)) => {
+                EngineTaskErrorSeverity::Temporary
+            }
+            Self::EngineBuildError(EngineBuildError::InvalidPayload(_)) => {
+                EngineTaskErrorSeverity::Temporary
+            }
+            Self::EngineBuildError(EngineBuildError::UnexpectedPayloadStatus(_)) => {
+                EngineTaskErrorSeverity::Temporary
+            }
```
theochap, line 89: To check: I think that this should instead be
```suggestion
Self::EngineBuildError(EngineBuildError::UnexpectedPayloadStatus(_)) => {
                EngineTaskErrorSeverity::Critical
            }
```
The `forkchoice_updated` should *never* return the Accepted variant

crates/node/engine/src/task_queue/tasks/build/error.rs
```diff
@@ -64,13 +75,26 @@ impl EngineTaskError for BuildTaskError {
...
-            Self::NoForkchoiceUpdateNeeded => EngineTaskErrorSeverity::Temporary,
-            Self::EngineSyncing => EngineTaskErrorSeverity::Temporary,
+            Self::EngineBuildError(EngineBuildError::FinalizedAheadOfUnsafe(_, _)) => {
+                EngineTaskErrorSeverity::Critical
+            }
+            Self::EngineBuildError(EngineBuildError::AttributesInsertionFailed(_)) => {
+                EngineTaskErrorSeverity::Temporary
+            }
+            Self::EngineBuildError(EngineBuildError::InvalidPayload(_)) => {
+                EngineTaskErrorSeverity::Temporary
```
theochap, line 85: This also should probably be
```suggestion
                EngineTaskErrorSeverity::Critical
```
- trying to insert an invalid payload _should_ never happen

crates/node/engine/src/task_queue/tasks/build/error.rs
```diff
@@ -64,13 +75,26 @@ impl EngineTaskError for BuildTaskError {
...
+            }
+            Self::EngineBuildError(EngineBuildError::EngineSyncing) => {
+                EngineTaskErrorSeverity::Temporary
+            }
             Self::GetPayloadFailed(_) => EngineTaskErrorSeverity::Temporary,
-            Self::NewPayloadFailed(_) => EngineTaskErrorSeverity::Temporary,
             Self::HoloceneInvalidFlush => EngineTaskErrorSeverity::Flush,
-            Self::MissingPayloadId => EngineTaskErrorSeverity::Critical,
-            Self::UnexpectedPayloadStatus(_) => EngineTaskErrorSeverity::Critical,
```
theochap, line 73: Some of these variants were either unused or outdated

crates/node/engine/src/task_queue/tasks/build/task.rs
```diff
@@ -44,6 +46,120 @@ impl BuildTask {
...
+                ));
+            }
+        }
+
+        // Fetch the payload ID from the FCU. If no payload ID was returned, something went wrong -
+        // the block building job on the EL should have been initiated.
+        update
+            .payload_id
+            .ok_or(BuildTaskError::EngineBuildError(EngineBuildError::MissingPayloadId))
+    }
```
theochap, line 161: This matches https://github.com/op-rs/kona/pull/2388/files#diff-47bc5e44b8483f9c829c8829f56e6f5b6d9211888f53f7afe36b44c185515ccbL49-L137

crates/node/engine/src/task_queue/tasks/build/task.rs
```diff
@@ -167,18 +273,17 @@ impl EngineTaskExt for BuildTask {
         .execute(state)
         .await
         {
-            Err(InsertTaskError::ForkchoiceUpdateFailed(
-                ForkchoiceTaskError::InvalidPayloadStatus(e),
-            )) if self.attributes.is_deposits_only() => {
+            Err(InsertTaskError::UnexpectedPayloadStatus(e))
+                if self.attributes.is_deposits_only() =>
+            {
```
theochap, line 278: This was a bug in the refactor. The error we should match on should be raised when we first try to call `new_payload` inside the `InsertTask`. See https://github.com/op-rs/kona/pull/2388/files#diff-47bc5e44b8483f9c829c8829f56e6f5b6d9211888f53f7afe36b44c185515ccbR169-R220 (for the former build code), and https://github.com/op-rs/kona/blob/572de05df03bf47dca903bb8d46261a11b3df56e/crates/node/engine/src/task_queue/tasks/insert/task.rs#L112-L120 (for the counterpart inside the `InsertTask`)

crates/node/engine/src/task_queue/tasks/insert/task.rs
```diff
@@ -134,7 +134,6 @@ impl EngineTaskExt for InsertTask {
                 safe_head: self.is_payload_safe.then_some(new_unsafe_ref),
                 ..Default::default()
             },
```
theochap, line 136: This is the same as https://github.com/op-rs/kona/pull/2388/files#diff-7f07dc34b612a89fb6775be8f4d4a4279d4d4eb60d9910ea7e5c5a30db3fd78dL140-L182

crates/node/engine/src/task_queue/tasks/synchronize/task.rs
```diff
@@ -0,0 +1,137 @@
...
+        let new_sync_state = state.sync_state.apply_update(self.state_update);
+
+        // Check if a forkchoice update is not needed, return early.
+        // A forkchoice update is not needed if...
+        // 1. The engine state is not default (initial forkchoice state has been emitted), and
+        // 2. The new sync state is the same as the current sync state (no changes to the sync
+        //    state).
+        if state.sync_state != Default::default() && state.sync_state == new_sync_state {
+            debug!(target: "engine", ?new_sync_state, "No forkchoice update needed");
+            return Ok(());
```
theochap, line 84: *IMPORTANT*: we _shouldn't_ retry the synchronize task there. Since the `sync_state` is only updated inside the `SynchronizeTask` (except inside the `ConsolidateTask`, when the block is not the last in the batch) - the engine will get stuck retrying the `SynchronizeTask`

crates/node/engine/src/task_queue/tasks/synchronize/task.rs
```diff
@@ -0,0 +1,137 @@
...
+            }
+            PayloadStatusEnum::Syncing => {
+                // If we're not building a new payload, we're driving EL sync.
+                warn!(target: "engine", "Attempting to update forkchoice state while EL syncing");
+                Ok(())
+            }
+            s => {
+                // Other codes are not expected.
+                Err(SynchronizeTaskError::UnexpectedPayloadStatus(s.clone()))
+            }
```
theochap, line 63: I _think_ in this case we should return a Critical error. The `forkchoice_updated` method shouldn't _ever_ return the other status enum if we're not providing payloads

crates/node/engine/src/task_queue/tasks/synchronize/error.rs
```diff
@@ -1,46 +1,34 @@
...
+impl EngineTaskError for SynchronizeTaskError {
     fn severity(&self) -> EngineTaskErrorSeverity {
         match self {
-            Self::NoForkchoiceUpdateNeeded => EngineTaskErrorSeverity::Temporary,
-            Self::EngineSyncing => EngineTaskErrorSeverity::Temporary,
-            Self::ForkchoiceUpdateFailed(_) => EngineTaskErrorSeverity::Temporary,
             Self::FinalizedAheadOfUnsafe(_, _) => EngineTaskErrorSeverity::Critical,
-            Self::UnexpectedPayloadStatus(_) => EngineTaskErrorSeverity::Critical,
+            Self::ForkchoiceUpdateFailed(_) => EngineTaskErrorSeverity::Temporary,
+            Self::UnexpectedPayloadStatus(_) => EngineTaskErrorSeverity::Temporary,
```
theochap, line 30: See below for the `UnexpectedPayloadStatus` that should be `Critical`

crates/node/engine/src/task_queue/tasks/synchronize/task.rs
```diff
@@ -0,0 +1,137 @@
...
+                        "Finished execution layer sync."
+                    );
+                    state.el_sync_finished = true;
+                }
+
+                Ok(())
+            }
+            PayloadStatusEnum::Syncing => {
+                // If we're not building a new payload, we're driving EL sync.
+                warn!(target: "engine", "Attempting to update forkchoice state while EL syncing");
```
refcell, line 57: This may be a loud log @ warn since this code path is expected to be hit frequently

crates/node/engine/src/task_queue/tasks/synchronize/task.rs
refcell: It should only critical if the error is actually invalid. Accepted is fine. We should be very explicit about what is a critical error

crates/node/engine/src/task_queue/tasks/synchronize/task.rs
```diff
@@ -0,0 +1,137 @@
...
+            s => {
+                // Other codes are not expected.
+                Err(SynchronizeTaskError::UnexpectedPayloadStatus(s.clone()))
+            }
+        }
+    }
+}
+
+#[async_trait]
+impl EngineTaskExt for SynchronizeTask {
```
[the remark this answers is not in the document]
theochap: I agree, I think we can work on that in a follow-up PR. I would do as you said and just add methods on a `Synchronize` struct and not implement the `EngineTaskExt` anymore

crates/node/engine/src/task_queue/tasks/synchronize/task.rs
theochap: Although I am also a bit unsure about returning a critical error here because we might also get affected if there is any change to the downstream api (if `alloy` ever decide to return `Accepted` inside `forkchoice_update`)

crates/node/engine/src/task_queue/tasks/synchronize/task.rs
theochap: Done

crates/node/engine/src/task_queue/tasks/synchronize/task.rs
theochap: Done

PR #2552, opened 2025-07-24
test(node/engine): add positive engine test

theochap:
## Description

Adds a small test that checks the engine never enqueues more than 1 task at a time under normal conditions


tests/node/rpc_test.go
```diff
@@ -127,10 +126,8 @@ func p2pBanPeer(t devtest.T, out *MixedOpKonaPreset) {
 		// We pick the first peer that is connected.
 		peerToBan := ""
 		for _, peer := range peers.Peers {
-			if peer.Connectedness == network.Connected {
-				peerToBan = peer.PeerID.String()
-				break
-			}
+			peerToBan = peer.PeerID.String()
+			break
```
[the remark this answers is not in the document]
theochap: Don't worry dear AI, us humans are confused by that too.

justfile
```diff
@@ -26,12 +26,12 @@ build-node:
 tests: test test-docs
 
 # Test for the native target with all features. By default, excludes online tests.
-test *args="-E '!test(test_online)'":
```
clabby, line 29: I'd prefer to keep the wildcard args; allows for passing a specific test, etc.

justfile
```diff
@@ -88,8 +88,8 @@ test-docs:
   cargo test --doc --workspace --locked
 
 # Build for the native target
-build-native *args='':
-  cargo build --workspace $@
+build-native binary_name='':
```
clabby, line 91: same - sometimes I like to pass a specific package with `-p` or something.

clabby (changes requested):
lgtm; prefer to revert the changes to the root justfile.

justfile
theochap: Actually those changes are not meant to be there

tests/devnets/large-kona-sequencer.yaml
```diff
@@ -21,6 +21,7 @@ optimism_package:
             image: "kona-node:local"
             extra_env_vars:
               KONA_NODE_RPC_WS_ENABLED: "true"
+              KONA_NODE_RPC_DEV_ENABLED: "true"
```
refcell, line 24: Do these have to be strings?

tests/node/rpc_test.go
```diff
@@ -127,10 +126,8 @@ func p2pBanPeer(t devtest.T, out *MixedOpKonaPreset) {
...
 		}
```
refcell, line 132: Why even loop here? just grab the first peer id string and just require that the list is not empty

refcell (approved):
Few comments but not strictly blocking

tests/devnets/large-kona-sequencer.yaml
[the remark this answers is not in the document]
theochap: If you pass the args to CLI you'd just add the arg. This is required for env vars

PR #2554, opened 2025-07-24
test(node): connection drop test

theochap:
## Description

Test what happens when the connection to the other nodes of the network drops (or gets cut off from blacklisting)


refcell (approved):
These are nice

theochap:
#2620

PR #2562, opened 2025-07-26
fix dead link

ganlerseian:
Hey team! Found and fixed dead link
https://github.com/ethereum-optimism/superchain-registry/blob/main/superchain/superchain.go#L80 - old link

https://github.com/ethereum-optimism/superchain-registry/blob/main/ops/internal/config/superchain.go#L70 - new link


emhane (approved):
thanks

PR #2566, opened 2025-07-28
chore(node/runtime): remove remaining runtime references in kona

theochap:
## Description

It seems the AI has missed some references in https://github.com/op-rs/kona/pull/2535. Removes the remaining runtime libraries in the node.


bin/node/src/flags/p2p.rs
```diff
@@ -318,19 +315,8 @@ impl P2PArgs {
...
-        l1_rpc: Option<Url>,
     ) -> anyhow::Result<alloy_primitives::Address> {
-        // First attempt to load the unsafe block signer from the runtime loader.
-        if let Some(url) = l1_rpc {
-            let mut loader = RuntimeLoader::new(url, Arc::new(config.clone()));
-            let runtime = loader.load_latest().await.map_err(|e| {
-                anyhow::anyhow!("Failed to load runtime: {} {:?}", e, std::error::Error::source(&e))
-            })?;
-            return Ok(runtime.unsafe_block_signer_address);
-        }
```
clabby, line 332: This we can't do unfortunately. We must load the current value at startup. Could be done with a much-simplified `RuntimeLoader` though.

PR #2572, opened 2025-07-28
feat(node/sequencer): implement remote signer skeleton

theochap:
## Description

This PR spikes the remote signer for kona node's sequencer. In particular it:
- Introduces a new `BlockSigner` enum which encapsulates the behavior of a `LocalSigner` and a `RemoteSigner`
- Propagates the changes of the `BlockSigner` enum along the stack. Still use the `LocalSigner` everywhere
- Adds the core of the remote signer code. In particular, it ensures we implement the `op-signer` interface, we handle CA and client certificates and support hot reloads for certificates through a file watcher (like the op-node does)

Follow-up: add CLI options to activate remote signer from the command line.

### DIFF
--- a/crates/node/service/src/actors/network/actor.rs
+++ b/crates/node/service/src/actors/network/actor.rs
@@ -2,6 +2,7 @@ use alloy_primitives::Address;
 use async_trait::async_trait;
 use kona_p2p::P2pRpcRequest;
 use kona_rpc::NetworkAdminQuery;
+use kona_sources::BlockSignerError;
 use libp2p::TransportError;
 use op_alloy_rpc_types_engine::{OpExecutionPayloadEnvelope, OpNetworkPayloadEnvelope};
 use thiserror::Error;
@@ -133,7 +134,7 @@ pub enum NetworkActorError {
     ChannelClosed,
     /// Failed to sign the payload.
     #[error("Failed to sign the payload: {0}")]
-    FailedToSignPayload(String),
+    FailedToSignPayload(#[from] BlockSignerError),
 }
 
 #[async_trait]
@@ -206,7 +207,7 @@ impl NodeActor for NetworkActor {
                     let sender_address = *handler.unsafe_block_signer_sender.borrow();
 
                     let payload_hash = block.payload_hash();
-                    let signature = signer.sign_block(payload_hash, chain_id, sender_address).await.map_err(|e| NetworkActorError::FailedToSignPayload(e.to_string()))?;
+                    let signature = signer.sign_block(payload_hash, chain_id, sender_address).await?;
 
                     let payload = OpNetworkPayloadEnvelope {
                         payload: block.execution_payload,


bin/node/src/flags/p2p.rs
```diff
@@ -400,12 +400,13 @@ impl P2PArgs {
...
-            .map(PrivateKeySigner::from_bytes)
-            .transpose()?
-            .map(|s| s.with_chain_id(Some(args.l2_chain_id.into())));
+        let local_signer =
+            self.sequencer_key.as_ref().map(PrivateKeySigner::from_bytes).transpose()?.map(|s| {
+                let signer: Arc<dyn BlockSigner + Send + Sync> =
+                    Arc::new(LocalSigner::from(s.with_chain_id(Some(args.l2_chain_id.into()))));
+
+                signer
+            });
```
refcell, line 409: Pulling this into a new function could allow us to unit test this a bit easier. Not a blocker for this PR, just a consideration.

crates/node/sources/src/signer/remote.rs
```diff
@@ -0,0 +1,385 @@
...
+
+        Ok(Some(task::spawn(Self::certificate_watcher_task(client, config, rx))))
+    }
+
+    async fn certificate_watcher_task(
+        client: Arc<RwLock<RpcClient>>,
+        config: RemoteSignerConfig,
+        rx: Receiver<Result<Event, notify::Error>>,
+    ) {
+        while let Ok(event) = rx.recv() {
```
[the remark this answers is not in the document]
refcell: This can be resolved by using tokio receiver as mentioned above

refcell (commented):
Overall looks good. Needs copilot suggestions for tokio channels

crates/node/sources/src/signer/remote.rs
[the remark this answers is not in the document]
theochap: I ended up removing the tokio channels and handle the events directly in the watcher

bin/node/src/flags/p2p.rs
theochap: Moving away from the traits allowed to remove that complicated piece of code

clabby (approved):
This looks great to me at a high-level. Glad to merge this and get the CLI flags and e2e tests in a follow-up.

crates/node/sources/src/signer/remote/handler.rs
```diff
@@ -0,0 +1,80 @@
...
+            self.client
+                .read()
+                .await
+                .request("opsigner_signBlockPayload", &params)
+                .await
+                .map_err(RemoteSignerError::SigningRPCError)?
+        };
+
+        // Parse the hex signature
+        let signature_bytes = hex::decode(response.signature.trim_start_matches("0x"))
```
clabby, line 68: Let's just use `alloy_primitives::hex` and not add a new dep.

PR #2589, opened 2025-07-29
patch(node/conductor): patch conductor bootstrap

### LINKED ISSUE #2571: fix(node/service): Support conductor bootstrap
## Overview

Currently, the `kona-node` supports operating within an `op-conductor` cluster. However, when bootstrapping a new cluster with [`op-conductor-ops`](https://github.com/ethereum-optimism/infra/tree/main/op-conductor-ops), `kona-node` fails to return an unsafe head pre-initialization.

We need to patch the `optimism_syncStatus` endpoint to return the current forkchoice of the EL prior to initialization, such that it never returns zero hashes prior to initialization.

theochap:
## Description

Close #2571

### DIFF
--- a/crates/node/engine/src/query.rs
+++ b/crates/node/engine/src/query.rs
@@ -87,7 +87,7 @@ impl EngineQueries {
         // safe head on the L1.
         if state.sync_state.unsafe_head().block_info.hash == B256::ZERO {
             let forkchoice_state =
-                L2ForkchoiceState::current(rollup_config, client.l2_provider()).await?;
+                L2ForkchoiceState::current(rollup_config, client.l2_engine()).await?;
             state.sync_state = EngineSyncState::new(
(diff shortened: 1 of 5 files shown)


crates/node/engine/src/query.rs
```diff
@@ -49,6 +50,9 @@ pub enum EngineQueries {
 /// An error that can occur when querying the engine.
 #[derive(Debug, thiserror::Error)]
 pub enum EngineQueriesError {
+    /// Failed to retrieve the L2 forkchoice state.
+    #[error("Failed to retrieve L2 forkchoice state: {0}")]
+    FailedToRetrieveL2ForkchoiceState(#[from] SyncStartError),
```
clabby, line 55: This is unused now

clabby (approved):

PR #2597, opened 2025-07-30
feat(supervisor:core): cyclic dependency checker on block validation

### LINKED ISSUE #2434: feat (supervisor/core): [Safety Checker] cyclic dependency check
- All candidate blocks at the same timestamp form a directed dependency graph.
- This graph must be acyclic — cycles imply mutual blocking and unsafe circular reasoning.

### How the Go Client Implements It

* Uses a recursive algorithm to expand all message dependencies at the same timestamp.
* Constructs a full message dependency graph.
* Detects cycles using **Kahn’s algorithm** (topological sort).
* Blocks are invalidated immediately if a cycle or invariant violation is found.

---

### Why This Doesn't Fit `kona-supervisor`

* The Rust design is **per-chain and modular** — each job handles only its own chain’s blocks.
* No recursive exploration of other chains’ dependencies is performed.
* Introducing recursion would violate isolation and introduce synchronization complexity.

---

## Proposed Design: Optimistic Promotion + Centralized Cycle Scanner

A hybrid design that maintains modular jobs but ensures eventual correctness.

---

### Promotion Jobs (Per Chain)

Each `SafetyCheckerJob`:

1. Attempts to promote its own candidate block optimistically.
2. If promotion fails (e.g., because a required block on another chain is not yet cross-safe), it records:

   ```
   Block A (candidate) → Block B (dependency)
   ```
3. This dependency is recorded in a shared **`DependencyGraph`**, which may be in memory or persisted.

```rust
type BlockID = (ChainId, BlockHash);

struct DependencyGraph {
    graph: HashMap<BlockID, HashSet<BlockID>>,
}
```

---

### Background Task: Periodic Cycle Scanner

* Runs periodically (e.g., every 5 seconds).
* Reads the `DependencyGraph`:

  1. **Runs Kahn’s algorithm** to detect cycles
  2. **On cycle detection**:
     - Invalidate the blocks
  3. **Clear the nodes from graph**

---

sadiq1971:
Closes #2434


dhyaniarun1993 (approved):
Nice. Great job

itschaindev (commented):
some debug logs would be good to have in `cross.rs`, otherwise lgtm.

crates/supervisor/core/src/safety_checker/cross.rs
```diff
@@ -25,36 +27,31 @@ where
...
-                self.verify_message_dependency(initiating_block, &message, required_level)?;
-            }
-        }
+    pub fn validate_block(&self, block: BlockInfo) -> Result<(), CrossSafetyError> {
+        self.for_each_dependent_block(&block, self.chain_id, |message, initiating_block| {
+            // Check whether the message passes interop timestamps related validation
+            self.validator
+                .validate_interop_timestamps(
+                    message.chain_id,  // initiating chain id
+                    message.timestamp, //initiating block timestamp
```
emhane, line 36:
```suggestion
                    message.timestamp, // initiating block timestamp
```

emhane (approved):

PR #2599, opened 2025-07-30
chore(workspace): Add `theochap` as author

clabby:
## Overview

Adds @theochap as an author; a bit overdue :)


clabby:
Needs #2598

PR #2614, opened 2025-07-31
Add observability to rewind progress

JoshdfG:


crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -477,13 +477,44 @@ where
...
+        }
+
         // Delete all derived blocks with number ≥ `block_info.number`
         {
             let mut cursor = self.tx.cursor_write::<DerivedBlocks>()?;
             let mut walker = cursor.walk(Some(block.number))?;
+            let mut processed_blocks = 0;
+            const LOG_INTERVAL: u64 = 100;
+
+            info!("Starting rewind of {} blocks...", total_blocks);
```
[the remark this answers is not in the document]
JoshdfG: @emhane thank you, I have updated the code.

emhane (changes requested):

crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -477,13 +477,52 @@ where
     pub(crate) fn rewind_to(&self, block: &BlockNumHash) -> Result<(), StorageError> {
         let block_pair = self.get_derived_block_pair(*block)?;
 
-        // Delete all derived blocks with number ≥ `block_info.number`
+        // Get the latest block number from DerivedBlocks
+        let latest_block = {
```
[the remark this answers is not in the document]
JoshdfG: sorry for the delay @emhane, I've been quite busy, I'd wait for @varun-doshi's response to know what to do next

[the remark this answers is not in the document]
dhyaniarun1993: Even without caching this is fine.

crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -491,13 +491,52 @@ where
...
             let mut walker = cursor.walk(Some(block.number))?;
+            let mut processed_blocks = 0;
+            const LOG_INTERVAL: u64 = 100;
+
+            info!(
+                target_block = ?block.number,
+                ?latest_block,
+                %total_blocks,
+                "Starting rewind derived block storage"
+            );
```
dhyaniarun1993, line 513:
```suggestion
            info!(
                target_block = ?block.number,
                chain_id = %self.chain_id,
                latest_block,
                total_blocks,
                "Starting rewind derived block storage"
            );
```

dhyaniarun1993 (commented):
Can you add the logic to the `log_provider` as well?

JoshdfG:
@dhyaniarun1993 Done. Hopefully this is what you meant.

crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -491,13 +491,55 @@ where
     pub(crate) fn rewind_to(&self, block: &BlockNumHash) -> Result<(), StorageError> {
         let block_pair = self.get_derived_block_pair(*block)?;
 
-        // Delete all derived blocks with number ≥ `block_info.number`
+        // Get the latest block number from DerivedBlocks
+        let latest_block = {
+            let mut cursor = self.tx.cursor_read::<DerivedBlocks>()?;
+            cursor.last()?.map(|(num, _)| num).unwrap_or(block.number)
+        };
```
emhane, line 498: what's the reason for getting the latest block like this now?

emhane (commented):
pending is test and screenshot. see prev review.

JoshdfG:
> let's also add a test that generates this log, see how tracing is enabled in tests
> 
> https://github.com/op-rs/kona/blob/34e4b54c28e3e484ba1566fdf6b11f639521b0b3/crates/node/sources/src/runtime/loader.rs#L178
> 
> 
> pls screenshot the logs and add picture to description of pr

Screenshot 2025-08-21 at 02 12 35

emhane:
> > let's also add a test that generates this log, see how tracing is enabled in tests
> > https://github.com/op-rs/kona/blob/34e4b54c28e3e484ba1566fdf6b11f639521b0b3/crates/node/sources/src/runtime/loader.rs#L178
> > 
> > pls screenshot the logs and add picture to description of pr
> 
> Okay, Done.

nice! the screenshot doesn't show the percentage progress, which is what we want to see, suggest either generating block refs with `arbitrary`, or letting the log interval be a field on the provider types, so that it can be set to a lower number in tests.

crates/supervisor/storage/src/providers/derivation_provider.rs
JoshdfG: Seems you don't like this approach used in generating the latest block, can you suggest a suitable approach?

JoshdfG:
> > > let's also add a test that generates this log, see how tracing is enabled in tests
> > > https://github.com/op-rs/kona/blob/34e4b54c28e3e484ba1566fdf6b11f639521b0b3/crates/node/sources/src/runtime/loader.rs#L178
> > > 
> > > pls screenshot the logs and add picture to description of pr
> > 
> > 
> > Okay, Done.
> 
> nice! the screenshot doesn't show the percentage progress, which is what we want to see, suggest either generating block refs with `arbitrary`, or letting the log interval be a field on the provider types, so that it can be set to a lower number in tests.

Screenshot 2025-08-24 at 07 45 59

emhane:
> > > > let's also add a test that generates this log, see how tracing is enabled in tests
> > > > https://github.com/op-rs/kona/blob/34e4b54c28e3e484ba1566fdf6b11f639521b0b3/crates/node/sources/src/runtime/loader.rs#L178
> > > > 
> > > > pls screenshot the logs and add picture to description of pr
> > > 
> > > 
> > > Okay, Done.
> > 
> > 
> > nice! the screenshot doesn't show the percentage progress, which is what we want to see, suggest either generating block refs with `arbitrary`, or letting the log interval be a field on the provider types, so that it can be set to a lower number in tests.
> 
> Screenshot 2025-08-24 at 07 45 59
> @emhane Can you check if this is okay? before i continue.

excellent! thanks!

JoshdfG:
> > > > > let's also add a test that generates this log, see how tracing is enabled in tests
> > > > > https://github.com/op-rs/kona/blob/34e4b54c28e3e484ba1566fdf6b11f639521b0b3/crates/node/sources/src/runtime/loader.rs#L178
> > > > > 
> > > > > pls screenshot the logs and add picture to description of pr
> > > > 
> > > > 
> > > > Okay, Done.
> > > 
> > > 
> > > nice! the screenshot doesn't show the percentage progress, which is what we want to see, suggest either generating block refs with `arbitrary`, or letting the log interval be a field on the provider types, so that it can be set to a lower number in tests.
> > 
> > 
> > Screenshot 2025-08-24 at 07 45 59
> > @emhane Can you check if this is okay? before i continue.
> 
> excellent! thanks!

@emhane I'm done, Please check if its okay.

crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -489,15 +489,97 @@ where
...
+
+                info!(
+                    target: "supervisor::storage",
+                    chain_id = %self.chain_id,
+                    block_number = %key,
+                    percentage = %format!("{:.2}%", percentage),
+                    processed_blocks,
+                    total_blocks,
+                    "Validated block for rewind"
+                );
```
emhane, line 556: we don't need this one, since we have the log below already

crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -489,15 +489,97 @@ where
...
+                "Starting rewind of derived block storage"
+            );
+
+            while let Some(Ok((key, _stored_block))) = walker.next() {
+                processed_blocks += 1;
+                let percentage = if total_blocks > 0 {
+                    (processed_blocks as f64 / total_blocks as f64 * 100.0).min(100.0)
+                } else {
+                    100.0
+                };
```
emhane, line 546: can move this down into the scope of guard `if processed_blocks % LOG_INTERVAL == 0 || processed_blocks == total_blocks`

crates/supervisor/storage/src/providers/log_provider.rs
```diff
@@ -142,23 +142,110 @@ where
...
-                    %stored_block,
-                    incoming_block = ?block,
-                    "Requested block to rewind does not match stored block",
+                    block_number = %key,
+                    block_hash = %stored_block.hash,
+                    percentage = %format!("{:.2}%", percentage),
+                    processed_blocks,
+                    total_blocks,
+                    "Validated block for rewind"
                 );
```
emhane, line 218: same as for block provider

crates/supervisor/storage/src/providers/log_provider.rs
```diff
@@ -142,23 +142,110 @@ where
...
+            while let Some(Ok((key, stored_block))) = walker.next() {
+                if key == block.number && block.hash != stored_block.hash {
+                    error!(
+                        target: "supervisor::storage",
+                        chain_id = %self.chain_id,
+                        %stored_block,
+                        incoming_block = ?block,
+                        "Requested block to rewind does not match stored block"
+                    );
+                    return Err(StorageError::ConflictError);
```
emhane, line 199: aware that this is out of scope of this pr, but this should probably rather be `SuperchainDAError::DataCorruption` @dhyaniarun1993

crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -489,15 +489,97 @@ where
...
+                chain_id = %self.chain_id,
+                target_block_number = %block.number,
+                target_block_hash = %block.hash,
+                latest_block,
+                total_blocks,
+                "Starting rewind of derived block storage"
+            );
+
+            while let Some(Ok((key, _stored_block))) = walker.next() {
+                processed_blocks += 1;
```
emhane, line 541: for this to be truly correct, we need to do call this instruction after we called `walker.delete_current()?;`, same for log provider

emhane (changes requested):
almost done!

crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -489,15 +489,97 @@ where
...
+                chain_id = %self.chain_id,
+                target_block_number = %block.number,
+                latest_block,
+                "Cannot rewind to future block"
+            );
+            return Err(StorageError::FutureData);
+        }
+
+        let total_blocks = latest_block.saturating_sub(block.number);
+        const LOG_INTERVAL: u64 = 100;
```
emhane, line 522: ah, so we also need to add a new field
```rust
pub(crate) struct DerivationProvider<'tx, TX> {
    tx: &'tx TX,
    chain_id: ChainId,
    #[doc(hidden)]
    observability_interval: usize,
}
```
so that in tests we can set this field very low. we will need an additional constructor `new_with_observability_interval` that takes an extra arg `observability_interval: Option<usize>`. in the existing constructor use this `LOG_INTERVAL` constant, which will have to be moved to the top of this module, and should be renamed to `DEFAULT_LOG_INTERVAL`.

emhane (changes requested):

crates/supervisor/storage/src/providers/derivation_provider.rs
JoshdfG: @emhane This ensures that we have the incremental percentage logging, I'll try and find another method, if we remove this, we won't get the percentages per blocks removed

emhane: in particular https://github.com/op-rs/kona/pull/2614#discussion_r2300808220

emhane (changes requested):

crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -489,15 +511,88 @@ where
...
+                // Remove the block
+                walker.delete_current()?;
+
+                processed_blocks += 1;
+
+                let percentage = if total_blocks > 0 {
+                    (processed_blocks as f64 / total_blocks as f64 * 100.0).min(100.0)
+                } else {
+                    100.0
+                };
```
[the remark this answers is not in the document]
JoshdfG: Okay, Done.

crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -7,20 +7,42 @@ use crate::{
...
+impl<'tx, TX> DerivationProvider<'tx, TX> {
+    pub(crate) const fn new(tx: &'tx TX, chain_id: ChainId) -> Self {
+        Self { tx, chain_id, observability_interval: DEFAULT_LOG_INTERVAL }
+    }
+
+    #[cfg(test)]
+    pub(crate) fn new_with_observability_interval(
+        tx: &'tx TX,
+        chain_id: ChainId,
+        observability_interval: Option<u64>,
```
emhane, line 38:
```suggestion
        observability_interval: u64,
```
it doesn't make sense to have an option here, my bad if i said to do that. this is since if `None` is passed as last arg then the regular constructor might as well be used.

crates/supervisor/storage/src/providers/derivation_provider.rs
```diff
@@ -7,20 +7,42 @@ use crate::{
...
 pub(crate) struct DerivationProvider<'tx, TX> {
     tx: &'tx TX,
     chain_id: ChainId,
+    #[doc(hidden)]
+    observability_interval: u64,
+}
+
+impl<'tx, TX> DerivationProvider<'tx, TX> {
+    pub(crate) const fn new(tx: &'tx TX, chain_id: ChainId) -> Self {
+        Self { tx, chain_id, observability_interval: DEFAULT_LOG_INTERVAL }
```
emhane, line 31:
```suggestion
        Self::new_with_observability_interval(tx, chain_id, DEFAULT_LOG_INTERVAL)
```

crates/supervisor/storage/src/providers/log_provider.rs
```diff
@@ -84,7 +84,7 @@ where
                 incoming_block = %block,
                 "Incoming log block is not consistent with the stored log block",
             );
-            return Err(StorageError::ConflictError)
+            return Err(StorageError::ConflictError);
```
emhane, line 87:
```suggestion
            return Err(StorageError::ConflictError)
```
pls revert unrelated format changes like any new semicolons. it makes the diff unnecessarily large and makes the pr have scope creep. make sure to run the formatter with nightly toolchain only, ie `cargo +nightly fmt`

emhane (changes requested):
make sure to apply the review changes to log storage provider too pls

PR #2641, opened 2025-08-04
feat(node/p2p): allow disabling the bootstore

theochap:
## Description

This PR adds a CLI argument to disable writing/recovering the bootstore to/from a file. This is useful for our integration tests because we want to ensure we always have the same behavior between runs.

Revamps the bootstore

### DIFF
--- a/crates/node/service/src/actors/network/actor.rs
+++ b/crates/node/service/src/actors/network/actor.rs
@@ -249,11 +249,7 @@ impl NodeActor for NetworkActor {
                         warn!(target: "node::p2p", "Failed to send unsafe block to network handler");
                     }
                 },
-                req = self.p2p_rpc.recv(), if !self.p2p_rpc.is_closed() => {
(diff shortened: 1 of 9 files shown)


crates/node/service/src/actors/network/actor.rs
```diff
@@ -249,11 +249,7 @@ impl NodeActor for NetworkActor {
                         warn!(target: "node::p2p", "Failed to send unsafe block to network handler");
                     }
                 },
-                req = self.p2p_rpc.recv(), if !self.p2p_rpc.is_closed() => {
-                    let Some(req) = req else {
-                        error!(target: "node::p2p", "The p2p rpc receiver channel has closed");
-                        return Err(NetworkActorError::ChannelClosed);
-                    };
+                Some(req) = self.p2p_rpc.recv(), if !self.p2p_rpc.is_closed() => {
```
clabby, line 252: Good catch

PR #2644, opened 2025-08-04
chore(node): Untrusted RPC Validation

### LINKED ISSUE #2318: feat(node): Support Untrusted RPC Validation
### Description

The `op-node` constructs eth clients with a [special flag](https://github.com/ethereum-optimism/optimism/blob/e12f48b39254088007720bdf535b77fa4686e8be/op-service/sources/eth_client.go#L55-L59) indicating whether to trust the rpc. If the RPC is local, we can trust it, but if it's over the wire, cached information from responses shouldn't be used and block-hash verification should be performed. As the `op-node` describes, deposits and batches can't be missed or faked, but the wrong L1 block could be fetched.

refcell:
### Description

Add untrusted rpc validation.

By default, this leaves rpc endpoints as _trusted_ to avoid the performance hit.

Down the road, these flags can be set to `false` by default which will force the providers to validate rpc responses.

Closes #2318

### DIFF
--- a/bin/node/README.md
+++ b/bin/node/README.md
@@ -78,8 +78,10 @@ kona-node \
 Many configuration options can be set via environment variables:
 
 - `KONA_NODE_L1_ETH_RPC` - L1 execution client RPC URL
+- `KONA_NODE_L1_TRUST_RPC` - Whether to trust the L1 RPC without verification (default: true)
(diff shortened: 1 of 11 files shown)


refcell:
## 📚 $\text{Stack Overview}$

* #2644 👈
* `main`

clabby (approved):

Issue #2623, opened 2025-08-01
[TRACKING] develop an internal testing framework to test node actors. Increase internal node coverage


theochap:
## Description

Now that we are past the Arid release (1.0-rc.1), we should increase testing efforts for the `kona-node`. One of the main selling points of the `kona-node` is its _reliability_. We should keep that in mind when writing new tests/frameworks.

The internal testing framework should allow to test the node actors in isolation. It should ensure that node actors behave properly under different sets of inputs and that we cover all the possible execution paths (happy/unhappy paths). 

## Tracking

Here is a tentative list of steps required to properly implement the testing framework. Ideally, the development of such a system should be driven by tests. Hence, we should iteratively build coverage to the different actors and add abstractions as we go. The task list will then match the list of node actors to test out.

- [ ] Engine Actor tests
- [ ] Derivation Actor tests
- [ ] L1Watcher Actor tests
- [ ] Sequencer Actor tests
- [ ] Rpc Actor tests
- [ ] Network Actor tests #2679 

Once all the actors are tested in isolation, we should write a couple of e2e actor tests that ensure the actor interactions are covered.

PR #2679, opened 2025-08-11
feat(net/test): add integration test skeleton

theochap:
## Description

Bootstraps the integration tests crate for kona services. For now we only have a very basic p2p test that spins off two networks and perform a check that those two are properly connected to each other

Progress on #2623


bin/node/src/commands/bootstore.rs
```diff
@@ -59,8 +59,12 @@ impl BootstoreCommand {
...
             .ok_or(anyhow::anyhow!("Chain ID {} not found in the registry", chain_id))?;
         println!("{} Bootstore (Chain ID: {})", chain.name, chain_id);
-        let bootstore = BootStore::from_chain_id(chain_id, self.bootstore.clone(), vec![]);
-        println!("Path: {}", bootstore.path.display());
+        let bootstore: BootStoreFile = self
+            .bootstore
+            .clone()
+            .map_or(BootStoreFile::Default { chain_id }, BootStoreFile::Custom);
+        let bootstore: BootStore = bootstore.try_into()?;
+        println!("Path: {}", self.bootstore.clone().unwrap_or_default().display());
```
refcell, line 67: Do we need to clone here?

crates/node/peers/src/store.rs
```diff
@@ -18,21 +18,92 @@ const MAX_PEERS: usize = 2048;
...
+    debug!(target: "bootstore", "Reading boot store from disk: {:?}", file);
+    let reader = BufReader::new(file);
+    match serde_json::from_reader(reader).map(|s: BootStore| s.peers) {
+        Ok(peers) => peers,
+        Err(e) => {
+            warn!(target: "bootstore", "Failed to read boot store from disk: {:?}", e);
+            VecDeque::new()
+        }
+    }
+}
```
refcell, line 99: This feels like it could be an associated method on the `BootStoreFile` so it's nicely exported and accessible without having a private, hidden function.

refcell (approved):
This looks fine to me

PR #2680, opened 2025-08-11
fix(supervisor/core): [Safety Checker] block validation

sadiq1971:
Refactors `map_dependent_block` to pass a lazy fetcher closure instead of fetching initiating blocks eagerly.
validate_block now verifies dependencies before fetching, preventing false validation error.


dhyaniarun1993 (approved):
looks good to go.

PR #2685, opened 2025-08-12
fix(supervisor): reorg at startup

### LINKED ISSUE #2683: fix(supervisor): reorg running at the startup

dhyaniarun1993:
Closes #2683


crates/supervisor/core/src/l1_watcher/watcher.rs
```diff
@@ -148,43 +148,49 @@ where
     async fn handle_new_latest_block(
         &self,
         incoming_block: Block,
-        previous_block: &mut BlockNumHash,
+        previous_block: &mut Option<BlockNumHash>,
```
emhane, line 151: prefer if this method returns the new previous block (which could be the same as the argument) instead of passing a mutable reference - would that work here?

emhane (changes requested):

PR #2687, opened 2025-08-12
chore: fix function name

hongmengning:
fix function name

### DIFF
--- a/tests/node/mixed_preset.go
+++ b/tests/node/mixed_preset.go
@@ -55,7 +55,7 @@ func (m *MixedOpKonaPreset) L2CLNodes() []dsl.L2CLNode {
 	return append(m.L2CLOpNodes, m.L2CLKonaNodes...)
 }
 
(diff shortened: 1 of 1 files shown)


tests/node/mixed_preset.go
```diff
@@ -55,7 +55,7 @@ func (m *MixedOpKonaPreset) L2CLNodes() []dsl.L2CLNode {
 	return append(m.L2CLOpNodes, m.L2CLKonaNodes...)
 }
 
-// L2CLNodes returns all the L2CL nodes in the network (op-nodes and kona-nodes).
+// L2ELNodes returns all the L2CL nodes in the network (op-nodes and kona-nodes).
```
emhane, line 58:
```suggestion
// L2ELNodes returns all the L2EL nodes in the network (op-reth, op-geth, etc.).
```

emhane (dismissed):
good catch!

PR #2693, opened 2025-08-12
feat(node/p2p): extend network actor test with block propagation. reduce test flakiness

theochap:
## Description

This PR achieves the following goals:
- Extend the simple p2p connection tests of #2679 with more comprehensive tests involving block signature and propagation (a follow PR will add the cases for higher block versions, it requires updating the rollup config with hardfork times)
- Fixes the network bootstrap logic to remove flakiness. In particular, it uses kernel allocated ports for discovery/gossip to avoid static modification races. 
- It adds random valid block building logic


bin/node/src/commands/bootstore.rs
```diff
@@ -59,8 +59,12 @@ impl BootstoreCommand {
...
             .ok_or(anyhow::anyhow!("Chain ID {} not found in the registry", chain_id))?;
         println!("{} Bootstore (Chain ID: {})", chain.name, chain_id);
-        let bootstore = BootStore::from_chain_id(chain_id, self.bootstore.clone(), vec![]);
-        println!("Path: {}", bootstore.path.display());
+        let bootstore: BootStoreFile = self
+            .bootstore
+            .clone()
+            .map_or(BootStoreFile::Default { chain_id }, BootStoreFile::Custom);
+        let bootstore: BootStore = bootstore.try_into()?;
+        println!("Path: {}", self.bootstore.clone().unwrap_or_default().display());
```
refcell, line 67: ibid - maybe we can avoid the clone?

crates/node/peers/src/store.rs
```diff
@@ -55,9 +126,6 @@ impl BootStore {
     /// [`BootStore::sync`] prior to dropping the store.
     pub fn add_enr(&mut self, enr: Enr) {
         self.add_rotate(enr);
-        if let Err(e) = self.write_to_file() {
-            warn!(target: "bootstore", "Failed to write boot store to disk: {:?}", e);
-        }
```
refcell, line 60: I'm not super opinionated here but this will change how frequently the bootstore is updated. I'd prefer to make this configurable (and off by default if we want) rather than remove it entirely. Not a big deal though i don't think.

crates/node/gossip/src/driver.rs
```diff
@@ -196,16 +196,18 @@ where
                 {
                     if id == listener_id {
                         info!(target: "gossip", "Swarm now listening on: {address}");
-                        break;
+
+                        self.addr = address.clone();
+
+                        return Ok(address);
```
refcell, line 202: I don't think we need the return here since the `Ok(())` was removed. Should be able to just be:
```suggestion
                        Ok(address)
```

crates/node/service/src/actors/network/driver.rs
```diff
@@ -30,17 +38,46 @@ pub enum NetworkDriverError {
...
+                .find_map(|p| match p {
+                    Protocol::Tcp(port) => Some(port),
+                    _ => None,
+                })
+                .ok_or_else(|| {
+                    NetworkDriverError::InvalidGossipListenAddr(gossip_listen_addr.clone())
+                })?;
+
+            self.discovery.disc.update_local_enr_socket(SocketAddr::new(ip_address, port), true);
+        }
```
refcell, line 76: This is a good candidate for a refactor in a separate method.

theochap: Would that really increase readability? We're not really doing that much here and we don't really have this pattern elsewhere. I can refactor this out if we ever decide to add this pattern in other parts of the codebase

crates/node/peers/src/store.rs
theochap: So this comment and the comment above are actually related to https://github.com/op-rs/kona/pull/2641. I don't think removing this sync method would fundamentally change how we're updating the bootstore. We still have a task that updates the bootstore frequently enough inside the network actor.

PR #2696, opened 2025-08-13
chore(docker): Add back cannon reproducible build

clabby:
## Overview

Adds the reproducible build recipe for `cannon` back in.


theochap (approved):
Looks good, should we add a CI task to test that this doesn't break in the future?

PR #2699, opened 2025-08-13
feat (supervisor): verify L1 chain consistency at startup

### LINKED ISSUE #2662: feat(supervisor): Verify consistency at startup
# Description

Currently, `handle_l1_reorg` is only triggered when the L1Watcher detects a mismatch (i.e., a reorg event during runtime). However, if the supervisor restarts, the previously persisted state may be out of sync with the current L1 chain due to a reorg that happened while the supervisor was offline.

Why this matters:
- If a reorg occurred while the supervisor was down, the stored chain state may reference blocks that are no longer canonical.
- On startup, the supervisor should verify and, if necessary, reorg all chains to match the current L1 canonical chain.

itschaindev:
Closes #2662 

The `ReorgTask` already contains the logic for verifying the consistency. I'm reusing it at startup by calling the `handle_l1_reorg` when initializing the l1 watcher.

### DIFF
--- a/crates/supervisor/service/src/service.rs
+++ b/crates/supervisor/service/src/service.rs
@@ -1,10 +1,13 @@
 //! Contains the main Supervisor service runner.
 
+use alloy_eips::BlockNumberOrTag;
 use alloy_primitives::ChainId;
 use alloy_provider::{RootProvider, network::Ethereum};
 use alloy_rpc_client::RpcClient;
+use alloy_rpc_types_eth::Block;
 use anyhow::Result;
 use jsonrpsee::client_transport::ws::Url;
+use kona_protocol::BlockInfo;
 use kona_supervisor_core::{
     ChainProcessor, CrossSafetyCheckerJob, ReorgHandler, Supervisor,
     config::Config,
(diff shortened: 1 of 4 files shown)


emhane (approved):
nice, appreciate the helpful code comments

crates/supervisor/service/src/service.rs
```diff
@@ -248,15 +248,33 @@ impl Service {
...
+                trace!(target: "supervisor::service", "Startup reorg check completed");
+            }
+
+            // Start the L1 watcher streaming loop.
+            let l1_watcher = L1Watcher::new(
+                l1_rpc.clone(),
+                database_factory,
+                event_senders,
+                cancel_token,
+                ReorgHandler::new(l1_rpc, chain_dbs_map, managed_nodes),
```
dhyaniarun1993, line 275: We can pass the `reorg_handler` here

crates/supervisor/core/src/l1_watcher/watcher.rs
```diff
@@ -88,13 +88,13 @@ where
                 }
                 latest_block = latest_head_stream.next() => {
                     if let Some(latest_block) = latest_block {
-                        info!(target: "supervisor::l1_watcher", "Latest L1 block received: {:?}", latest_block.header.number);
+                        trace!(target: "supervisor::l1_watcher", "Latest L1 block received: {:?}", latest_block.header.number);
```
dhyaniarun1993, line 91: instead of logging here, Can we log in the `handle_new_latest_block` method?
Also please log the previous and incoming block

crates/supervisor/core/src/l1_watcher/watcher.rs
```diff
@@ -88,13 +88,13 @@ where
...
                         previous_latest_block = self.handle_new_latest_block(latest_block, previous_latest_block).await;
                     }
                 }
                 finalized_block = finalized_head_stream.next() => {
                     if let Some(finalized_block) = finalized_block {
-                        info!(target: "supervisor::l1_watcher", "Finalized L1 block received: {:?}", finalized_block.header.number);
+                        trace!(target: "supervisor::l1_watcher", "Finalized L1 block received: {:?}", finalized_block.header.number);
```
dhyaniarun1993, line 97: instead of logging here, Can we log in the handle_new_latest_block method?
Also please log the previous and incoming block

dhyaniarun1993 (approved):

PR #2763, opened 2025-08-25
chore(providers): Local Buffered Providers

refcell:
### Description

Local buffered providers for execution extension support.


crates/providers/providers-local/src/lib.rs
```diff
@@ -0,0 +1,58 @@
...
+//! ## Features
+//!
+//! - **BufferedL2Provider**: A pure in-memory L2 provider that serves data from cached blocks
+//! - **ChainStateBuffer**: LRU cache for managing chain state with reorg support
+//! - **Chain Event Handling**: Support for processing execution extension notifications
+//! - **No External Dependencies**: Operates entirely from in-memory state without RPC calls
+//!
+//! ## Usage
+//!
+//! ```rust,ignore
```
theochap, line 22: Same for `rust,ignore`

refcell: Will remove these docs

crates/providers/providers-local/src/buffer.rs
```diff
@@ -0,0 +1,478 @@
...
+    /// LRU cache for blocks by hash
+    blocks_by_hash: RwLock<LruCache<B256, CachedBlock>>,
+    /// LRU cache for blocks by number
+    blocks_by_number: RwLock<LruCache<u64, B256>>,
+    /// Current canonical chain head
+    canonical_head: RwLock<Option<B256>>,
+    /// Maximum reorg depth to support
+    max_reorg_depth: u64,
+    /// Cache capacity
+    capacity: usize,
```
theochap, line 102: Do we need to store the capacity field in the struct? We are already using it to build both caches

crates/providers/providers-local/src/buffer.rs
```diff
@@ -0,0 +1,478 @@
...
+            kona_macros::set!(gauge, Metrics::REORG_DEPTH, depth as f64);
+        }
+
+        // Update canonical head
+        let mut canonical_head = self.canonical_head.write().await;
+        *canonical_head = Some(new_head);
+
+        // We need to invalidate cached blocks that are no longer canonical
+        // For now, we'll clear the entire cache on deep reorgs
+        if depth > 10 {
```
theochap, line 229: Why are we hardcoding a value here?

PR #2764, opened 2025-08-25
chore (supervisor): add observability to chain processor

### LINKED ISSUE #2743: chore(supervisor): add observability in chain processor

itschaindev:
Closes #2743 

If the logging is too verbose, I can enable only for `local_safe` and `finalized`.


itschaindev:
Closing it because:
- The current way of implementation is redundant and can be done with changing from `tracing` to `info` inside the handlers.
- For logging at certain interval, the logs will miss the blocks and can cause confusion to users looking at the logs as to why are the logs being skipped.

Till we find a better way of doing it, closing this PR.

PR #2770, opened 2025-08-26
fix(test/kona): fix l1 confirmation delay for kona and sequence attribute builder pipeline

theochap:
## Description
Try to fix the l1 confirmation delay in the devstack for kona which should reduce transfer flakiness in sysgo
Fixes the sequencer attribute builder pipeline to include txs from the mempool automatically

Depends on #2768


theochap:
Superseeded by #2783

PR #2774, opened 2025-08-26
chore (supervisor): build info in metrics

### LINKED ISSUE #2760: chore (supervisor): add app build metrics
Add the following metrics to prometheus and use it in grafana dashboard:

- Build Timestamp
- App Version
- Git SHA
- Build Profile
- Target Triple

itschaindev:
Closes #2760 

The Overview now looks like:

Screenshot 2025-08-26 at 19 30 52


bin/supervisor/src/flags/version.rs
```diff
@@ -0,0 +1,59 @@
+//! [`VersionInfo`] metrics
```
dhyaniarun1993, line 1: Defining this in the flag could be misleading. Can we move it to the `bin/supervisor/src/metrics/version.rs`

dhyaniarun1993: I think we ca remove the file since metrics argument is already being tested in the utility crate

dhyaniarun1993 (approved):

PR #2782, opened 2025-08-27
chore(proof): Remove duplicate FlushableCache trait bound in OraclePipeline

Kourin1996:
This PR just removes duplicate `FlushableCache` trait bound in `OraclePipeline`


refcell (approved):
Strange ci didn't catch this

PR #2790, opened 2025-08-29
docs: fix incorrect crates.io link path

saylor-mik87786:
incorrect: https://crates.io/crate/kona-protocol
correct: https://crates.io/crates/kona-protocol

before:
image

after:
image


emhane (approved):
`lychee` workflow didn't catch this, thx

PR #2793, opened 2025-08-29
Fix typos

bethoffman:


emhane (approved):
thanks, seems like `typos` workflow didn't catch these

PR #2794, opened 2025-08-29
fix: correct SHA short length in rollup build script

strmfos:
Fix inconsistency in rollup build.rs where sha_short variable was defined as 7 characters (&sha[0..7]) but the environment variable VERGEN_GIT_SHA_SHORT was set using 8 characters (&sha[..8]).


bin/rollup/build.rs
```diff
@@ -29,7 +29,7 @@ fn main() -> Result<(), Box<dyn Error>> {
 
     emitter.emit_and_set()?;
     let sha = env::var("VERGEN_GIT_SHA")?;
-    let sha_short = &sha[0..7];
+    let sha_short = &sha[0..8];
```
emhane, line 32: can you use a constant here pls so it's not a magic number

emhane (changes requested):

emhane (approved):

PR #2801, opened 2025-09-01
fix: correct rollup version module documentation and environment variables

avorylli:
Fix rollup binary version module inconsistencies:
- Update module documentation to reference kona-rollup instead of kona-node
- Correct environment variable references to use KONA_ROLLUP_* prefix
- Ensures rollup binary has distinct version constants separate from node binary

### DIFF
--- a/bin/rollup/src/version.rs
+++ b/bin/rollup/src/version.rs
@@ -1,4 +1,4 @@
-//! Version information for kona-node.
+//! Version information for kona-rollup.
 
 #![allow(dead_code)]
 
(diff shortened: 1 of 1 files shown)


emhane:
hm, this change is not helpful. if you want to contribute, pls find issues labelled good-first-issue in the repo issues, thx.

emhane:
on second look this might be correct, @theochap ?

theochap:
@avorylli you'll just need to fix the lint CI job and we can get that one in

PR #2803, opened 2025-09-02
fix: ChannelOut last-frame flag set after read

Fibonacci747:
Set Frame.is_last only after reading compressed data, marking last when the channel is closed and no bytes remain. Prevents premature “last” frames that could break downstream reassembly logic and aligns with protocol::channel expectations and docs.

### DIFF
--- a/crates/batcher/comp/src/channel_out.rs
+++ b/crates/batcher/comp/src/channel_out.rs
@@ -125,9 +125,10 @@ where
             return Err(ChannelOutError::MaxFrameSizeTooSmall);
         }
 
-        // Construct an empty frame.
+        // Construct an empty frame with a provisional `is_last = false`.
+        // The `is_last` flag is determined after reading based on remaining bytes.
         let mut frame =
-            Frame { id: self.id, number: self.frame_number, is_last: self.closed, data: vec![] };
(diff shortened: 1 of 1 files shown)


Fibonacci747:
> > Set Frame.is_last only after reading compressed data, marking last when the channel is closed and no bytes remain. Prevents premature “last” frames that could break downstream reassembly logic and aligns with protocol::channel expectations and docs.
> 
> Would you mind linking the logic you're referring to? This is a non trivial change so you should provide some more justifications as of why it is needed

is_last was set before reading data from compressor, which could create multiple "last" frames violating the protocol requirement of exactly one is_last = true per channel.
in crates/protocol/protocol/src/frame.rs:126
- protocol requires "Exactly one frame per channel should have this flag set to true"
in crates/protocol/protocol/src/channel.rs:155-174
- is_ready() logic depends on proper is_last setting
crates/protocol/derive/src/stages/frame_queue.rs:78-82
- drops frames after is_last
Fix: Set is_last = true only after reading data when self.closed && self.ready_bytes() == 0, ensuring the flag is set only when no more data remains.

theochap:
After reading the change more closely I doubt that there can be a situation where `is_closed` is true and the frame is not the last frame. Do you have an example of situation that can trigger that?

PR #2804, opened 2025-09-02
feat(node/test): add l2 reorg tests

theochap:
## Description

Add a couple of l2 reorg tests with an appropriate network configuration.


refcell (approved):
Approving to unblock. One nit otherwise

PR #2810, opened 2025-09-03
fix(docs): correct return type documentation for header_by_hash

GarmashAlex:
Fix incorrect return type documentation in TrieDBProvider::header_by_hash

The function signature returns `Result<Header, Self::Error>` but the documentation
incorrectly stated `Ok(Bytes): The [Header]`. This was inconsistent with the
established pattern in the codebase where documentation must exactly match


theochap:
According to our [contribution guidelines](https://github.com/op-rs/kona/blob/main/CONTRIBUTING.md), we won't be accepting PRs to fix grammar and minor typos. We thank you for contributing to kona!

PR #2811, opened 2025-09-04
refactor: replace custom msm_required_gas with revm implementation

mikemillaa:
Replace duplicate `msm_required_gas` implementation with revm's official function.

The TODO comment indicated this should be done once the msm module became public. 
Since revm v26+, `bls12_381_utils::msm_required_gas` is now publicly available.


mikemillaa:
> seems like lint needs fixing

FIxed now

PR #2816, opened 2025-09-05
Add UL CL bootnodes

trianglesphere:
Uniswap labs runs bootnodes dedicated to the CL as well as to the EL.
I have added rather than replaced because both EL and CL bootnodes
are present in this file.


emhane (approved):

theochap (changes requested):
@trianglesphere would you mind providing a link to the source/origin of those enodes? Just making sure we're not merging anything without checking

trianglesphere:
CL bootnodes for op-node are stored here: https://github.com/ethereum-optimism/optimism/blob/f2e2921bc62a61198340e517fee07315b06a9733/op-node/p2p/config.go#L28-L46

Realistically, all of them should be switched over otherwise it will take longer for kona-node to discover other OP Stack peers (though there is heavy cross talk on discv5).

trianglesphere:
@theochap it turns out you use both EL and CL bootnodes so I have added rather than replaced the UL bootnodes.

I also force pushed to fix lint and rebase on latest

theochap (approved):
Thanks for providing more explanations!

theochap:
@trianglesphere you'll just have to update the tests and I can merge

PR #2843, opened 2025-09-10
fix(node): correct RPC environment variable semantic mismatch

mikemillaa:
The RPC configuration had a confusing semantic mismatch between the CLI flag and environment variable. The flag `--rpc.disabled` expects a boolean to disable RPC when true, but the environment variable was named `KONA_NODE_RPC_ENABLED` which suggests the opposite behavior.

This caused user confusion where setting `KONA_NODE_RPC_ENABLED=true` would actually disable the RPC server instead of enabling it. Users naturally expect an "ENABLED" variable to enable functionality when set to true.

Changed the environment variable from `KONA_NODE_RPC_ENABLED` to `KONA_NODE_RPC_DISABLED` to match the CLI flag semantics. Also updated the comment from "Whether to enable" to "Whether to disable" for consistency.


theochap (approved):

emhane (approved):
thanks

Issue #1427, opened 2025-04-14
chore(book): Update execution customization section


clabby:
## Overview

With https://github.com/op-rs/kona/pull/1400, the semantics of customizing the EVM that `kona` uses changes. There's a [section in the book](https://op-rs.github.io/kona/proof/exec-ext.html) about customizing the executor that will need to be updated after this PR lands.

PR #2844, opened 2025-09-10
docs(book): update executor customization for alloy-evm

mikemillaa:
Align the “executor customization” section with the changes from PR #1400 (switch to alloy-evm for stateless block building).

close https://github.com/op-rs/kona/issues/1427


mikemillaa:
@theochap Is this good to go? holding for a long time, and #2860 and #2871 cc @emhane

PR #2845, opened 2025-09-10
chore(cli&rollup): more cleanup of `anyhow` in CLI and rollup

mikemillaa:
More cleanup of `anyhow` in CLI and rollup after https://github.com/op-rs/kona/issues/916


bin/rollup/src/cli.rs
```diff
@@ -23,7 +22,7 @@ pub struct Cli {
 
 impl Cli {
     /// Runs the rollup binary.
-    pub fn run(self) -> Result<()> {
-        unimplemented!("Rollup CLI is not yet implemented")
+    pub fn run(self) -> RollupResult<()> {
+        Err(crate::error::RollupError::NotImplemented)
```
emhane, line 26: pls undo this change

bin/rollup/src/error.rs
emhane, line 1: pls remove this file, we don't need it because we keep the `unimplemented` panic. use `CliError` directly where needed

crates/utilities/cli/src/error.rs
```diff
@@ -0,0 +1,39 @@
...
+        chain_id: u64,
+    },
+
+    /// Error initializing metrics.
+    #[error("Failed to initialize metrics: {source}")]
+    MetricsInitialization {
+        /// The underlying error.
+        #[source]
+        source: Box<dyn std::error::Error + Send + Sync>,
+    },
```
emhane, line 35: since these just have one field, they can be a 1-ary tuple for sake of simplicity

emhane, line 35:
```suggestion
    MetricsInitialization(#[from] BuildError),
```
this variant should use ` metrics_exporter_prometheus::BuildError`

PR #2848, opened 2025-09-10
feat(node/sequencer): spike attempt to unblock engine when building on a sequencer payload

theochap:
## Description

We are currently blocking the engine from processing other tasks when payloads are being built. Instead we may want to be able to let the EL build payloads in the background and process requests in the meantime.

The changes to get this working are not that big, but it seems there is some race condition between the l1 blocks and the sequencer which may cause issues.

### DIFF
--- a/crates/node/engine/src/task_queue/tasks/build/task.rs
+++ b/crates/node/engine/src/task_queue/tasks/build/task.rs
@@ -13,11 +13,8 @@ use kona_genesis::RollupConfig;
 use kona_protocol::{L2BlockInfo, OpAttributesWithParent};
 use op_alloy_provider::ext::engine::OpEngineApi;
 use op_alloy_rpc_types_engine::{OpExecutionPayload, OpExecutionPayloadEnvelope};
-use std::{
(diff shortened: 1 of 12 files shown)


crates/node/engine/src/task_queue/tasks/build/task.rs
```diff
@@ -50,24 +47,47 @@ pub struct BuildTask {
...
-        Self { engine, cfg, attributes, is_attributes_derived, payload_tx }
+        Self {
+            engine,
+            cfg,
+            next_attributes,
+            is_attributes_derived,
+            last_payload_data,
+            payload_id_tx,
+        }
     }
```
emhane, line 91: this can be derived with `derive_more::Constructor`

emhane (approved):
cool!

op-will (commented):
The general approach looks good, but I think there may be a number of cases in which `payload_id` and `last_payload_data` can get out of sync between the Engine and the Sequencer.

theochap:
Closing stale PR.

PR #2857, opened 2025-09-12
fix: type parsing and update Discv5 configuration for compatibility

santamasa:
fixed a type parsing issue by specifying `Directive` explicitly, so the compiler knows exactly what to expect:

```rust
add_directive("discv5=error".parse::<Directive>()?)
```
additionally, direct assignments to `discovery.interval` and `discovery.forward` are now replaced with `ConfigBuilder`.
this ensures settings are applied without touching private fields.

also switched to `build_with_handler()` instead of `build()`, allowing retrieval of `handler` and `enr_receiver` directly, keeping compatibility with the latest discv5 API.

these changes remove compiler errors and align the code with the current discv5 API conventions.


emhane:
are you still working on this @santamasa ?

PR #2860, opened 2025-09-14
feat: add inflight task metrics to engine queue

mikemillaa:
track inflight tasks by type in task queue

- Add ENGINE_INFLIGHT_TASKS gauge metric with task type labels
- Increment on enqueue, decrement on completion and queue clear
- Conditionally compiled with 'metrics' feature

Seeing occasional slowdowns during high load periods
but couldn't tell if it was build tasks backing up or derivation
falling behind. This gives us per-task-type visibility to identify
which part of the pipeline is the bottleneck.

### DIFF
--- a/crates/node/engine/src/metrics/mod.rs
+++ b/crates/node/engine/src/metrics/mod.rs
@@ -72,6 +72,10 @@ impl Metrics {
     /// Identifier for the counter that tracks the number of times the engine has been reset.
     pub const ENGINE_RESET_COUNT: &str = "kona_node_engine_reset_count";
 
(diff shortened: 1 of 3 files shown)


emhane (changes requested):
can you pls add a grafana panel for this to the dashboard and a screenshot of that panel to the pe description

mikemillaa:
> can you pls add a grafana panel for this to the dashboard and a screenshot of that panel to the pe description

@emhane Thanks for the review! I've added the requested Grafana panel to the existing dashboard.

mikemillaa:
@emhane @theochap

PR #2866, opened 2025-09-16
fix: avoid redundant hash computation in block_info_and_transactions_by_hash

mikemillaa:
Remove unnecessary `hash_slow()` call when creating BlockInfo in the alloy chain provider.

Since we already verify the header hash matches the requested hash in `verify_header_hash()`, we can directly use the input hash parameter instead of recomputing it.


emhane (approved):
nice

PR #2871, opened 2025-09-17
feat(node): add sequencer key file loading support

mikemillaa:
Fixes design inconsistency where sequencer keys could only be passed via CLI args while regular P2P keys support file loading.

This adds `--p2p.sequencer.key.path` parameter to match the existing `--p2p.priv.path` pattern.

This resolves the[ TODO mentioned](https://github.com/ethereum-optimism/optimism/blob/44d3429d2913c8c90fbb0848e14c1022b8ca2267/op-devstack/sysgo/l2_cl_kona.go#L196) in optimism's devstack where they noted the missing file support for sequencer keys.


bin/node/src/flags/signer.rs
```diff
@@ -21,6 +22,14 @@ pub struct SignerArgs {
         conflicts_with = "endpoint"
     )]
     pub sequencer_key: Option<B256>,
+    /// An optional path to a file containing the sequencer private key.
+    /// This is mutually exclusive with `p2p.sequencer.key`.
+    #[arg(
+        long = "p2p.sequencer.key.path",
+        env = "KONA_NODE_P2P_SEQUENCER_KEY_PATH",
+        conflicts_with = "endpoint"
+    )]
```
emhane, line 31: should probably add `p2p.sequencer.key` to `conflicts_with` attribute then

emhane (changes requested):

PR #2874, opened 2025-09-18
feat(jovian/min-base-fee): implement min-base fee in kona

theochap:
## Description

Implements min-base-fee in kona-node and kona-proof. Ensures that the payload attributes are correctly parsed and that the computed base fee in kona-proof cannot be below the min-base fee.


crates/proof/executor/src/builder/env.rs
```diff
@@ -80,31 +100,35 @@ where
...
+            }
+        } else if config.is_canyon_active(payload_attrs.payload_attributes.timestamp) {
+            // If the payload attribute timestamp is past canyon activation,
+            // use the canyon base fee params from the rollup config.
+            Ok((config.chain_op_config.as_canyon_base_fee_params(), 0))
+        } else {
+            // If the payload attribute timestamp is prior to canyon activation,
+            // use the default base fee params from the rollup config.
+            Ok((config.chain_op_config.as_base_fee_params(), 0))
+        }
```
emhane, line 132: this block of code seems weird, how about just checking jovian up top, etc. and returning directly and skipping the if elses

emhane (changes requested):

PR #2886, opened 2025-09-24
chore(node): fix missing semicolons in return statements

ilikecodewithmyhand:
Added missing semicolons to return statements across multiple files. Found some instances where return statements were missing trailing semicolons, causing potential formatting inconsistencies.


emhane:
this is redundant

emhane:
please check out the existing issues! thanks!

PR #2900, opened 2025-10-01
chore: Fix typos across multiple files

kerimmoresh:
Hello Team, I fixed some typos and inaccuracies across multiple files.


bin/host/src/interop/cfg.rs
```diff
@@ -109,7 +109,7 @@ pub enum InteropHostError {
     /// Task failed to execute to completion.
     #[error("Join error: {0}")]
     ExecutionError(#[from] tokio::task::JoinError),
-    /// A RPC error.
+    /// An RPC error.
```
emhane, line 112: pls revert

emhane (changes requested):

bin/host/src/interop/cfg.rs
kerimmoresh: @emhane done, please check, thanks!

PR #2910, opened 2025-10-04
*(tests)* Replace context.TODO() with bounded context in reorg tests

viktorking7:
## Summary

Replace `context.TODO()` with a bounded context using timeout in supervisor reorg test utilities to prevent test hanging and improve reliability.

## Changes

- Replace `context.TODO()` with `context.WithTimeout()` (30s) in `StopL1CL()` method
- Add `time` import to support timeout functionality

## Why

Using `context.TODO()` in test utilities can cause tests to hang indefinitely if the Kurtosis API call stalls, making test failures hard to diagnose and debug. A bounded context with timeout provides better test reliability and clearer failure modes.

### DIFF
--- a/bin/host/src/interop/cfg.rs
+++ b/bin/host/src/interop/cfg.rs
@@ -72,11 +72,6 @@ pub struct InteropHost {
         env
     )]
     pub l1_beacon_address: Option<String>,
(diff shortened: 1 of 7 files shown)


viktorking7:
@emhane ready to merge?

PR #2927, opened 2025-10-09
rpc(node): remove unsupported safeHeadAtL1Block endpoint

sashaodessa:
### Summary
Remove the unsupported `safeHeadAtL1Block` RPC method from the public interface and its no-op implementation. The endpoint always returned `MethodNotFound` and is no longer required post-interop.

### Changes
- Delete method declaration from `crates/node/rpc/src/jsonrpsee.rs`.
- Remove implementation from `crates/node/rpc/src/rollup.rs`.

### Rationale
- Reduces API surface and avoids false expectations for clients.
- No runtime behavior change: the endpoint previously returned `MethodNotFound`.


theochap:
We should add support for the `safeDB` at some point in the near future so we'll need to implement that endpoint. Closing this PR

PR #2966, opened 2025-10-23
feat(jovian): check that single batches only contain transactions. import latest test suite from monorepo

theochap:
## Description

This PR updates the derivation pipeline to ensure that single batches at jovian activation must be empty. It also updates the monorepo's test suite (action and acceptance).

### DIFF
--- a/crates/protocol/protocol/src/batch/single.rs
+++ b/crates/protocol/protocol/src/batch/single.rs
@@ -156,12 +156,14 @@ impl SingleBatch {
             }
         }
 
-        // If this is the first block in the interop hardfork, and the batch contains any
+        // If this is the first block in the jovian or interop hardfork, and the batch contains any
         // transactions, it must be dropped.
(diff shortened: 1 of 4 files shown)


crates/protocol/protocol/src/batch/single.rs
```diff
@@ -156,12 +156,14 @@ impl SingleBatch {
             }
         }
 
-        // If this is the first block in the interop hardfork, and the batch contains any
+        // If this is the first block in the jovian or interop hardfork, and the batch contains any
         // transactions, it must be dropped.
-        if cfg.is_first_interop_block(self.timestamp) && !self.transactions.is_empty() {
+        if (cfg.is_first_jovian_block(self.timestamp) || cfg.is_first_interop_block(self.timestamp)) &&
+            !self.transactions.is_empty()
```
theochap: Corresponding PR from the monorepo https://github.com/ethereum-optimism/optimism/pull/17984

PR #2967, opened 2025-10-24
chore: reduce L1WatcherRPC poll interval from 13s to 5s

### LINKED ISSUE #2962: Lower probability of L1WatcherRPC Missing Blocks
Right now the poller for L1WatcherRPC is configured to poll the provider [every 13 seconds](https://github.com/op-rs/kona/blob/5186e7e5fc19763ac6b7babfdd6b601776185276/crates/node/service/src/actors/l1_watcher_rpc.rs#L173), which is problematic because it is longer than L1 blocktime. 

This should be lowered to 5 seconds to make it much less probable that the poller misses blocks. 5s makes it so that multiple attempts are made per block in the happy path, and it builds in a buffer for request retries, network delays, etc.

Note: this is a bandaid until the larger #2954 task can be done which will guarantee no blocks are missed.

thunggis:
Reduces the L1WatcherRPC polling interval from 13 seconds to 5 seconds to significantly lower the probability of missing L1 blocks.

Closes #2962

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -98,7 +98,7 @@ jobs:
           mv "${{ steps.meta.outputs.bake-file }}" "${{ runner.temp }}/bake-meta.json"
 
       - name: Upload meta bake definition
-        uses: actions/upload-artifact@v4
+        uses: actions/upload-artifact@v5
         with:
           name: bake-meta
           path: ${{ runner.temp }}/bake-meta.json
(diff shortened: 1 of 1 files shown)


crates/node/service/src/actors/l1_watcher_rpc.rs
```diff
@@ -170,7 +170,7 @@ impl NodeActor for L1WatcherRpc {
         let mut head_stream = BlockStream::new(
             &self.state.l1_provider,
             BlockNumberOrTag::Latest,
-            Duration::from_secs(13),
+            Duration::from_secs(5),
```
theochap: Can you use 4 secs instead of 5? So that 12 secs (one l1 block) is a multiple of the duration?

theochap (approved):

PR #2988, opened 2025-10-28
feat(node/rollup-boost): Integrate rollup-boost library for native flashblocks

teddyknox:
- Introduces optional rollup-boost integration in the node’s engine path.
  - Adds `RollupBoostServerLike` abstraction and a concrete adapter to the rollup-boost Engine API server, with conversions to bridge `op-alloy` 0.21.0 types used by our public API and `op-alloy` 0.20.0 types required by `reth` v1.8.x.
  - When configured, `EngineClient` routes `newPayloadV3/V4`, `forkchoiceUpdatedV3`, and `getPayloadV3/V4` through rollup-boost; otherwise, it falls back to the L2 engine provider.

- Adds a compatibility layer for `op-alloy` type versions.
  - New `compat` module provides lossless conversions between `op-alloy` 0.20.0 and 0.21.0 envelopes and payloads to keep public APIs on 0.21.0 while interoperating with `reth` and rollup-boost dependencies.

- Wires rollup-boost through the service and RPC layers.
  - `RollupNodeBuilder` creates and injects a rollup-boost server instance, plumbs an `execution_mode` handle, and passes it to engine and RPC actors.
  - Admin RPC adds `setExecutionMode` and `getExecutionMode` to change/inspect execution mode at runtime and updates a gauge metric.

- Extends the CLI with rollup-boost configuration.
  - New flags to configure execution mode, builder/L2 endpoints and JWTs, selection policy, external state root, and Flashblocks settings.
  - Validation ensures required builder settings are present when execution mode is not disabled.

- Dependency and crate updates.
  - Adds `rollup-boost` and `parking_lot` where needed.
  - Updates lockfile entries for `reth`, `revm`, and related transitive dependencies to align with rollup-boost.

### Notable Changes

- Engine client:
  - Optional rollup-boost path for `new_payload_v3/v4`, `forkchoice_updated_v3`, `get_payload_v3/v4`.
- New modules:
  - `crates/node/engine/src/compat.rs` (op-alloy 0.20/0.21 conversions).
  - `crates/node/engine/src/rollup_boost.rs` (rollup-boost adapter + trait).
- Service:
  - Rollup-boost server creation and injection; execution mode handle exposed to RPC.
- RPC:
  - Admin methods `setExecutionMode`/`getExecutionMode` and associated state/metrics.

### Operational Notes

- By default, execution mode is enabled at the CLI level. If you do not intend to run with a builder, set `--rollup-boost.execution-mode=disabled`.
- When execution mode is not disabled, you must provide `--builder-engine-rpc` and `--builder-engine-jwt-secret` (and any Flashblocks settings if enabled).

### Risk/Compatibility

- Behavior change at startup: missing builder configuration with enabled execution mode will cause early validation failures. Disable the feature or provide builder settings.
- Public RPC surface grows with two admin methods; no changes to existing method semantics.
- Type conversions are structural and lossless between `op-alloy` 0.20.0 and 0.21.0.

### DIFF
--- a/.github/workflows/proof.yaml
+++ b/.github/workflows/proof.yaml
@@ -95,6 +95,10 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v5
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          large-packages: false
       - uses: ./.github/actions/setup
         with:
           components: llvm-tools-preview
--- a/.github/workflows/rust_ci.yaml
+++ b/.github/workflows/rust_ci.yaml
@@ -42,6 +42,10 @@ jobs:
         uses: actions/checkout@v5
         with:
           submodules: true
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          large-packages: false
       - uses: ./.github/actions/setup
         with:
           channel: stable
@@ -76,6 +80,10 @@ jobs:
         uses: actions/checkout@v5
         with:
           submodules: true
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          large-packages: false
       - uses: ./.github/actions/setup
         with:
           channel: nightly
@@ -103,6 +111,10 @@ jobs:
         uses: actions/checkout@v5
         with:
           submodules: true
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          large-packages: false
       - uses: ./.github/actions/setup
       - name: build benches
         run: just benches
@@ -119,6 +131,10 @@ jobs:
         uses: actions/checkout@v5
         with:
           submodules: true
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          large-packages: false
       - uses: ./.github/actions/setup
         with:
           channel: nightly
@@ -135,6 +151,10 @@ jobs:
         uses: actions/checkout@v5
         with:
           submodules: true
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          large-packages: false
       - uses: ./.github/actions/setup
       - name: doclint
         run: just lint-docs
@@ -147,6 +167,10 @@ jobs:
         uses: actions/checkout@v5
         with:
           submodules: true
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          large-packages: false
       - uses: ./.github/actions/setup
       - name: doctest
         run: just test-docs
@@ -166,6 +190,10 @@ jobs:
         uses: actions/checkout@v5
         with:
           submodules: true
+      - name: Free Disk Space (Ubuntu)
+        uses: jlumbroso/free-disk-space@main
+        with:
+          large-packages: false
       - uses: ./.github/actions/setup
       - name: Install clang
         run: sudo apt install clang libclang-dev


bin/node/src/commands/node.rs
```diff
@@ -132,15 +195,42 @@ pub struct NodeCommand {
...
-            l1_config_file: None,
-            node_mode: NodeMode::Validator,
+            l2_client_timeout: 1000,
+
+            builder_engine_rpc: Some(Url::parse("http://localhost:8551").unwrap()),
+            builder_engine_jwt_secret: None,
+            builder_client_timeout: 1000,
+            // builder_health_check_interval: 60,
+            // builder_max_unsafe_interval: 10,
+            rollup_boost_execution_mode: ExecutionMode::Enabled,
```
theochap: This default value doesn't match with the above default one (by default, in the CLI, execution mode is disabled). Just checking it is expected

bin/node/src/commands/node.rs
```diff
@@ -114,10 +128,59 @@ pub struct NodeCommand {
...
+    pub rollup_boost_external_state_root: bool,
+    /// Allow all engine API calls to builder even when marked as unhealthy
+    /// This is default true assuming no builder CL set up
+    #[arg(
+        long,
+        visible_alias = "rollup-boost.ignore-unhealthy-builders",
+        env,
+        default_value = "false"
+    )]
+    pub rollup_boost_ignore_unhealthy_builders: bool,
```
theochap: I am wondering if here we shouldn't just reuse the `RollupBoostArgs` from the rollup-boost cli directly. This will make kona more resistant to breakages. One thing is that it seems that by default `ExecutionMode` will be enabled, but we can wrap the subcommand around an option to be sure that rollup-boost is disabled if not specified

crates/node/rpc/src/admin.rs
```diff
@@ -47,6 +54,42 @@ pub struct AdminRpc {
...
+    /// - `network_sender`: The sender to the network actor.
+    /// - `execution_mode`: The rollup boost execution mode.
+    ///
+    /// # Returns
+    ///
+    /// A new [`AdminRpc`] instance.
+    pub fn new(
+        sequencer_sender: Option<SequencerQuerySender>,
+        network_sender: NetworkAdminQuerySender,
+        rollup_boost_execution_mode: Arc<Mutex<ExecutionMode>>,
```
theochap: Let's try not to use a `Mutex` here. We could use channels instead

bin/node/src/commands/node.rs
teddyknox: Good callout, this should be disabled.

bin/node/src/commands/node.rs
teddyknox: Updated in 5d184022a870033978372eb98307d21ca3b9976d.

theochap:
Reworked it in #3024 to make CI pass

crates/node/engine/src/client.rs
```diff
@@ -83,39 +76,153 @@ pub struct EngineClient {
...
+            rollup_boost::PayloadSource::L2,
+        )?;
+        let builder_client = rollup_boost::RpcClient::new(
+            http::Uri::from_str(self.builder.to_string().as_str())?,
+            self.builder_jwt,
+            self.builder_timeout,
+            rollup_boost::PayloadSource::Builder,
+        )?;
+
+        let rollup_boost_server: Box<dyn RollupBoostServerLike + Send + Sync + 'static> =
```
theochap: I don't think we really want to have a `Box<dyn ...>` here. In a follow-up PR, I'll use generics instead.

crates/node/engine/Cargo.toml
```diff
@@ -54,6 +55,11 @@ serde_json.workspace = true
 # metrics
 metrics = { workspace = true, optional = true }
 
+# rollup boost
+rollup-boost.workspace = true
+http.workspace = true
+parking_lot.workspace = true
```
theochap: We should remove `parking_lot` and `http` by upstreaming changes to `rollup-boost`

crates/node/engine/README.md
```diff
@@ -31,37 +31,6 @@ The engine implements a task-driven architecture where forkchoice synchronizatio
...
-
-// Create task queue watchers
-let (state_sender, _) = tokio::sync::watch::channel(state);
-let (queue_sender, _) = tokio::sync::watch::channel(0);
-
-// Create the engine
-let engine = Engine::new(state, state_sender, queue_sender);
-# Ok(())
-# }
-```
```
theochap: I took the liberty to remove that example since it was quite outdated and not that helpful to understand how the `Engine` works

bin/node/src/flags/engine/flashblocks.rs
```diff
@@ -0,0 +1,138 @@
...
+    /// Websocket connection configuration
+    #[command(flatten)]
+    pub flashblocks_ws_config: FlashblocksWebsocketFlags,
+}
+
+impl Default for FlashblocksFlags {
+    fn default() -> Self {
+        Self {
+            flashblocks: false,
+            flashblocks_builder_url: Url::parse("ws://127.0.0.1:1111").unwrap(),
```
einar-oplabs: Consider using named constants instead of hardcoded values and a value of `locahost` not to assume ip4.

einar-oplabs (commented):
I had a look.  Ideally I recommend a PR this size should be split up.

crates/node/engine/src/rollup_boost.rs
```diff
@@ -0,0 +1,205 @@
...
+};
+use std::{fmt::Debug, sync::Arc};
+use thiserror::Error;
+
+use rollup_boost::BlockSelectionPolicy;
+use url::Url;
+
+/// Configuration for the rollup-boost server.
+#[derive(Clone, Debug)]
+pub struct RollupBoostArgs {
```
op-will:
```suggestion
pub struct RollupBoostServerArgs {
```

crates/node/engine/src/rollup_boost.rs
op-will: This file hosts structs for servers and clients and doesn't always distinguish between them. If we can update names to be indicative, it'll save context switching and misunderstandings at dev/review time.

bin/node/src/flags/engine/flashblocks.rs
theochap: [9e114ca](https://github.com/op-rs/kona/pull/2988/commits/9e114cad13812df17f762d773ebef107b50e0841)

PR #2997, opened 2025-10-29
perf(ci/tests): parallelize action-tests

### LINKED ISSUE #2983: perf(ci/action-tests): parallelize action tests execution over several runners in CI
## Description

Full execution of action tests in CI takes about an hour. We have to bring this number down as this has become the main bottleneck when merging PRs for the rust stack.

We can use https://github.com/hashicorp-forge/go-test-split-action to split the `just action-tests-simple` recipe (which is a simple `go test` under the hood) over 3-4 CI runners.

einar-oplabs:
Resolves #2983 

Warnings about missing cache keys are expect outside of `main`.

### DIFF
--- a/.github/workflows/proof.yaml
+++ b/.github/workflows/proof.yaml
@@ -8,17 +8,23 @@ on:
 env:
   CARGO_TERM_COLOR: always
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   action-tests:
     name: FPP actions (${{ matrix.kind }})
     runs-on: ubuntu-latest
     timeout-minutes: 90
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix:
         kind: ["single", "interop"]
+        parallel: [4]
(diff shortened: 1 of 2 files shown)


einar-oplabs:
Status: There has been one [successful run](https://github.com/op-rs/kona/actions/runs/18973158963). The warnings about caching is because cache is only activated on `main` and can be ignored.

After upgrading the dependencies in 69ee2501 it yields new warnings, see below, and uses the naive strategy.  I have reverted the dependencies for now, so this can get merged.  We can deal with upgrading the actions another time.

theochap (changes requested):
Can you use the OSS commit convention for your PRs (https://www.conventionalcommits.org/en/v1.0.0/)? For instance https://github.com/op-rs/kona/pull/2994
Also, can you erase all your commits and have one commit that contains all your changes? You can reset to main, recommit and force push

theochap:
One last thing before merging: why is [Proof / FPP actions (${{ matrix.kind }}) sequential (pull_request)](https://github.com/op-rs/kona/actions/runs/19038352356/job/54368536968?pr=2997) skipped?

PR #3001, opened 2025-10-31
fix: remove is_zero check from blob_data as 4844 allows zeros blob submitted to L1

bxue-l2:
Problem

Kona blob source checks if the entire blob is zero, see [here](https://github.com/op-rs/kona/blob/15cf172fa4086ad81da8a46704d67fea31b5cbf6/crates/protocol/derive/src/sources/blob_data.rs#L159).

If it is zero, `BlobDecodingError::MissingData` is converted into a critical error [here](https://github.com/op-rs/kona/blob/15cf172fa4086ad81da8a46704d67fea31b5cbf6/crates/protocol/derive/src/errors/sources.rs#L48).

But it is fine for a 4844 blob to have all 0 field elements, its kzg commitment would be a point at infinity.

The check isn't enforced by [go-implementation](https://github.com/ethereum-optimism/optimism/blob/247b3992554fa6868c5718945f09d2bb85adfb2b/op-node/rollup/derive/blob_data_source.go#L154).

For an all-zero blob, it satisfies the encoding scheme described [here](https://specs.optimism.io/protocol/ecotone/derivation.html?highlight=encoding#blob-encoding) and [here](https://specs.optimism.io/protocol/ecotone/derivation.html?highlight=encoding#ecotone-blob-retrieval), as the length bytes are 0. By the time of decoding, all the data is [truncated](https://github.com/op-rs/kona/blob/e41d8c20c5badf2fc960ae9e6b4029fab4160b96/crates/protocol/derive/src/sources/blob_data.rs#L92). Then the empty array would be safely dropped by frame_queue in the [parse_frames](https://github.com/op-rs/kona/blob/8bba740643d5542f49c71bd015fa057b8eb1f3a4/crates/protocol/protocol/src/frame.rs#L231) function, which returns Ok when there is error.

### DIFF
--- a/crates/protocol/derive/src/sources/blob_data.rs
+++ b/crates/protocol/derive/src/sources/blob_data.rs
@@ -197,7 +197,8 @@ mod tests {
     fn test_fill_zero_blob() {
         let mut blob_data = BlobData::default();
         let blobs = vec![Box::new(Blob::ZERO)];
-        assert_eq!(blob_data.fill(&blobs, 0), Err(BlobDecodingError::MissingData));
+        // consider a blob made entirely of zero bytes a regular blob
+        assert_eq!(blob_data.fill(&blobs, 0), Ok(true));
     }
 
     #[test]


theochap:
Did you see this error happening on a devnet?

bxue-l2:
> Did you see this error happening on a devnet?

No, I was analyzing the codepath and found this. But I can try later tody.

bxue-l2:
@theochap I can produce it. Here are a list of screenshot. I am running a kurtosis devnet using an older kurtosis package that supported to prague, see [here](https://github.com/Layr-Labs/hokulea/tree/master?tab=readme-ov-file#local-manual-testing). The kona version for the test is kona-client/v1.1.6.

I submitted a zero formatted blob, which is generated from the command 
```
dd if=/dev/zero of=zero.blob bs=131072 count=1
```
Then submission takes place at l1 block number 121
Screenshot 2025-10-31 at 10 05 49 AM


I run the derivation procedures on kurtosis. Until l1 block 120 for L2 claimed block 318, the system still operates as fine.
Screenshot 2025-10-31 at 10 19 55 AM


Then I derive for L2 claimed block at 400 blocks which would cover the l1 zero blob submitted earlier. We encountered the error
Screenshot 2025-10-31 at 10 19 09 AM

bxue-l2:
With the fix, I am able to derive it

Screenshot 2025-10-31 at 10 45 43 AM

theochap:
Awesome, thanks a lot for taking care of that. Going to approve

theochap:
Awesome, thanks! Would you mind rebasing on main? The acceptance tests are failing because of a new mise update. I fixed it in the last commit to main

PR #3019, opened 2025-11-06
fix(cli): log actual bound port when KONA_METRICS_PORT=0

### LINKED ISSUE #2987: chore(cli) Fix metrics logging at startup
There are two issues with metrics logging:

1. We initialize metrics before logging: https://github.com/op-rs/kona/blob/1b136596d612de0ba5d17aee49df4fd52873a3fd/bin/node/src/cli.rs#L55-L64
2. If we reverse the order, we log the metrics port as `0` if the `KONA_METRICS_PORT` env var is set to `0`: https://github.com/op-rs/kona/blob/1b136596d612de0ba5d17aee49df4fd52873a3fd/crates/utilities/cli/src/prometheus.rs#L14-L34

For 2, if kona is started with the `KONA_METRICS_PORT=0` env var/value, the local metrics port will be a random available port selected by the os, which is desirable in a number of cases.

The issue is that there is no programatic visibility into the port that was selected, and the metrics log ends up being "Serving metrics at: http://127.0.0.1:0". This causes issues with downstream processes like tests that attempt to get the metrics port (right now by parsing from logs).

This issue is to 
1. fix the initialization order so metrics initialization logging gets printed
2. log the port that metrics is actually running on rather than what was provided (i.e. fix the `0` case)

crStiv:
Fixes #2987

When `KONA_METRICS_PORT=0` is set, the OS assigns a random available port, but the metrics log was showing `http://127.0.0.1:0` instead of the actual bound port.

This fix creates a temporary `TcpListener` to get the OS-assigned port when `metrics_port == 0`, then uses that address for both Prometheus initialization and logging.

**Changes:**
- Detect when `metrics_port == 0` and bind a temporary listener to get the actual port
- Log the actual bound address instead of the configured address


crates/utilities/cli/src/prometheus.rs
```diff
@@ -3,16 +3,28 @@
...
 /// Start a Prometheus metrics server on the given port.
 pub fn init_prometheus_server(addr: IpAddr, metrics_port: u16) -> Result<(), BuildError> {
-    let prometheus_addr = SocketAddr::from((addr, metrics_port));
-    let builder = PrometheusBuilder::new().with_http_listener(prometheus_addr);
+    // If port is 0, we need to bind first to get the actual port assigned by the OS
+    let actual_addr = if metrics_port == 0 {
+        // Create a temporary listener to get the OS-assigned port
+        let listener =
+            TcpListener::bind((addr, 0)).expect("Failed to bind listener for metrics server");
+        let bound_addr = listener.local_addr().expect("Failed to get local address from listener");
```
einar-oplabs: Would it be possible and desirable  to report a `BuildError` instead of crashing?

PR #3035, opened 2025-11-13
feat(registry): Support custom embedded chain config

Inphi:
Add the ability to customize the embedded chain configs provided by kona-registry.
Custom chain configs are merged into the superchain configs without conflicts. Any conflicts found will result in a build failure. The merge operation is idempotent. The resulting configs generated in the etc directory have a minimal diff, allowing users to easily verify the changes introduced.

Breaking changes:
- A couple kona-registry types were moved to kona-genesis so that the kona-registry build script could use the types to parse custom chain configs.

### DIFF
--- a/.github/workflows/proof.yaml
+++ b/.github/workflows/proof.yaml
@@ -51,17 +51,6 @@ jobs:
         with:
           gotestsum_version: 1.12.1
 
-      - name: Download JUnit Summary from Previous Workflow
-        id: download-artifact
-        uses: dawidd6/action-download-artifact@v11
-        with:
-          workflow_conclusion: success
-          name: junit-test-summary
-          if_no_artifact_found: warn
-          use_unzip: true
-          # todo: uncomment this before merging to main
-          # branch: main
-
       - name: Build Action Tests
         run: |
           source <(cargo llvm-cov show-env --export-prefix)
@@ -74,19 +63,11 @@ jobs:
         with:
           total: ${{ matrix.parallel }}
           index: ${{ matrix.index }}
-          junit-summary: ./junit-test-summary.xml
           working-directory: ./tests/optimism/op-e2e/actions/proofs
 
       - name: Run Action Tests
         run: |
           just "action-tests-${{ matrix.kind }}-run" "${{ steps.test_split.outputs.run }}"
-
-      - name: Upload test artifacts
-        uses: actions/upload-artifact@v4
-        with:
-          name: junit-test-summary-${{ matrix.index }}
-          path: tests/optimism/op-e2e/actions/proofs/node-summary.xml
-          retention-days: 1
 
       - name: Upload coverage to codecov.io
         uses: codecov/codecov-action@v5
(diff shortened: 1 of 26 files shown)


crates/protocol/genesis/src/superchain/chain_list.rs
```diff
@@ -62,11 +62,21 @@ pub struct Chain {
     pub explorers: Vec<String>,
     /// The Superchain Level.
     pub superchain_level: u64,
+    /// Governed by Optimism flag.
+    #[cfg_attr(feature = "tabled", tabled(skip))]
```
Inphi: The added fields in this struct are already in the superchain-registry configs in the etc directory. I updated the struct so that parsing the superchain-registry configs do not strip them out.

crates/protocol/registry/build.rs
```diff
@@ -73,4 +82,234 @@ fn main() {
...
+}
+
+fn merge_custom_configs() {
+    let kona_custom_configs =
+        std::env::var("KONA_CUSTOM_CONFIGS").unwrap_or_else(|_| "false".to_string()) == "true";
+    println!("cargo:rerun-if-env-changed=KONA_CUSTOM_CONFIGS");
+    println!("cargo:rerun-if-env-changed=KONA_CUSTOM_CONFIGS_TEST");
+
+    // if we're running tests, bust the cache if the base etc configs are updated. This ensures that the
+    // test build can be repeeated after modifying the base configs
```
theochap:
```suggestion
    // test build can be repeated after modifying the base configs
```

crates/protocol/registry/build.rs
```diff
@@ -73,4 +82,234 @@ fn main() {
...
+    if std::env::var("KONA_CUSTOM_CONFIGS_TEST").unwrap_or_else(|_| "false".to_string()) == "true" {
```
theochap:
```suggestion
    if std::env::var("KONA_CUSTOM_CONFIGS_TEST") == Some("true") {
```

crates/protocol/registry/build.rs
```diff
@@ -73,4 +82,234 @@ fn main() {
...
+    println!("cargo:rerun-if-changed={}", custom_configs_path.display());
+
+    let target_chain_list = Path::new("etc/chainList.json");
+    let target_superchains = Path::new("etc/configs.json");
+
+    validate_chain_configs(&custom_chain_list_path, &custom_configs_path);
+
+    merge_chain_list(&custom_chain_list_path, target_chain_list);
+    merge_superchain_configs(&custom_configs_path, target_superchains);
+    validate_chain_configs(target_chain_list, target_superchains);
```
theochap: Nit: Do we really want to merge both both custom and SCR configs? Maybe it is much more of a hassle to store both in separate files, but I'd say that this makes it easier to debug/know which dependency changed (if it is the scr or the custom configs)

Inphi: Either we merge the config files or merge them at runtime/lazy_static time. Could be worth doing the latter to preserve the sanctity of the superchain-registry configuration. But yeah, it'll be difficult to know what custom configs got included if we do this at runtime.
Perhaps we could introduce a `pub fn debug_print_configs()` function to `kona-registry` that dumps the configs. So that the user can verify what got added in.

theochap: Sounds good, would you mind adding that `debug_print_configs` method so that we can dump the custom configs if necessary?

Inphi: ahh this can't work because we may not have an std::env to read from in a no_std environment. And we need std::env to tell us where to locate the custom file config. Merging the configs at build time is probably the best approach.

Inphi:
@theochap should be fixed now. could you please kickoff the ci workflow

PR #3041, opened 2025-11-14
chore(ci): rework test justfile to try and improve CI times

theochap:
## Description

Build op-reth from workspace root instead of optimism crate. This should trigger CI cache


einar-oplabs (approved):
LGTM

op-will (approved):

PR #3046, opened 2025-11-15
fix: update channel rlp length tracking

davidjsonn:
ChannelOut wasn't bumping its rlp_length counter after accepting a batch, so the per-channel RLP limit was enforced on stale numbers and input_bytes() lied about how much data went in. Now the counter increments as soon as the batch is written, so both the limit check and the reported stats match reality.


op-will (dismissed):
👍

PR #3057, opened 2025-11-19
refactor: remove redundant Ok(...?) in list_walker fetch_leaves

GarmashAlex:
The match arms for TrieNode::Extension in fetch_leaves returned Ok(...?) which re-wrapped a Result whose error was already being propagated with ?. Returning the Result directly simplifies the code while preserving identical behavior and error types.


einar-oplabs (approved):
Looks like a safe simplification.

PR #3058, opened 2025-11-19
refactor(protocol/protocol): switch to nested structs

einar-oplabs:
The `L1BlockInfo___` structs contains overlapping fields.  This branch factors out the non-deprecated fields e.g. `L1BlockInfoBedrockBase` and embeds this as a field in `L1BlockInfoBedrock`:

```
pub struct L1BlockInfoBedrock {
    #[serde(flatten)]
    base: L1BlockInfoBedrockBase,
    /// The fee overhead for L1 data. Deprecated in ecotone.
    pub l1_fee_overhead: U256,
    /// The fee scalar for L1 data. Deprecated in ecotone.
    pub l1_fee_scalar: U256,
}
```
The purpose is reuse (think OOP inheritance) instead of repetition.  As a side-effect this increases encapsulation.

It establishes a partial order and a chain of fully embedded structs:

    L1BlockInfoBedrockBase < L1BlockInfoEcotoneBase < L1BlockInfoIsthmus < L1BlockInfoJovian

Further

    L1BlockInfoBedrockBase < L1BlockInfoBedrock

and

    L1BlockInfoEcotoneBase < L1BlockInfoEcotone

This is deemed necessary to get around deprecated fields in `L1BlockInfoBedrock` and `L1BlockInfoEcotone`.

To hide the implementation details, constructors have been added and destructuring is discouraged.

There is no single way to do this in Rust, but this is one way. A similar way is used [`op-alloy`](https://github.com/alloy-rs/op-alloy/blob/main/crates/rpc-types/src/transaction.rs).


crates/protocol/protocol/src/info/jovian.rs
```diff
@@ -28,32 +33,89 @@ use crate::DecodeError;
...
+    }
+
+    fn blob_base_fee_scalar(&self) -> u32 {
+        self.base().blob_base_fee_scalar()
+    }
+
+    fn base_fee_scalar(&self) -> u32 {
+        self.base().base_fee_scalar()
+    }
+}
```
theochap: Can we have a blanket implementation of the `L1BlockInfoEcotoneBaseFields` for every type that implements `HasBaseField<L1BlockInfoEcotone>`? That way you only need to specify `impl HasBaseField<L1BlockInfoEcotone>` for L1BlockInfoJovian` and it automatically derive all those methods?

einar-oplabs: If what you have in mind is something like:

```rust
impl<S,T> MyTrait for T
where
	T: HasBaseField<S>
	S: MyTrait
{}
```
then unfortunately no.
I was trying to do exactly that, but as far as I understand this wont work, because `S` is not constrained (It would be if first line was `MyTrait<S>`).  It can be done for a concrete value of `S`, e.g. `L1BlockInfoEcotoneBase`, but there are only two traits, namely `L1BlockInfoBedrockBase` and `L1BlockInfoEcotoneBase`, that are used twice, so I ended up reasoning that it's simply more clear to write it explicitly. 

There are some discussion in the Rust community under the term "delegation" (to subfields that is), but there are no neat ways to do it as of now. A recursive trait impl would have been neater.

If you prefer the blanket impl, I am happy to use it. I don't have a strong preference either that.

crates/protocol/protocol/src/info/ecotone.rs
```diff
@@ -97,23 +165,105 @@ impl L1BlockInfoEcotone {
...
+        time: u64,
+        base_fee: u64,
+        block_hash: B256,
+        sequence_number: u64,
+        batcher_address: Address,
+        blob_base_fee: u128,
+        blob_base_fee_scalar: u32,
+        base_fee_scalar: u32,
+        empty_scalars: bool,
+        l1_fee_overhead: U256,
```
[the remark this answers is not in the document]
einar-oplabs: There is in fact: the [new type pattern](https://doc.rust-lang.org/rust-by-example/generics/new_types.html), [longer read](https://www.worthe-it.co.za/blog/2020-10-31-newtype-pattern-in-rust.html), that allows you to make zero-cost, mutually-incompatible types out of a base type, such as `u64`. We could do a refactor, but I would say it's out of scope for this PR.

But you are right, it is actually a weakness in the constructor approach, at least most editors should show the parameter names.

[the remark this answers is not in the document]
theochap: Agreed with @einar-oplabs

crates/protocol/protocol/src/info/jovian.rs
theochap: Hm I see. The only thing that worries me a bit is the added number of lines in this PR when it is supposed to refactor a part of the codebase to make simpler. The fact that we're adding 800 lines of code more makes me think that there may be something we're doing wrong here.

I trust your gut, if you feel there is no way of doing it in a less complex/more straightforward way. I'd say it is worth trying to go the recursive trait way if this helps reduce the additional number of LOC but up to you

theochap (approved):
Approving to unblock. Feel free to go ahead and merge if my comment's not relevant

crates/protocol/protocol/src/info/ecotone.rs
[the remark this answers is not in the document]
op-will: @einar-oplabs do you mind creating an issue for it after merging? If we spell out the issue and what we're hoping to achieve, maybe we'll get help from an open source contributor.

crates/protocol/protocol/src/info/jovian.rs
einar-oplabs: Besides now using the `ambassador` crate, I did some factoring and reuse of the encode/decode functions and now it is somewhat better:

> +1,146 -703

While I understand and have sympathy for defining each type in isolation, I think there is value and security in reusing encoding and decoding logic.

PR #3059, opened 2025-11-19
fix(protocol/derive): fix singular batch extraction error handling

sebastianst:
This is the same fix as https://github.com/ethereum-optimism/optimism/pull/18283 for Kona.

Fixed for the post-Holocene `BatchValidator` and old `BatchQueue`. Unit tests added.

The derivation rule was technically in the specs already, but they get clarified with https://github.com/ethereum-optimism/specs/pull/863.

### DIFF
--- a/crates/proof/std-fpvm/src/mips64/io.rs
+++ b/crates/proof/std-fpvm/src/mips64/io.rs
@@ -62,7 +62,7 @@ impl BasicKernelInterface for Mips64IO {
     fn exit(code: usize) -> ! {
         unsafe {
             let _ = syscall::syscall1(SyscallNumber::Exit as usize, code);
-            panic!()
+            panic!("exit syscall returned unexpectedly with code: {}", code)
         }
     }
 }


crates/protocol/derive/src/stages/batch/batch_stream.rs
```diff
@@ -85,19 +86,17 @@ where
...
         l1_origins: &[BlockInfo],
-    ) -> PipelineResult<()> {
+    ) -> Result<(), SpanBatchError> {
         if let Some(span) = self.span.take() {
-            self.buffer.extend(
-                span.get_singular_batches(l1_origins, parent).map_err(|e| {
-                    PipelineError::BadEncoding(PipelineEncodingError::from(e)).crit()
-                })?,
-            );
+            self.buffer.extend(span.get_singular_batches(l1_origins, parent)?);
```
theochap: That used to be treated as a critical error but now it has been downgraded to simple temporary errors (see the error handling you've added below). Is that right? I think you're approach looks more in-line with the spec but I am just making sure we're not missing anything here. It is also odd this has not been covered by tests.

crates/protocol/derive/src/stages/batch/batch_stream.rs
```diff
@@ -194,7 +193,17 @@ where
...
-        self.get_single_batch(parent, l1_origins).map(Batch::Single)
+        match self.get_single_batch(parent, l1_origins) {
+            Ok(Some(single_batch)) => Ok(Batch::Single(single_batch)),
+            Ok(None) => Err(PipelineError::NotEnoughData.temp()),
+            Err(e) => {
+                warn!(target: "batch_span", "Extracting singular batches from span batch failed: {}", e);
+                // If singular batch extraction fails, it should be handled the same as a
+                // dropped batch during span batch prefix checks.
+                self.flush();
+                Err(PipelineError::NotEnoughData.temp())
```
theochap: Errors coming from `try_hydrate_buffer` used to be systematically treated as critical errors and not they're always resolved to temporary errors. Just wanted to check that we want to do that, and if so we may want to link the relevant part of the spec here.

crates/protocol/protocol/src/batch/span.rs
```diff
@@ -1802,11 +1825,13 @@ mod tests {
...
-        let block =
+        let l1_a =
+            BlockInfo { number: 10, timestamp: 5, hash: l1_block_hash, ..Default::default() };
+        let l1_b =
             BlockInfo { number: 11, timestamp: 10, hash: l1_block_hash, ..Default::default() };
-        let second_block =
+        let l1_c =
             BlockInfo { number: 12, timestamp: 21, hash: l1_block_hash, ..Default::default() };
-        let l1_blocks = vec![block, second_block];
+        let l1_blocks = vec![l1_a, l1_b, l1_c];
```
theochap: Why did we need to add a new block here?

crates/protocol/derive/src/stages/batch/batch_stream.rs
```diff
@@ -399,6 +411,83 @@ mod test {
...
+        let mut stream = BatchStream::new(prev, config, provider);
+        let err = stream.next_batch(l2_safe_head, &l1_blocks).await.unwrap_err();
+
+        assert_eq!(err, PipelineError::NotEnoughData.temp());
+        assert!(stream.span.is_none());
+        assert_eq!(stream.span_buffer_size(), 0);
+
+        let logs = trace_store.get_by_level(tracing::Level::WARN);
+        assert_eq!(logs.len(), 1);
+        assert!(logs[0].contains("Extracting singular batches from span batch failed: Future batch L1 origin before safe head"));
```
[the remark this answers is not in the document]
sebastianst: I deliberately asserted on the full error chain, because I want to make sure we hit that particular code path.

crates/protocol/derive/src/stages/batch/batch_stream.rs
```diff
@@ -399,6 +411,83 @@ mod test {
...
+    }
```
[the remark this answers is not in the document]
sebastianst: I just followed the (admittedly not so great) pattern of the existing tests in this file. I would prefer not to scope-creep this PR and refactor testing inside `batch_stream.rs` in general.

crates/protocol/protocol/src/batch/span.rs
sebastianst: Because I also added a previously missing check that we even have the L1 blocks available for all L1 origins in the span batch:
https://github.com/op-rs/kona/blob/08102f1aae3b61d2ad0bb457f2ada3c922091b1a/crates/protocol/protocol/src/batch/span.rs?plain=1#L414-L425

So this is just fixing the test setup so that all L1 origins are available for the span batch in this test. Otherwise the test doesn't cover the code path that should be tested and the warning log assertion failed.

If we're hitting `unable to find L1 origin for batch` it _should_ actually be a logic error given the other new checks (above with warning "batch L1 origin is before safe head L1 origin") should already catch the preconditions for this and how the derivation pipeline handles L1 origins (it only removes those older than the safe head's L1 origin).

theochap (approved):

PR #3062, opened 2025-11-20
eliminate redundant type conversions in blob processing

Doryu001:
Optimizes blob field element processing by caching `i as usize` conversions instead of repeating them multiple times within the same expression.


theochap (commented):
Instead of doing these conversions, you could just set `FIELD_ELEMENTS_PER_BLOB` to be a usize.

PR #3076, opened 2025-11-21
feat(node/actor): Make SequencerActor generic over new tx gossip trait

### LINKED ISSUE #3068: feat(node/actor): Make SequencerActor generic over new tx gossip trait
This is a follow-up to #3045. See that PR for more context.

We want to inject `SequencerActor` dependencies to break dependence on implementation and facilitate testing. This pattern has been started  but not completed: https://github.com/op-rs/kona/blob/15dac0759081de6683f900b4c20226eb8fa3b806/crates/node/service/src/actors/sequencer/actor.rs#L60-L66

This issue is to remove dependence on `gossip_payload_tx` in favor of a generic trait being injected to provide tx gossip functions.
https://github.com/op-rs/kona/blob/15dac0759081de6683f900b4c20226eb8fa3b806/crates/node/service/src/actors/sequencer/actor.rs#L78

op-will:
Closes #3068

### DIFF
--- a/crates/node/service/src/service/standard/node.rs
+++ b/crates/node/service/src/service/standard/node.rs
@@ -47,8 +47,7 @@ pub struct RollupNode {
     pub(crate) sequencer_config: SequencerConfig,
 }
 
-#[async_trait]
-impl RollupNodeService for RollupNode {
+impl RollupNode {
     type DataAvailabilityWatcher = L1WatcherRpc;
 
     type AttributesBuilder = StatefulAttributesBuilder<AlloyChainProvider, AlloyL2ChainProvider>;


op-will:
> I would prefer a different prefix than `unsafe_` for unsafe head, as I associate it with the Rust keyword `unsafe`, but I will leave it up to you.
> 
> LGTM

@einar-oplabs good point. I'm ambivalent. I agree with you, but I don't like differing from the spec. The spec calls it unsafe head, so I think erring on the side of consistency with the spec is the way to go.

crates/node/service/src/actors/sequencer/actor.rs
```diff
@@ -86,26 +85,30 @@ where
     pub rollup_config: Arc<RollupConfig>,
     /// Watch channel to observe the unsafe head of the engine.
     pub unsafe_head_rx: watch::Receiver<L2BlockInfo>,
+    /// A client to asynchronously sign and gossip built payloads to the network actor.
+    pub unsafe_payload_gossip_client: G,
 }
 
-impl<AB, C, OS, BB> CancellableContext for SequencerActor<AB, C, OS, BB>
+impl<AB, BB, C, G, OS> CancellableContext for SequencerActor<AB, BB, C, G, OS>
```
theochap: Same comment on the naming for the generics

crates/node/service/src/actors/network/gossip.rs
```diff
@@ -0,0 +1,49 @@
...
+pub trait UnsafePayloadGossipClient: Send + Sync + Debug {
+    /// This is a fire-and-forget function that schedules the provided
+    /// [`OpExecutionPayloadEnvelope`] to be gossiped. The implementation should return as
+    /// quickly as possible and offers no guarantees that the payload actually was gossiped
+    /// successfully.
+    async fn schedule_execution_payload_gossip(
+        &self,
+        payload: OpExecutionPayloadEnvelope,
+    ) -> Result<(), UnsafePayloadGossipClientError>;
+}
```
theochap: Question: should we actually implement that as a two way method that waits for the payload to actually be gossiped to return? Maybe that's impossible because the network actor would deadlock if we tried to wait

crates/node/service/src/actors/sequencer/admin_api_impl.rs
```diff
@@ -1,16 +1,19 @@
...
+use crate::{
+    BlockBuildingClient, Conductor, OriginSelector, SequencerAdminQuery, UnsafePayloadGossipClient,
+};
 use alloy_primitives::B256;
 use kona_derive::AttributesBuilder;
 use kona_rpc::SequencerAdminAPIError;
 
 /// Handler for the Sequencer Admin API.
-impl<AB, C, OS, BB> SequencerActor<AB, C, OS, BB>
+impl<AB, BB, C, G, OS> SequencerActor<AB, BB, C, G, OS>
```
theochap: Same comment for naming

crates/node/service/src/actors/sequencer/actor.rs
```diff
@@ -381,12 +373,13 @@ where
 }
 
 #[async_trait]
-impl<AB, C, OS, BB> NodeActor for SequencerActor<AB, C, OS, BB>
+impl<AB, BB, C, G, OS> NodeActor for SequencerActor<AB, BB, C, G, OS>
```
theochap: Same comment for naming

crates/node/service/src/actors/sequencer/error.rs
```diff
@@ -20,4 +22,7 @@ pub enum SequencerActorError {
     /// An error occurred while attempting to build a payload.
     #[error(transparent)]
     BuildError(#[from] BuildTaskError),
+    /// An error occurred while attempting to schedule unsafe payload gossip.
+    #[error(transparent)]
```
theochap: Instead of using a transparent error here, we should probably wrap the error message into something more explicit like (`An error occurred while attempting to schedule unsafe payload gossip: {0}.`). I looked up the inner error type and it seems that the error raised is: `Error sending request: {0}.` which gives very little information on where/how the error occurred

crates/node/service/src/actors/network/gossip.rs
op-will: Might create an issue to evaluate if this is necessary / make it unbounded. Seems like it's had a capacity of 256 for a while without any noticeable issues.

crates/node/service/src/actors/sequencer/actor.rs
op-will: 03dae6c2cf8b814bf98f854fd9a8543147aebd46

PR #3083, opened 2025-11-24
Refine bootstore path handling to avoid redundant clones

justinbay:
reuse the existing bootstore path borrow instead of cloning twice, keep cloning only for the BootStoreFile::Custom conversion, add Path import to support the borrow-based formatting


justinbay:
please review

PR #3087, opened 2025-11-24
feat(node/actor): Expose unsafe head fetch functionality in BlockBuildingClient

### LINKED ISSUE #3069: feat(node/actor): Make SequencerActor generic over new unsafe head access trait
This is a follow-up to #3045. See that PR for more context.

We want to inject `SequencerActor` dependencies to break dependence on implementation and facilitate testing. This pattern has been started  but not completed: https://github.com/op-rs/kona/blob/15dac0759081de6683f900b4c20226eb8fa3b806/crates/node/service/src/actors/sequencer/actor.rs#L60-L66

This issue is to remove dependence on `unsafe_head_rx` in favor of a generic trait being injected to provide access to the unsafe head.
https://github.com/op-rs/kona/blob/15dac0759081de6683f900b4c20226eb8fa3b806/crates/node/service/src/actors/sequencer/actor.rs#L88

It could be tricky to decide the functionality this new trait should contain, given the different functionality of the current `watch::Receiver<L2BlockInfo>` implementation, so we should probably:
1. View all current uses to make sure the trait will support their needs
2. Err on the side of exposing a minimal interface that can evolve to meet later needs

op-will:
closes #3069

This also:
- Removes BlockBuildingClient from the output of EngineActor construction in favor of BlockBuildingClient being constructed in RollupNode
- Removes unnecessary waiting logic for unsafe head to update after initial reset within SequencerActor

### DIFF
--- a/crates/node/rpc/src/admin.rs
+++ b/crates/node/rpc/src/admin.rs
@@ -232,7 +232,7 @@ pub enum SequencerAdminAPIError {
     #[error("Error receiving response: {0}.")]
     ResponseError(String),
 
-    /// Error overriding leader.
+    /// Error stopping sequencer.
     #[error("Error stopping sequencer: {0}.")]
     StopError(String),
 
--- a/crates/node/service/src/actors/engine/actor.rs
+++ b/crates/node/service/src/actors/engine/actor.rs
@@ -65,7 +65,7 @@ pub struct SealRequest {
 pub struct EngineActor {
     /// A channel to receive [`OpAttributesWithParent`] from the derivation actor.
     attributes_rx: mpsc::Receiver<OpAttributesWithParent>,
-    /// The [`EngineActorState`] used to build the actor.
+    /// The [`EngineConfig`] used to build the actor.
     builder: EngineConfig,
     /// A channel to receive build requests.
     /// Upon successful processing of the provided attributes, a `PayloadId` will be sent via the
@@ -238,6 +238,15 @@ impl CancellableContext for EngineContext {
     }
 }
 
+struct SequencerChannels {
+    build_request_rx: Option<mpsc::Receiver<BuildRequest>>,
+    build_request_tx: Option<mpsc::Sender<BuildRequest>>,
+    seal_request_rx: Option<mpsc::Receiver<SealRequest>>,
+    seal_request_tx: Option<mpsc::Sender<SealRequest>>,
+    unsafe_head_rx: Option<watch::Receiver<L2BlockInfo>>,
+    unsafe_head_tx: Option<watch::Sender<L2BlockInfo>>,
+}
+
 impl EngineActor {
     /// Constructs a new [`EngineActor`] from the params.
     pub fn new(config: EngineConfig) -> (EngineInboundData, Self) {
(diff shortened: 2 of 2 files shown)


crates/node/service/src/actors/engine/actor.rs
```diff
@@ -65,46 +63,69 @@ pub struct SealRequest {
...
     /// A channel to receive [`OpAttributesWithParent`] from the derivation actor.
     attributes_rx: mpsc::Receiver<OpAttributesWithParent>,
-    /// A channel to receive [`OpExecutionPayloadEnvelope`] from the network actor.
-    unsafe_block_rx: mpsc::Receiver<OpExecutionPayloadEnvelope>,
-    /// Handler for inbound queries to the engine.
-    inbound_queries: mpsc::Receiver<EngineQueries>,
-    /// A channel to receive reset requests.
-    reset_request_rx: mpsc::Receiver<ResetRequest>,
+    /// The [`EngineActorState`] used to build the actor.
+    builder: EngineConfig,
```
theochap: Comment mismatch

crates/node/service/src/actors/engine/actor.rs
```diff
@@ -65,46 +63,69 @@ pub struct SealRequest {
...
+    pub reset_request_tx: mpsc::Sender<ResetRequest>,
+    /// A channel to send rollup boost admin queries to the engine actor.
+    pub rollup_boost_admin_query_tx: mpsc::Sender<RollupBoostAdminQuery>,
+    /// A channel to send rollup boost health queries to the engine actor.
+    pub rollup_boost_health_query_tx: mpsc::Sender<RollupBoostHealthQuery>,
+    /// A channel to use to send [`SealRequest`] payloads to the engine actor.
+    ///
+    /// This is `Some` when the node is in sequencer mode, and `None` when the node is in validator
+    /// mode.
+    pub seal_request_tx: Option<mpsc::Sender<SealRequest>>,
```
theochap: What is the logic behind reordering those fields? Just asking because I am not aware of any rules regarding field ordering. Just so that we have the same convention

crates/node/service/src/actors/engine/actor.rs
```diff
@@ -227,21 +240,35 @@ impl CancellableContext for EngineContext {
...
+                Some(build_request_tx),
+                Some(seal_request_rx),
+                Some(seal_request_tx),
+                Some(unsafe_head_rx),
+                Some(unsafe_head_tx),
+            )
         } else {
-            (None, None, None)
+            (None, None, None, None, None, None)
         };
```
theochap: nit: Wondering if we could not group all the sequencer channels together in one wrapper struct to avoid returning a 6-tuple here

crates/node/service/src/actors/sequencer/actor.rs
```diff
@@ -362,21 +357,13 @@ where
...
-        //
-        // We know that the reset has concluded when the unsafe head watch channel is updated.
-        if self.unsafe_head_rx.changed().await.is_err() {
-            error!(target: "sequencer", "Failed to receive unsafe head update after reset request");
-            return Err(SequencerActorError::ChannelClosed);
-        }
-
-        Ok(())
+            err.into()
+        })
```
theochap:
```suggestion
        self.block_building_client.reset_engine_forkchoice().await.inspect_err(|err| error!(target: "sequencer", ?err, "Failed to send reset request to engine"))?
```

crates/node/service/src/actors/sequencer/builder.rs
```diff
@@ -153,15 +140,16 @@ where
         Ok(SequencerActor {
             admin_api_rx: self.admin_api_rx.expect("admin_api_rx is required"),
             attributes_builder: self.attributes_builder.expect("attributes_builder is required"),
-            block_engine: self.block_engine.expect("block_engine is required"),
+            block_building_client: self
+                .block_building_client
+                .expect("block_building_client is required"),
```
theochap: Not related to this PR, but we should remove all these `.expect` here and use the type system to make sure this never panics

crates/node/service/src/service/node.rs
```diff
@@ -226,9 +219,18 @@ impl RollupNode {
...
+                        "build_request_tx is None in sequencer mode. This should never happen.",
+                    ),
+                    reset_request_tx: reset_request_tx.clone(),
+                    seal_request_tx: seal_request_tx.expect(
+                        "seal_request_tx is None in sequencer mode. This should never happen.",
+                    ),
+                    unsafe_head_rx: unsafe_head_rx.expect(
+                        "unsafe_head_rx is None in sequencer mode. This should never happen.",
+                    ),
+                };
```
theochap: Also here: we should remove all these `expect`s

theochap (approved):
A few nits. Let's open follow-up issues for the `.expect`s to remove. This is old code that needs to be cleaned up

crates/node/service/src/actors/sequencer/actor.rs
op-will:
```
|___________^ expected `Result<(), SequencerActorError>`, found `()`
```

Can do

```
self.block_building_client.reset_engine_forkchoice().await.inspect_err(|err| error!(target: "sequencer", ?err, "Failed to send reset request to engine"))?;

Ok(())
```

but the `map_err` approach does it all in one go.

crates/node/service/src/actors/engine/api.rs
```diff
@@ -53,10 +56,16 @@ pub struct QueuedBlockBuildingClient {
...
     pub reset_request_tx: mpsc::Sender<ResetRequest>,
+    /// A channel to receive the latest unsafe head [`L2BlockInfo`].
+    pub unsafe_head_rx: watch::Receiver<L2BlockInfo>,
 }
 
 #[async_trait]
 impl BlockBuildingClient for QueuedBlockBuildingClient {
+    async fn get_unsafe_head(&self) -> BlockEngineResult<L2BlockInfo> {
+        Ok(*self.unsafe_head_rx.borrow())
+    }
```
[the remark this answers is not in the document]
op-will: Not currently, no. If we ever do separate the actors, it could return an RPC error, so I left the Result<...> return type on the trait so we don't need to go back and add error handling to references.

crates/node/rpc/src/admin.rs
```diff
@@ -232,6 +232,10 @@ pub enum SequencerAdminAPIError {
     #[error("Error receiving response: {0}.")]
     ResponseError(String),
 
+    /// Error overriding leader.
+    #[error("Error stopping sequencer: {0}.")]
```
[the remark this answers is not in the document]
op-will: ffcd4fcbaf20ac5dc68c9c59f5ada94d74d3cd3a

PR #3096, opened 2025-11-26
chore(providers): recover provider

theochap:
## Description

Nit PR to recover from known slot interval if we cannot query the beacon node for the config/spec endpoint. This is the only place in the derivation process where we're using that endpoint (ideally I'd like to make this RPC call go away but I want to maintain forward compatibility).

The main use of this PR is to make `kona-node` fully compatible with anvil (running both as an EL and a CL client).

### DIFF
--- a/crates/providers/providers-alloy/src/blobs.rs
+++ b/crates/providers/providers-alloy/src/blobs.rs
@@ -12,6 +12,9 @@ use kona_derive::{BlobProvider, BlobProviderError};
 use kona_protocol::BlockInfo;
 use std::{boxed::Box, string::ToString, vec::Vec};
 
+/// The default slot interval used for the time to slot conversion.
+const DEFAULT_SLOT_INTERVAL: u64 = 12;
+
 /// A boxed blob with index.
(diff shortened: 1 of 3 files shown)


crates/providers/providers-alloy/src/blobs.rs
```diff
@@ -33,28 +33,37 @@ pub struct OnlineBlobProvider<B: BeaconClient> {
...
+
+        if let Ok(slot_interval) =
+            beacon_client.config_spec().await.map(|r| r.data.seconds_per_slot)
+        {
+            return Self { beacon_client, genesis_time, slot_interval };
+        }
+
+        const MAINNET_CHAIN_ID: u64 = 1;
+        const SEPOLIA_CHAIN_ID: u64 = 11155111;
+        const DEFAULT_SLOT_INTERVAL: u64 = 12;
```
op-will: [nit] Are these defined anywhere else that we can just reference here?

Not a big issue, just seems strange to have these very well-known values defined here.

PR #3097, opened 2025-11-26
chore(node/actor): Remove .expect(...) usage

### LINKED ISSUE #3090: chore(node/actor): Remove `.expect("...")` usage
[RollupNode](https://github.com/op-rs/kona/blob/0dd0d0125d3116a53884f8c6569531944692d656/crates/node/service/src/service/node.rs#L139) and [SequencerActorBuilder](https://github.com/op-rs/kona/blob/0dd0d0125d3116a53884f8c6569531944692d656/crates/node/service/src/actors/sequencer/builder.rs#L152) currently use `.expect("...")` for assertions that can cause a panic. This issue is to review the places in the node crate in which kona can panic and instead use the type system.

This issue comes out of this [comment](https://github.com/op-rs/kona/pull/3087#discussion_r2558062447)

varun-doshi:
Fixes #3090

### DIFF
--- a/crates/node/service/src/actors/sequencer/admin_api_impl_test.rs
+++ b/crates/node/service/src/actors/sequencer/admin_api_impl_test.rs
@@ -24,16 +24,17 @@ fn test_builder() -> SequencerActorBuilder<
     MockUnsafePayloadGossipClient,
 > {
     let (_admin_api_tx, admin_api_rx) = mpsc::channel(20);
-    SequencerActorBuilder::new()
-        .with_active_status(true)
-        .with_admin_api_receiver(admin_api_rx)
-        .with_attributes_builder(TestAttributesBuilder { attributes: vec![] })
-        .with_block_building_client(MockBlockBuildingClient::new())
-        .with_cancellation_token(CancellationToken::new())
-        .with_origin_selector(MockOriginSelector::new())
-        .with_recovery_mode_status(false)
-        .with_rollup_config(Arc::new(RollupConfig::default()))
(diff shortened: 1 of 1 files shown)


crates/node/service/src/service/node.rs
```diff
@@ -220,16 +220,10 @@ impl RollupNode {
...
+                    build_request_tx: build_request_tx.unwrap(),
                     reset_request_tx: reset_request_tx.clone(),
-                    seal_request_tx: seal_request_tx.expect(
-                        "seal_request_tx is None in sequencer mode. This should never happen.",
-                    ),
-                    unsafe_head_rx: unsafe_head_rx.expect(
-                        "unsafe_head_rx is None in sequencer mode. This should never happen.",
-                    ),
+                    seal_request_tx: seal_request_tx.unwrap(),
+                    unsafe_head_rx: unsafe_head_rx.unwrap(),
```
op-will: Here's where the validation / error handling should be

varun-doshi:
Changes made:
- `SequencerActorBuilder` does not hold any optional fields. Fields that do not have Default, must be passed in during the `new` call. 
- `build` cannot fail since all fields must have been set by then
- instead of creating a `sequencer_builder` in the first block and then creating the `sequencer_actor` in the second block, the code will now create the `sequencer_actor` at the end of the first block itself. 
- also seems to me that doing `build_request_tx.unwrap()` is safe since it will never be `None` if `mode.is_sequencer()`

varun-doshi:
checking CI failiures

Issue #3070, opened 2025-11-21
chore(node/actor) Add unit tests for SequencerActor functionality


op-will:
May depend on:
* #3068 
* #3068 

Now that SequencerActor dependencies are injected and easily mockable, this task is to add unit tests for SequencerActor functionality. Specific areas of focus are:
- [ ] Block building loop/timing
- [ ] Block building error handling (how all build/seal errors are handled)
- [ ] Block building race conditions
    * e.g. unsafe head changing between build and seal calls
    * other cases in which current state dictates that block building is not possible
- [ ] Admin API handling

It will likely be useful to create tooling to create mocks for dependencies. Some dependencies will ultimately be reused across actors, so this tooling could be useful in multiple places.

PR #3101, opened 2025-11-26
chore(node/actor): Adds tests for Sequencer Admin API

op-will:
Also:
- Adds admin RPC method to inspect recovery mode status
- Adds handling for new admin RPC method in SequencerActor
- Adds test utils via automock
- References existing test utils in crates/protocol/derive

Ref: #3070


einar-oplabs (approved):
Looks like a clean improvement.

Issue #3094, opened 2025-11-25
chore(test/conductor): reactivate conductor-tests in CI


theochap:
## Description

Our CI is currently skipping the conductor tests because the preset we're using doesn't contain any conductors. We should reenable those tests: https://github.com/op-rs/kona/blob/6992b5f26b1567eb37fa006d31d0c5a2c3714e40/tests/node/common/conductor_test.go#L25-L99.

PR #3116, opened 2025-11-30
chore(tests/conductor): add three new conductor test scenarios

### LINKED ISSUE #3095: chore(tests/conductor): increase coverage of kona-node in a conductor set
## Description

We should add more tests with conductor enabled, in particular:
- Sequencer failover - what happens when a sequencer fails/goes offline for a while? Normally our conductor should recover gracefully by forwarding block building to another sequencer
- Adding a new sequencer to the set - we should disable/stop a sequencer for a while and see if he catches up
- We should probably add more unit tests for initial conductor sync with the new testing pattern that @op-will introduced.

Himess:
Fixes #3095

### DIFF
--- a/tests/node/common/conductor_test.go
+++ b/tests/node/common/conductor_test.go
@@ -129,3 +129,222 @@ func testTransferLeadershipAndCheck(t devtest.T, oldLeader, targetLeader conduct
 		)
 	})
 }
+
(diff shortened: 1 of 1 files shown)


Himess:
See https://github.com/op-rs/kona/issues/3094#issuecomment-3615828716

PR #3118, opened 2025-12-01
feat(docker): Build cannon kona prestates for custom chains

Inphi:
This patch adds support for custom chain configurations when invoking the `just cannon` recipe to build cannon kona prestates.
This is useful for devnets and unannounced chains to also enjoy reproducible prestate builds.

## Testing

- [x] Build canonical prestates of kona-client releases and ensure the hashes are [standard prestates hashes](https://github.com/ethereum-optimism/superchain-registry/blob/17e5b8f2639324402f192d379cba0d09cddd2a6c/validation/standard/standard-prestates.toml).
- [x] Verify that custom prestates contain the custom chain configs


theochap (approved):
Looks good, thanks for taking care of it

Issue #3073, opened 2025-11-21
feat(node/actor): Make EngineActor generic over its dependencies and inject them


op-will:
# Description
#3045 changed the `NodeActor` trait and how actors are created within `RollupNode`. The purpose of that change was to allow dependencies to be injected into the various actors in order to loosen dependence on their implementation details, which allows the implementations to change without breaking interactions and facilitates unit testing.

`SequencerActor` was used as an example:
https://github.com/op-rs/kona/blob/15dac0759081de6683f900b4c20226eb8fa3b806/crates/node/service/src/actors/sequencer/actor.rs#L60-L66

This issue is to make the `EngineActor` actor generic over its dependencies by creating traits to wrap their functionality, configuring them within `RollupNode`, and injecting them into into `EngineActor`.

Note: the main goal is to be able to test all functionality in and downstream of the actor.

PR #3123, opened 2025-12-02
feat(node/EngineActor): Wraps `EngineClient` in a trait to allow mocking in task tests

op-will:
The EngineClient struct is a dependency of all `EngineActor` tasks and makes it extremely hard to unit test most parts of the `EngineActor` without running a dummy server (which would introduce many different ways for unit tests to fail that are not related to the code under test).

This creates an `EngineClient` trait, moving the `EngineClient` struct to `OpEngineClient`, and makes all tasks generic over this trait.

This is a larger change since generics are infectious, and all of the references up to and through `EngineActor` also needed to become generic.

Also creates mocks and examples of how to use this client in testing (see: `BuildTask` tests).

ref: #3073 since this is one step in making the EngineActor unit testable

### DIFF
--- a/crates/node/engine/src/task_queue/tasks/build/mod.rs
+++ b/crates/node/engine/src/task_queue/tasks/build/mod.rs
@@ -7,4 +7,4 @@ mod error;
 pub use error::{BuildTaskError, EngineBuildError};
 
 #[cfg(test)]
-mod tests;
+mod task_test;
--- a/crates/node/engine/src/task_queue/tasks/build/task_test.rs
+++ b/crates/node/engine/src/task_queue/tasks/build/task_test.rs
@@ -1,8 +1,8 @@
-//! Tests for BuildTask::start_build
+//! Tests for BuildTask::execute
 
 use crate::{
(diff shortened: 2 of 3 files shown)


crates/node/service/src/actors/engine/actor.rs
```diff
@@ -314,19 +331,24 @@ impl EngineActor {
...
+    L2Provider: Provider<Optimism>,
+    EngineClient_: EngineClient<L1Provider, L2Provider> + 'static,
+{
     /// Starts a task to handle engine queries.
     fn start_query_task(
         &self,
         mut inbound_query_channel: tokio::sync::mpsc::Receiver<EngineQueries>,
         mut rollup_boost_admin_query_rx: tokio::sync::mpsc::Receiver<RollupBoostAdminQuery>,
         mut rollup_boost_health_query_rx: tokio::sync::mpsc::Receiver<RollupBoostHealthQuery>,
+        rollup_boost: Arc<RollupBoostServer>,
```
theochap: Would it be possible to move the rollup-boost methods into the `L2Provider`? At the end of the day, our `L2Provider` will always be compatible with rollup-boost interface

PR #3128, opened 2025-12-03
fix(registry): avoid extra allocation in default_blob_schedule keys

Snezhkko:
Replaced name().to_string().to_lowercase() with name().to_lowercase() when building blob_schedule keys in 
L1Config. name() returns &'static str, so going straight to to_lowercase() avoids an intermediate allocation without changing behavior. Tests already use the same lowercase form for lookups.


Snezhkko:
> Please fix lint issues (run `just f`)

corrected

PR #3140, opened 2025-12-04
fix(engine): disallow invalid EngineSyncState construction

### LINKED ISSUE #3139: feat(node/engine): Disallow invalid `EngineSyncState` construction
Some Engine tasks have checks to make sure that the `EngineSyncState` they are working with is valid. Example: https://github.com/op-rs/kona/blob/d7d4b5da8022f4b7a060d4ef6723b5b236ff7f2e/crates/node/engine/src/task_queue/tasks/build/task.rs#L90-L99

This feels like a code smell because why would we allow invalid `EngineSyncState`s to get created only to cause problems downstream?

This task is to add error handling to the [apply_update](https://github.com/op-rs/kona/blob/d7d4b5da8022f4b7a060d4ef6723b5b236ff7f2e/crates/node/engine/src/state/core.rs#L82) function as well as any other instantiations of `EngineSyncState` to
1. Guarantee all cases of invalid `EngineSyncState` are caught 
2. Cause an error right when the problem occurs, instead of somewhere downstream
3. Create and perform validation logic in a single place rather than its various uses

Himess:
Move validation logic from downstream Engine tasks into the EngineSyncState::apply_update function to catch invalid states at construction time rather than downstream.

Changes:
- Add EngineSyncStateError enum for validation errors
- Update apply_update to return Result with validation
- Remove duplicate validation checks from SynchronizeTask and BuildTask
- Update error types to include InvalidSyncState variant
- Add unit tests for validation logic

This ensures:
- All invalid EngineSyncState cases are caught immediately
- Validation logic exists in a single place
- Errors occur at the source rather than propagating downstream

Closes #3139

### DIFF
--- a/crates/node/engine/src/test_utils/engine_state.rs
+++ b/crates/node/engine/src/test_utils/engine_state.rs
@@ -76,13 +76,16 @@ impl TestEngineStateBuilder {
         let mut state = EngineState::default();
 
         // Set unsafe head (required)
-        state.sync_state = state.sync_state.apply_update(EngineSyncStateUpdate {
-            unsafe_head: Some(self.unsafe_head),
-            cross_unsafe_head: Some(self.cross_unsafe_head.unwrap_or(self.unsafe_head)),
-            local_safe_head: Some(self.local_safe_head.unwrap_or(self.unsafe_head)),
-            safe_head: Some(self.safe_head.unwrap_or(self.unsafe_head)),
-            finalized_head: Some(self.finalized_head.unwrap_or(self.unsafe_head)),
-        });
+        state.sync_state = state
+            .sync_state
+            .apply_update(EngineSyncStateUpdate {
+                unsafe_head: Some(self.unsafe_head),
+                cross_unsafe_head: Some(self.cross_unsafe_head.unwrap_or(self.unsafe_head)),
+                local_safe_head: Some(self.local_safe_head.unwrap_or(self.unsafe_head)),
+                safe_head: Some(self.safe_head.unwrap_or(self.unsafe_head)),
+                finalized_head: Some(self.finalized_head.unwrap_or(self.unsafe_head)),
+            })
+            .expect("invalid sync state in test builder");
 
         state.el_sync_finished = self.el_sync_finished;
         state


crates/node/engine/src/state/core.rs
```diff
@@ -4,6 +4,29 @@ use crate::Metrics;
...
 use serde::{Deserialize, Serialize};
+use thiserror::Error;
+
+/// An error that occurs when constructing or updating an [`EngineSyncState`].
+///
+/// This error type represents invalid state configurations that violate the safety level
+/// invariants. The safety levels must maintain the ordering:
+/// `finalized <= safe <= local_safe <= cross_unsafe <= unsafe`
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Error)]
+pub enum EngineSyncStateError {
```
op-will:
```suggestion
pub enum InvalidEngineSyncStateError {
```

Himess:
I'll investigate the root cause and see where this inconsistent state originates from. Will update here once I find something.
@op-will

Himess:
@op-will Done .
That was hard lol :)
Could you look again ?

crates/node/engine/src/state/core.rs
```diff
@@ -184,55 +252,90 @@ mod test {
...
-                unsafe_head: Some(unsafe_head),
-                ..Default::default()
-            });
+            self.sync_state = self
+                .sync_state
+                .apply_update(EngineSyncStateUpdate {
+                    unsafe_head: Some(unsafe_head),
+                    ..Default::default()
+                })
+                .expect("valid sync state update");
```
op-will: I think we'll want all of these setters to return an error

crates/node/engine/src/task_queue/tasks/build/task.rs
```diff
@@ -88,25 +88,40 @@ impl<EngineClient_: EngineClient> BuildTask<EngineClient_> {
...
+        };
+
         let new_forkchoice = state
             .sync_state
             .apply_update(EngineSyncStateUpdate {
-                unsafe_head: Some(attributes_envelope.parent),
+                unsafe_head: Some(parent),
+                cross_unsafe_head: Some(parent),
+                local_safe_head: local_safe_update,
+                safe_head: safe_update,
```
op-will: This seems dangerous. Here we just want to update `unsafe_head`. If `cross_unsafe_head`, `local_safe_head`, or `safe_head` are head of that, we have a problem. 

We probably do want to err in that case. Setting them all to be equal would hide the error behind a plausibly, but not likely, accurate state.

PR #3144, opened 2025-12-04
nit(node/service): move all the l1 arguments into a single struct

theochap:
## Description

Really simple PR that moves all the l1 related config args into the same struct.

### DIFF
--- a/crates/node/service/src/service/node.rs
+++ b/crates/node/service/src/service/node.rs
@@ -22,22 +22,29 @@ use tokio_util::sync::CancellationToken;
 
 const DERIVATION_PROVIDER_CACHE_SIZE: usize = 1024;
 
+/// The configuration for the L1 chain.
+#[derive(Debug, Clone)]
+pub struct L1Config {
+    /// The L1 chain configuration.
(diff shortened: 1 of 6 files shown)


crates/node/service/src/service/node.rs
```diff
@@ -22,22 +22,29 @@ use tokio_util::sync::CancellationToken;
...
 
+/// The configuration for the L1 chain.
+#[derive(Debug, Clone)]
+pub struct L1Config {
+    /// The L1 chain configuration.
+    pub chain: Arc<L1ChainConfig>,
+    /// Whether to trust the L1 RPC.
+    pub trust_rpc: bool,
+    /// The L1 beacon client.
+    pub beacon: OnlineBeaconClient,
```
op-will: nit: Doesn't hurt to be slightly more specific about these.

```suggestion
    pub chain_config: Arc<L1ChainConfig>,
    /// Whether to trust the L1 RPC.
    pub trust_rpc: bool,
    /// The L1 beacon client.
    pub beacon_client: OnlineBeaconClient,
```

PR #3146, opened 2025-12-05
feat(node): add OS signal handling for graceful shutdown

### LINKED ISSUE #3091: chore(node): Handle termination gracefully
All of kona's actors receive a `CancellationToken` to be able to gracefully shutdown if signaled, but the kona process itself does not handle OS signals and initiate a graceful shutdown via the cancellation token. This issue is to add this OS signal handling to support gracefully shutting down kona actors.

Himess:
## Summary
- Added `shutdown_signal()` function to listen for SIGTERM and SIGINT
- Spawned a task in `RollupNode::start()` that triggers the existing `CancellationToken` when OS signals are received
- All actors already support graceful shutdown via CancellationToken, this just adds the missing OS signal handling

## Changes
- Modified `crates/node/service/src/service/node.rs`
- Cross-platform support: SIGTERM on Unix, Ctrl+C (SIGINT) on all platforms

Closes #3091

### DIFF
--- a/crates/node/service/src/actors/sequencer/actor.rs
+++ b/crates/node/service/src/actors/sequencer/actor.rs
@@ -33,7 +33,7 @@ use tokio_util::sync::{CancellationToken, WaitForCancellationFuture};
 
 /// The handle to a block that has been started but not sealed.
 #[derive(Debug)]
-struct UnsealedPayloadHandle {
+pub(super) struct UnsealedPayloadHandle {
     /// The [`PayloadId`] of the unsealed payload.
     pub payload_id: PayloadId,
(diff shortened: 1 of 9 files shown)


Himess:
> Looks good, instead of using a separate thread, let's move the handler to the place where we instantiate the actors (in the macro's select statement).
> 
> I would like to see some additional simplifications to the cancellation logic before we close the linked issue. In particular, we should handle actor cancellation at the rollup_node::start level

 Done! Moved the signal handling into the macro's select loop as you suggested. 
The shutdown_signal function is now in util.rs and gets called directly from spawn_and_wait instead of spawning a separate task.

PR #3148, opened 2025-12-05
chore(kurtosis): remove deprecated kurtosis deployments from kona.

theochap:
## Description

We dropped support for kurtosis a few months ago but never got around cleaning up the test folder. This PR does some basic clean-ups and updates the test docs


tests/supervisor/l1reorg/l1reorg/init_test.go
op-will: Does it make sense to have `l1reorg/l1reorg/` or can we pull the files up a dir?

PR #3152, opened 2025-12-05
refactor(node/service): remove SequencerActorBuilder

### LINKED ISSUE #3151: refactor(node/service): remove SequencerActorBuilder

einar-oplabs:
closes #3151


crates/node/service/src/actors/sequencer/actor.rs
```diff
@@ -28,7 +28,10 @@ use std::{
     sync::Arc,
     time::{Duration, Instant, SystemTime, UNIX_EPOCH},
 };
-use tokio::{select, sync::mpsc};
+use tokio::{
+    select,
+    sync::mpsc::{self, Receiver},
```
op-will: is this update necessary? 

I think for disambiguation we reference Senders/Receivers as `mpsc::Sender`/`mpsc::Receiver` everywhere, so we can revert this import and continue that pattern.

op-will (approved):
Looks great.

Only small nitpicks, so please address as you see fit and merge after you verify tests pass 👍

crates/node/service/src/actors/sequencer/actor.rs
```diff
@@ -136,6 +110,33 @@ where
...
+        admin_api_rx: mpsc::Receiver<SequencerAdminQuery>,
+        attributes_builder: AttributesBuilder_,
+        block_building_client: BlockBuildingClient_,
+        cancellation_token: CancellationToken,
+        conductor: Option<Conductor_>,
+        is_active: bool,
+        in_recovery_mode: bool,
+        origin_selector: OriginSelector_,
+        rollup_config: Arc<RollupConfig>,
+        unsafe_payload_gossip_client: UnsafePayloadGossipClient_,
```
theochap: Let's either move those args in a struct to pass as a constructor or set the fields of that struct as public to directly build it. Either way we should avoid to silence the clippy warnings.

If there are some closely related fields there we can also group them in a wrapper type

theochap (approved):
Love the negative diff, a few comments

PR #3175, opened 2025-12-09
feat(metrics): total sequenced transactions

einar-oplabs:


einar-oplabs:
There seem to be buggy behavior in the current version of the `inc!` macro. It has 3 arms, but it seems only arm 0 and 1 have been used in code.  There are no tests.  When trying to trigger the arm 2 (the one that allows for an increment parameter), it is actually matched by arm 1 instead, but then fails with `leftover tokens`.  If I swap arm 1 and 2 it succeeds. However, I don't if that changes the behavior of other macro calls, so we need to feel comfortable before merging this one.

op-will (approved):
Looks good!

PR #3180, opened 2025-12-10
chore(docker/ci): add nightly docker image builds

### LINKED ISSUE #3163: chore(docker/ci): add a CI job to cut nightly docker images of `kona`
## Description

We should add a CI job that cuts nightly docker images of `kona-node`/`kona-client`/`kona-host`.

ilikecodewithmyhand:
Closes #3163

Adds a scheduled workflow to build nightly docker images for `kona-node`, `kona-host`, and `kona-supervisor`.

Runs daily at 2 AM UTC, builds for both amd64 and arm64. Images are pushed to ghcr.io with `nightly` and `nightly-YYYY-MM-DD` tags.

Can also be triggered manually via workflow_dispatch.

### DIFF
--- a/.github/workflows/docker-nightly.yaml
+++ b/.github/workflows/docker-nightly.yaml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        target: [kona-node, kona-host, kona-supervisor]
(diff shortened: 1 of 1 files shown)


theochap:
Looks good, could you add support for kona-client as well?

Issue #2993, opened 2025-10-29
Incorrect use of `BaseFeeConfig::as_base_fee_params` after Canyon activation


kien-rise:
## Problem

When Canyon is activated, `BaseFeeConfig::as_canyon_base_fee_params` must be used instead of `BaseFeeConfig::as_base_fee_params`. Currently, the wrong one is being used: https://github.com/op-rs/kona/blob/kona-client%2Fv1.1.3/crates/proof/executor/src/util.rs#L76

For your reference, here is the code from OP, where `EIP1559DenominatorCanyon` is properly used: https://github.com/ethereum-optimism/optimism/blob/v1.14.3/op-node/rollup/attributes/engine_consolidate.go#L117-L119

## Suggested changes

```diff
  pub(crate) fn encode_holocene_eip_1559_params(
      config: &RollupConfig,
      attributes: &OpPayloadAttributes,
  ) -> ExecutorResult<Bytes> {
      Ok(encode_holocene_extra_data(
          attributes.eip_1559_params.ok_or(ExecutorError::MissingEIP1559Params)?,
-         config.chain_op_config.as_base_fee_params(),
+         config.chain_op_config.as_canyon_base_fee_params(),
      )?)
  }
```

```diff
  pub(crate) fn encode_jovian_eip_1559_params(
      config: &RollupConfig,
      attributes: &OpPayloadAttributes,
  ) -> ExecutorResult<Bytes> {
      Ok(encode_jovian_extra_data(
          attributes.eip_1559_params.ok_or(ExecutorError::MissingEIP1559Params)?,
-         config.chain_op_config.as_base_fee_params(),
+         config.chain_op_config.as_canyon_base_fee_params(),
          attributes
              .min_base_fee
              .ok_or(ExecutorError::InvalidExtraData(EIP1559ParamError::MinBaseFeeNotSet))?,
      )?)
  }
```

To prevent confusion or accidental misuse, I suggest renaming the functions as follows:

* Rename `as_base_fee_params` to `pre_canyon_base_fee_params` (or simply `pre_canyon`)
* Rename `as_canyon_base_fee_params` to `post_canyon_base_fee_params` (or simply `post_canyon`)

PR #3188, opened 2025-12-11
fix: use proper base fee params post-Canyon

kien-rise:
Fixes https://github.com/op-rs/kona/issues/2993


theochap:
@kien-rise would you mind fixing the lint jobs?

kien-rise:
@theochap I have fixed the CI error.

PR #3194, opened 2025-12-12
fix(ci): reduce ci usage by only running jobs on merge group

theochap:
## Description

CI usage has been high, this reduces the usage by running expensive jobs only in the mq


theochap:
Force merging that one so that the changes are applied

PR #3197, opened 2025-12-16
feat(p2p/dns): parse dns provided in CLI for p2p listeners

theochap:
## Description

Add support for input DNS(es) for p2p listener ips. Useful for docker context


op-will (approved):
👍

PR #3199, opened 2025-12-16
feat(node): Kona Light CL - First iteration

pcw109550:
# Light CL

Public Design doc at https://github.com/ethereum-optimism/design-docs/pull/359. The PR description may overlap, but more implementation oriented.

## Summary

This PR implements "follow mode" for the Kona rollup node, enabling it to sync safe/finalized L2 heads + CurrentL1 view from an external L2 Consensus Layer (CL) node instead of deriving them from L1 data. This provides a "light CL" mode that trusts an external source while maintaining local EL execution.

**CLI Flag**: `--l2.follow.source <URL>` or **Env**: `KONA_NODE_L2_FOLLOW_SOURCE=<URL>`

### Component Flow in Follow Mode

When follow mode is enabled (`--l2.follow.source` provided):

1. **Rollup Node Service** orchestrates and spawns **Follow Actor** (Derivation Actor not created)
2. **Engine Actor** completes initial EL sync
3. **Engine Actor** signals **Follow Actor** that initial EL sync is complete
4. **Follow Actor** starts polling **External L2 CL** for sync status via `optimism_syncStatus` RPC (every 2 seconds)
5. **Follow Actor** validates L1 block canonicality by checking against **L1 EL RPC**
6. **Follow Actor** sends validated `FollowStatus` to **Engine Actor**
7. **L2 Sequencer Gossip** continues to send unsafe blocks to **Engine Actor** (P2P remains active)

### Key Architectural Points

1. **New Follow Actor**: Added as a new actor in the Rollup Node Service
   - Waits for initial EL sync signal before starting polls (mirrors DerivationActor pattern)
   - Polls external L2 CL source at 2-second intervals after initial EL sync completes
   - Validates L1 blocks(L1 Origins) are canonical before triggering updates

2. **Bidirectional Communication with Engine Actor**:
   - **Engine → Follow**: Notifies when initial EL sync completes (via `el_sync_complete` signal)
   - **Follow → Engine**: Sends `FollowStatus` with external safe/finalized heads

3. **Compatible with Sequencer Mode**: A sequencer can use follow mode for safe/finalized heads while still producing unsafe blocks

#### 1. **FollowActor** (`crates/node/service/src/actors/follow/actor.rs`)
- Gates on initial EL sync signal before starting polls
- Polls external L2 CL via `optimism_syncStatus` RPC
- Validates L1 block canonicality (checks external L1 origins against local L1 chain)
- Sends validated `FollowStatus` to Engine Actor

#### 2. **FollowTask** (`crates/node/engine/src/task_queue/tasks/follow/task.rs`)
- New engine task type for processing external sync status
- Calls `SynchronizeTask` to apply safe/finalized head updates
- Priority: Higher than Consolidate/Finalize, lower than Insert (unsafe blocks). Consolidate/Finalize will never be scheduled when follow mode.

#### 3. **5-Case Follow Source Algorithm** (`crates/node/service/src/actors/engine/actor.rs:790-880`)
- **Case 1**: External safe ahead of local unsafe → full update (sets unsafe = external safe)
- **Case 2a**: Block not found locally → safe-only update (EL still syncing)
- **Case 2b**: Query error → skip update
- **Case 3**: Hashes match → consolidation (safe-only update)
- **Case 4**: Hash mismatch → reorg detected (full update)

#### 4. **Channel Wiring** (`crates/node/service/src/service/node.rs:154-174`)
- In follow mode: Creates dummy channels for derivation, routes `el_sync_complete_rx` to FollowActor
- In normal mode: Creates DerivationActor with standard channels

### Components in Follow Mode

| Component | Status | Behavior |
|-----------|--------|----------|
| DerivationActor | Not created | Disabled when follow enabled |
| FollowActor | Active | Polls external source after EL sync, validates, sends status |
| Engine Actor | Active | Processes FollowTask, updates state |
| L2Finalizer | Inactive | Queue stays empty (no derived attributes), harmless no-op |
| NetworkActor | Active | Receives unsafe blocks via P2P gossip |
| SequencerActor | Optional | Can produce unsafe blocks when in sequencer mode |

### L2Finalizer Behavior

The L2Finalizer component becomes inactive in follow mode but remains safely enabled:
- **Normal mode**: Finalizer tracks derived L2 blocks and finalizes them when L1 finalizes
- **Follow mode**: Finalizer receives L1 updates but has nothing to finalize (queue empty, `enqueue_for_finalization()` never called)
- **Result**: Component is inactive but harmless, finalization handled by FollowTask instead

### Testing

Commit https://github.com/op-rs/kona/pull/3199/commits/a50256c68a690449409667bb0665362a4facc1bc is baked to dev image `us-docker.pkg.dev/oplabs-tools-artifacts/dev-images/kona-node:a50256c-light-cl` and running as a 
- Replica with follow source enabled
- Sequencer as a leader with follow source enabled.

Note: rebased so relevant commit is not included at this PR's commit trail.

Acceptance Tests: 
- [x] `TestFollowL2_Safe_Finalized_CurrentL1`
- [x] `TestFollowL2_WithoutCLP2P`
- [ ] `TestFollowL2_ReorgRecovery`

For running the existing monorepo acceptance tests, build kona-node at place it at `/tmp/kona-node`. Then clone monorepo(Patch https://github.com/ethereum-optimism/optimism/tree/pcw109550/temp-kona-light-cl-acceptance-test-monkey-patch) and `cd op-acceptance-tests/tests/sync/follow_l2`. Run
```sh
DEVSTACK_L2CL_KIND=kona RUST_BINARY_PATH_KONA_NODE=/tmp/kona-node go test ./... -run ^[testname]$   -count=1 -v 
```

`TestFollowL2_ReorgRecovery` fails:

For op-node:
1. op-node Sequencer light CL triggers a reset because it detected L1 reorg while block building.
2. This is notified to the batcher, and batcher will re-batch the new L2 blocks and write to L1 which becomes canonical L2 safe blocks
3. op-node Verifier Light CL disabled will derive as usual, also triggering reset because it detects L1 reorg inside derivation pipeline.
4. op-node Sequencer Light CL enabled will consolidate because it already reset.
5. op-node Verifier Light CL enabled will reorg because the external source experienced L2 reorg.

For kona-node(which makes `TestFollowL2_ReorgRecovery` not pass)
1. kona-node sequencer Light CL does not reorg because its derivation pipeline is disabled, not emitting `ResetError::ReorgDetected`.
2. kona-node verifier Light CL disabled will derive as usual, also triggering reset because it detects L1 reorg inside derivation pipeline.
3. kona-node sequencer Light CL did not reorg, but keep building with invalid L2 blocks. 
4. Leading safe head to stall. 

We may explicitly detect L1 reorg at follow actor and trigger a explicit reset resulting in reorg. Need to find out why there is this difference at kona node, and implement a fix at follow up PR. 

All the happy case acceptance tests are passing.

## Known Limitations

**`current_l1` not mirrored**: External `current_l1` not propagated to local state
- **Blocker**: Pre-existing issue with `current_l1` tracking (documented in `current_l1.md`)
- **Plan**: Fix in separate PR, then integrate with follow mode

**L2 Reorg not working**: Light CL kona sequencer does not trigger L2 reorg when L1 reorg. Causing `TestFollowL2_ReorgRecovery` acceptance test to fail.
- **Blocker**: Find out the reason why op-node has L1 reorg detection at block building, but not at kona node. 
- **Plan**: Fix in separate PR.


bin/node/src/flags/engine/providers.rs
```diff
@@ -131,3 +131,12 @@ impl Default for L2ClientArgs {
...
 }
+
+/// L2 follow client arguments.
+#[derive(Clone, Debug, Default, clap::Args)]
+pub struct FollowClientArgs {
+    /// URL of the L2 follow source RPC API.
+    /// The source must be the L2 CL RPC.
+    #[arg(long, visible_alias = "l2.follow.source", env = "KONA_NODE_L2_FOLLOW_SOURCE")]
+    pub l2_follow_source: Option<Url>,
+}
```
op-will:
```suggestion
/// L2 derivation delegate connection arguments.
#[derive(Clone, Debug, Default, clap::Args)]
pub struct DerivationDelegateArgs {
    /// URL of the L2 derivation delegate RPC API.
    /// The url must be the L2 CL RPC.
    #[arg(long, visible_alias = "l2.follow.url", env = "KONA_NODE_L2_DERIVATION_DELEGATE_URL")]
    pub l2_derivation_delegate_url: Option<Url>,
}
```

bin/node/src/follow.rs
```diff
@@ -0,0 +1,180 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `number` - The L1 block number to fetch
+    ///
+    /// # Returns
+    ///
+    /// Returns the [`BlockInfo`] for the requested L1 block, or an error if
+    /// the block cannot be fetched or does not exist.
+    async fn l1_block_info_by_number(&self, number: u64) -> Result<BlockInfo, FollowClientError>;
```
op-will: Is there a reason to bundle this into `FollowClient`? 

It makes it seem like we're fetching the L1 block from the 3rd party that we trust for derivation, when we never should be. It would probably be clearer to make `FollowActor` generic over `Provider` and inject a `RootProvider` into it. Otherwise, it looks like we're asking the `follow_client` to validate data that it gave us, rather than obviously validating it using our own sources:

https://github.com/op-rs/kona/blob/c626ee7a4dc9607605b310ca7c03a40501f65c99/crates/node/service/src/actors/follow/actor.rs#L170-L185

PR #3209, opened 2025-12-18
fix: error when min_base_fee is missing during Jovian activation

Olexandr88:


theochap:
There is no requirements for the `RollupConfig` to specify a `min_base_fee` after `Jovian`'s activation. If the value is not set, a default (`0`) is used.

PR #3211, opened 2025-12-19
providers-alloy: `BeaconClient.filtered_beacon_blobs` fetches and validates only necessary blobs, and does not fallback to sidecars endpoint

### LINKED ISSUE #3136: perf(node): fetch only necessary blobs and add test coverage for duplicates
We've just encountered an incident on Unichain mainnet where the batcher managed to send identical blob data in two separate blob transactions, and those were included in the same block. This tripped up our op-node blob fetching code for the new `getBlobs` beacon endpoint that is in use since Fusaka activated. It got fixed with https://github.com/ethereum-optimism/optimism/pull/18492.

The same bug may be present in Kona's blob fetching code.

geoknee:
Closes #3136


crates/providers/providers-alloy/src/beacon_client.rs
```diff
@@ -88,6 +87,26 @@ pub trait BeaconClient {
...
+    let kzg_blob = Blob::new(blob.0);
+    let commitment =
+        kzg_settings.get().blob_to_kzg_commitment(&kzg_blob).map(|blob| blob.to_bytes()).unwrap();
+    kzg_to_versioned_hash(commitment.as_slice())
+}
+
+use thiserror::Error;
+
+#[derive(Error, Debug)]
+pub enum BeaconClientError {
```
geoknee: I am getting a linting error 
```
warning: enum `BeaconClientError` is reachable but cannot be named
   --> crates/providers/providers-alloy/src/beacon_client.rs:102:1
    |
102 | pub enum BeaconClientError {
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^ reachable at visibility `pub`, but can only be named at visibility `pub(crate)`
    |
    = note: requested on the command line with `-W unnameable-types`
   ```

File: crates/providers/providers-alloy/Cargo.toml [hunk 2]
```diff
@@ -58,3 +58,5 @@ metrics = [ "dep:metrics", "kona-derive/metrics" ]
 
 [dev-dependencies]
 tokio.workspace = true
+httpmock = "0.8.2"
```
theochap: When we're working with workspaces, it is usually better to pull dependency versions at the top-level `kona/Cargo.toml` (so that all crates use the same dependency version). Here, you can just use:
```suggestion
httpmock.workspace = true
```

crates/providers/providers-alloy/src/beacon_client.rs
```diff
@@ -88,6 +87,26 @@ pub trait BeaconClient {
     ) -> Result<Vec<BoxedBlobWithIndex>, Self::Error>;
 }
 
+/// blob_versioned_hash computes the versioned hash of a blob.
```
theochap: We can leverage the rust doc comments to point to the method name, so that when we change the name we need to change the associated doc.
```suggestion
/// [`blob_versioned_hash`] computes the versioned hash of a blob.
```

crates/providers/providers-alloy/src/beacon_client.rs
```diff
@@ -193,7 +204,7 @@ impl BeaconClient for OnlineBeaconClient {
             kona_macros::inc!(gauge, Metrics::BEACON_CLIENT_ERRORS, "method" => "spec");
         }
 
-        result
+        result.map_err(|err| BeaconClientError::Http(err))
```
theochap: nit, you could also use (for forward compatibility):
```suggestion
        Ok(result?)
```

crates/providers/providers-alloy/src/beacon_client.rs
```diff
@@ -209,14 +220,14 @@ impl BeaconClient for OnlineBeaconClient {
             kona_macros::inc!(gauge, Metrics::BEACON_CLIENT_ERRORS, "method" => "genesis");
         }
 
-        result
+        result.map_err(|err| BeaconClientError::Http(err))
```
theochap: Same comment as above

crates/providers/providers-alloy/src/beacon_client.rs
```diff
@@ -121,59 +140,51 @@ impl OnlineBeaconClient {
...
+            .query(&[("versioned_hashes", &params.join(",").as_str())])
+            .send()
+            .await?
+            .error_for_status()?;
+
+        let bundle = response.json::<GetBlobsResponse>().await?;
+
+        // Map blobs into versioned hashes for validation and
+        // matching against input:
+        let response_blob_hashes = bundle.data.iter().map(blob_versioned_hash).collect::<Vec<_>>();
```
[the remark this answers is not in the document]
theochap: Also, since it seems the method `blob_versioned_hash` is only called in this method, we could move the function definition inside `filtered_beacon_blobs` and either define a closure or keep the whole method:
(This is legal rust!)
```rust
async fn filtered_beacon_blobs(...) -> ... {
          /// blob_versioned_hash computes the versioned hash of a blob.
          fn blob_versioned_hash(blob: &FixedBytes<131072>) -> B256 {
              let kzg_settings = EnvKzgSettings::Default;
              let kzg_blob = Blob::new(blob.0);
              let commitment =
                  kzg_settings.get().blob_to_kzg_commitment(&kzg_blob).map(|blob| blob.to_bytes()).unwrap();
              kzg_to_versioned_hash(commitment.as_slice())
          }
         
         ....
}
```

Or (using a closure)
```rust
async fn filtered_beacon_blobs(...) -> ... {
          /// blob_versioned_hash computes the versioned hash of a blob.
          let blob_versioned_hash = |blob: &FixedBytes<131072>| -> B256 {
              let kzg_settings = EnvKzgSettings::Default;
              let kzg_blob = Blob::new(blob.0);
              let commitment =
                  kzg_settings.get().blob_to_kzg_commitment(&kzg_blob).map(|blob| blob.to_bytes()).unwrap();
              kzg_to_versioned_hash(commitment.as_slice())
          }
         
         ....
}

File: crates/providers/providers-alloy/Cargo.toml (hunk 2 above)
Comment (geoknee) [reply crates/providers/providers-alloy/Cargo.toml]: Did that here [0aa3dce](https://github.com/op-rs/kona/pull/3211/commits/0aa3dce00e19128b9509b92020a9aa6bf72973eb)

File: crates/providers/providers-alloy/src/beacon_client.rs [hunk 7]
```diff
@@ -88,6 +87,26 @@ pub trait BeaconClient {
     ) -> Result<Vec<BoxedBlobWithIndex>, Self::Error>;
 }
 
+/// blob_versioned_hash computes the versioned hash of a blob.
+fn blob_versioned_hash(blob: &FixedBytes<131072>) -> B256 {
+    let kzg_settings = EnvKzgSettings::Default;
+    let kzg_blob = Blob::new(blob.0);
+    let commitment =
+        kzg_settings.get().blob_to_kzg_commitment(&kzg_blob).map(|blob| blob.to_bytes()).unwrap();
```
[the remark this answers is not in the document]
geoknee: [65f35f4](https://github.com/op-rs/kona/pull/3211/commits/65f35f40e0f15a9fc8230e4acdf62913b5a8a715)

crates/providers/providers-alloy/src/beacon_client.rs
```diff
@@ -121,59 +147,52 @@ impl OnlineBeaconClient {
...
+        // whose hash matches the input:
+        blob_hashes
+            .iter()
+            .map(|blob_hash| -> Result<BoxedBlobWithIndex, BeaconClientError> {
+                let matching_data = returned_blobs_mapped_by_hash
+                    .get(&blob_hash.hash)
+                    .ok_or(BeaconClientError::BlobNotFound(blob_hash.hash.to_string()))?;
+                Ok(BoxedBlobWithIndex { blob: Box::new(**matching_data), index: blob_hash.index })
+            })
+            .collect::<Result<Vec<_>, BeaconClientError>>()
```
theochap: We probably talked about that already, but aren't we not supporting the deprecated endpoint anymore? It's fine if we don't, just making sure we're doing the change consciously

Cargo.toml
```diff
@@ -253,6 +253,7 @@ arbtest = "0.3.2"
 proptest = "1.9.0"
 criterion = "0.5.1"
 mockall = "0.14.0"
+httpmock = "0.8.2"
```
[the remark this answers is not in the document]
theochap: Would you mind opening an issue for that?

crates/providers/providers-alloy/src/beacon_client.rs
```diff
@@ -229,3 +248,97 @@ impl BeaconClient for OnlineBeaconClient {
...
+            .filtered_beacon_blobs(
+                slot,
+                &[
+                    IndexedBlobHash { index: 0, hash: blob_hash_of_interest },
+                    IndexedBlobHash { index: 2, hash: blob_hash_of_interest },
+                ],
+            ) // ask for blobs 0 and 2, which happen to have identical data and hashes
+            .await
+            .expect_err("Expected error when mocking an incorrect response from the beacon server");
+        incorrect_blobs_mock.assert();
```
theochap: Can we put that snippet in a separate test?

theochap (approved):
Looks great!

crates/providers/providers-alloy/src/beacon_client.rs
geoknee: Yes this PR is removing the fallback to the deprecated sidecars endpoint.

crates/providers/providers-alloy/src/beacon_client.rs
```diff
@@ -121,59 +147,52 @@ impl OnlineBeaconClient {
...
+        let bundle = response.json::<GetBlobsResponse>().await?;
+
+        let returned_blobs_mapped_by_hash = bundle
+            .data
+            .iter()
+            .map(|data| -> Result<_, BeaconClientError> {
+                let recomputed_hash = blob_versioned_hash(data)?;
+                Ok((recomputed_hash, data))
+            })
+            .collect::<Result<HashMap<_, _>, BeaconClientError>>()?;
```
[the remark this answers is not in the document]
geoknee: It is not an error to return multiple blobs with the same versioned hash, I am assuming here the versioned hash is 1-1 with the blob data itself (we don't need to store duplicates in this hashmap). The code below will just cause the blob data to be looked up multiple times in the case of duplicates.

Cargo.toml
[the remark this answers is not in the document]
geoknee: https://github.com/op-rs/kona/issues/3243

crates/providers/providers-alloy/src/beacon_client.rs
geoknee: I refactored it here [9645443](https://github.com/op-rs/kona/pull/3211/commits/964544359c927d71aefca4f672f883af986066be)

Issue #3219, opened 2025-12-19
Consolidate `DerivationState` into `DerivationActor`


op-will:
Note: Prerequisites for this task may be other sub-tasks of #3072.

Issue #3216, opened 2025-12-19
Remove `EngineContext`


op-will:
All dependencies passed to `start(...)` can, instead be declared on the `EngineActor` struct and injected into it via its constructor.

Himess:
Can i take it ?

op-will:
> Can i take it ?

There is a PR incoming for this ([op-will/consolidateDerivationActorInbound](https://github.com/op-rs/kona/tree/op-will/consolidateDerivationActorInbound) which builds off of [op-will/consolidateEngineActorInboundChannels](https://github.com/op-rs/kona/tree/op-will/consolidateEngineActorInboundChannels)). Do you mind taking another issue?

Issue #3215, opened 2025-12-19
Consolidate `EngineActorState` into `EngineActor`


op-will:
This entails:
1. Making EngineActor generic over `EngineClient`
2. Consolidating `Engine` into `EngineActor`

For 2, There is no reason why `EngineActor` can't/shouldn't have `Engine` logic. The alternative is to formalize that dependency via a trait and make `EngineActor` generic over it, but it is accomplishing EngineActor logic.

Issue #3213, opened 2025-12-19
Inject `inbound_request_rx` into EngineActor


op-will:
Note: this assumes #3206 gets merged.

Rather than having the EngineActor create its own channels, the `Receiver` can be injected.

PR #3229, opened 2026-01-05
cleanup(node/actor): Consolidate DerivationActor inbound channels

### LINKED ISSUE #3223: cleanup(node/actor): Consolidate DerivationActor inbound channels into single request channel
See parts 3 & 4 of [this design doc](https://github.com/ethereum-optimism/design-docs/blob/9bdd48aba0cd567386ef4ce03693a6b67bac17c3/protocol/kona-node-actor-simplification.md#3-and-4-actor-dependencies-are-unclear) for more info.

`DerivationActor` can and should expose a single inbound request channel and a clear enum of requests it accepts (see [this PR](https://github.com/op-rs/kona/pull/3206) for an example).

Actors that send `DerivationActor` information, directly or indirectly, should be injected with a `DerivationActorClient` trait, over which they are made to be generic.

For instance `EngineActor` sends `DerivationActor` updates over `derivation_signal_rx`, `el_sync_complete_rx`, and `engine_l2_safe_head`. Instead, 
1. These can each be a request type that `DerivationActor` accepts
2. `EngineActor` can directly call derivation actor via an `EngineDerivationClient` trait
3. All uses of `watch::Receiver.changed()` and `watch::Receiver.has_changed()` can become clearer and race-free

For 3, we use the `changed` state of `engine_l2_safe_head` for external storage of state that can and should live in `DerivationActor`. The goal appears to be to prevent more derivation loops when the L1 block has changed but the engine hasn't acknowledged the last safe head update. If we store the required state, we can make this check more explicit and prevent race conditions (i.e. the call to `engine_l2_safe_head.borrow_and_update()` that _might_ be marking an unseen safe head as seen if it has changed since last read).

We can avoid these race conditions entirely and prevent missing any updates by making each case of an update a request on the `DerivationActor`'s inbound queue. This may require us to ignore stale requests (i.e. a stale safe head update immediately after a reset), but that should be easy if we're storing the pending safe head and checking the acknowledged one against it.

op-will:
See: #3223 

This also reworks aspects of `EngineActor`. See referenced issues.

closes #3223, #3213, #3215, #3216, #3219, and #3072

### DIFF
--- a/crates/node/service/src/actors/derivation/actor.rs
+++ b/crates/node/service/src/actors/derivation/actor.rs
@@ -199,6 +199,56 @@ where
         }
     }
 
+    async fn handle_derivation_actor_request(
+        &mut self,
+        request_type: DerivationActorRequest,
+    ) -> Result<(), DerivationError> {
+        match request_type {
+            DerivationActorRequest::ProcessEngineSignalRequest(signal) => {
+                self.signal(*signal).await;
+                self.waiting_for_signal = false;
+            }
+            DerivationActorRequest::ProcessFinalizedL1Block(finalized_l1_block) => {
+                // Attempt to finalize the block. If successful, notify engine.
+                if let Some(l2_block_number) = self.finalizer.try_finalize_next(*finalized_l1_block)
+                {
+                    self.engine_client
+                        .send_finalized_l2_block(l2_block_number)
+                        .await
+                        .map_err(|e| DerivationError::Sender(Box::new(e)))?;
+                }
+            }
+            DerivationActorRequest::ProcessL1HeadUpdateRequest(l1_head) => {
+                info!(target: "derivation", l1_head = ?*l1_head, "Processing l1 head update");
+
+                // If derivation isn't idle and the message hasn't observed a safe head update
+                // already, check if the safe head has changed before continuing.
+                // This is to prevent attempts to progress the pipeline while it is
+                // in the middle of processing a channel.
+                if !self.derivation_idle && self.awaiting_engine_l2_safe_head_update {
+                    info!(target: "derivation", "Safe head hasn't changed, skipping derivation.");
+                } else {
+                    self.attempt_derivation().await?;
+                }
+            }
+            DerivationActorRequest::ProcessEngineSafeHeadUpdateRequest(safe_head) => {
+                info!(target: "derivation", safe_head = ?*safe_head, "Received safe head from engine.");
+                self.engine_l2_safe_head = *safe_head;
+                self.awaiting_engine_l2_safe_head_update = false;
+
+                self.attempt_derivation().await?;
+            }
+            DerivationActorRequest::ProcessEngineSyncCompletionRequest => {
+                info!(target: "derivation", "Engine finished syncing, starting derivation.");
+                self.has_engine_sync_completed = true;
+
+                self.attempt_derivation().await?;
+            }
+        }
+
+        Ok(())
+    }
+
     /// Attempts to process the next payload attributes.
     ///
     /// There are a few constraints around stepping on the derivation pipeline.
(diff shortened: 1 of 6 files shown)


crates/node/service/src/actors/derivation/actor.rs
```diff
@@ -468,54 +285,30 @@ where
...
-                }
-                msg = self.l1_head_updates.changed() => {
-                    if let Err(err) = msg {
-                        error!(
-                            target: "derivation",
-                            ?err,
-                            "L1 head update stream closed without cancellation. Exiting derivation task."
-                        );
-                        return Ok(());
+                    match request_type {
```
einar-oplabs: We could leave a comment somewhere about re-adding bias towards `DerivationActorRequest::ProcessEngineSignalRequest` when we can prioritize request variants.

op-will: Good call! This is something I wanted to visit during review: are we sure we actually need the bias? 

I know that there's a chance of processing an L1 head update followed by a reset or something, but that is a possibility via a race even if we re-add the bias, and I'm not sure being unbiased actually causes any problems.

If we think it does here and/or in other actors, we can create a pattern to read from the inbound channel, add to a priority queue, and process that queue in a separate task. I think we'll want to avoid that complexity unless we're sure it meaningfully improves things, though.

op-will: Noting that all acceptance tests pass, so I'm not sure the bias is needed.

einar-oplabs: > are we sure we actually need the bias?

No.  I suspect it was added to prioritize cancellation events.

I support leaving it unbiased until we can observe undesired behavior.  If we decide to add the priority queue, we will before forced to reconsider their mutual priorities anyway.

I think you can resolve this.

crates/node/service/src/actors/engine/engine_request_processor.rs
```diff
@@ -0,0 +1,320 @@
...
+                        }
+                    }
+                    EngineTaskErrorSeverity::Temporary => {
+                        trace!(target: "engine", ?err, "Temporary error draining engine tasks");
+                    }
+                }
+            }
+        }
+
+        self.send_derivation_actor_safe_head_if_updated().await?;
```
theochap: Maybe here, the `if_updated` part should be done explicitly in this method instead of delegating it to the interface. It is a matter of personal preference but I don't like to call methods that are doing a no-op, this is a source of error in my opinion

crates/node/service/src/actors/engine/engine_request_processor.rs
```diff
@@ -0,0 +1,320 @@
...
+    }
+
+    /// Attempts to send the [`crate::DerivationActor`] the safe head if updated.
+    async fn send_derivation_actor_safe_head_if_updated(&mut self) -> Result<(), EngineError> {
+        let engine_safe_head = self.engine.state().sync_state.safe_head();
+        if engine_safe_head == self.last_safe_head_sent {
+            info!(target: "engine", safe_head = ?engine_safe_head, "Safe head unchanged");
+            // This was already sent, so do not send it.
+            return Ok(());
+        }
```
theochap: Let's inline this check

crates/node/service/src/actors/engine/engine_request_processor.rs
```diff
@@ -0,0 +1,320 @@
...
+                        // Send the result.
+                        let response_payload = reset_res
+                            .as_ref()
+                            .map(|_| ())
+                            .map_err(|e| EngineClientError::ResetForkchoiceError(e.to_string()));
+                        if reset_request.result_tx.send(response_payload).await.is_err() {
+                            warn!(target: "engine", "Sending reset response failed");
+                        }
+
+                        reset_res?;
```
theochap: We should propagate the error here only if failing to reset is a critical error. Maybe that is always the case for now, but probably safe to inspect the error, log it and then propagate it to the parent task. We should also ensure that reset errors are always worth killing the node for

crates/node/service/src/actors/engine/rpc_request_processor.rs
```diff
@@ -0,0 +1,106 @@
...
+
+                            },
+                            EngineRpcRequest::RollupBoostHealthRequest(health_query) => {
+                                trace!(target: "engine", ?health_query, "Received rollup boost health query.");
+
+                                let health = self.rollup_boost_server.get_health();
+                                health_query.sender.send(health.into()).unwrap();
+                            },
+
+                        }
```
theochap: Maybe worth moving those methods to a separate handler? The code is pretty nested there (4 or 5 levels of nestedness)

crates/node/service/src/service/node.rs
```diff
@@ -122,6 +111,93 @@ impl RollupNode {
...
+            engine_queue_length_rx,
+        );
+
+        Ok(EngineActor::new(
+            cancellation_token,
+            engine_request_rx,
+            engine_processor,
+            engine_rpc_processor,
+        ))
+    }
```
theochap: The logic to build the engine here is a bit complicated and I am slightly worried that we're pushing so much explicit behavior at the rollup-node-service level. It may be fine to keep it that way for this PR, but we should think of a way to fit all the actors under similar patterns so that we can make the `RollupNode` more generic/easier to modify. This doesn't need to be done here and it will probably require changing the other actors but I'd like us to keep this goal in mind for future changes

crates/node/service/src/actors/engine/engine_request_processor.rs
op-will: This logic is called in multiple places so the check is DRY'd by pulling it into this helper.

crates/node/service/src/actors/engine/engine_request_processor.rs
op-will: https://github.com/op-rs/kona/pull/3229/files#r2683956074

crates/node/service/src/service/node.rs
op-will: Hopefully we can remove `StartData` from `RpcActor`, which should allow us to remove `StartData` from `NodeActor`, and then this function can just take a Vec of `NodeActor`. 

That will push this configuration up into the NodeCommand, but
1. To me, it makes more sense for it to be there since the intention is for this to be a lib
2. We can still have all of the construction helpers in this lib so they can be used by our CLI as well as anyone else using the lib (if they choose)

crates/node/service/src/actors/engine/engine_request_processor.rs
op-will: > maybe it only makes sense to err if that send fails?

Did this here: https://github.com/op-rs/kona/pull/3229/commits/0769d7ca22c93d729e7770627e16816b3aa93517

crates/node/service/src/actors/engine/engine_request_processor.rs
```diff
@@ -0,0 +1,320 @@
...
+
+        Ok(())
+    }
+
+    /// Checks if the EL has finished syncing, notifying the derivation actor if it has.
+    async fn check_el_sync(&mut self) -> Result<(), EngineError> {
+        if self.engine.state().el_sync_finished {
+            if self.el_sync_complete {
+                return Ok(());
+            }
```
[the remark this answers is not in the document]
op-will: https://github.com/op-rs/kona/commit/0769d7ca22c93d729e7770627e16816b3aa93517

PR #3240, opened 2026-01-09
feat(cli): support dynamic port assignment for metrics server

teddyknox:
#### Problem

When `KONA_METRICS_PORT=0` was set, the OS would assign an available port, but there was no way to discover what port was actually assigned. The log message would show `:0` instead of the actual port, making it impossible to connect to the metrics endpoint.

#### Solution

Replaced the `metrics-exporter-prometheus` built-in HTTP listener with a manually managed server. The implementation:

1. Binds a `TcpListener` directly (supporting port 0)
2. Retrieves the actual bound address via `local_addr()`
3. Uses `PrometheusBuilder::build_recorder()` without the HTTP listener
4. Serves metrics manually via `hyper`
5. Returns the actual `SocketAddr` to callers

#### Design Rationale

The `metrics-exporter-prometheus` crate's `with_http_listener().install()` API doesn't expose the actual bound address—it only accepts a `SocketAddr` and returns `Result<(), BuildError>`. There's no way to query what port the OS assigned.

We considered two approaches:
- **Pre-bind and hope**: Bind a listener to get a port, drop it, pass that port to the builder. This has a race condition where another process could grab the port.
- **Manual serving**: Build the recorder without HTTP, bind our own listener, serve manually. No race condition.

We chose manual serving, which is the same pattern used by [reth's metrics infrastructure](https://github.com/paradigmxyz/reth/blob/main/crates/node/metrics/src/server.rs). This approach:
- Eliminates the race condition entirely (same listener is used throughout)
- Gives full control over the HTTP server
- Allows returning the actual bound address

#### Breaking Changes

- `init_prometheus_server()` is now `async` and takes `SocketAddr` instead of `(IpAddr, u16)`
- `MetricsArgs::init_metrics()` is now `async`
- Metrics initialization must now happen inside a tokio runtime

Call sites in `kona-node` and `kona-supervisor` have been updated to initialize metrics inside their async blocks.

#### Usage

```rust
// Get the actual address (useful when port is 0)
if let Some(addr) = args.init_metrics_with_addr().await? {
    println!("Metrics available at http://{}", addr);
}

// Or just initialize without needing the address
args.init_metrics().await?;
```


bin/node/src/cli.rs
```diff
@@ -61,21 +61,19 @@ impl Cli {
...
-        match self.subcommand {
-            Commands::Node(ref node) => node.init_cli_metrics(&self.global.metrics)?,
-            _ => {
-                tracing::debug!(target: "cli", "No CLI metrics initialized for subcommand: {:?}", self.subcommand)
-            }
-        }
-
         // Run the subcommand.
+        // Async commands (node, net) init metrics inside the tokio runtime.
+        // Sync utility commands (registry, bootstore, info) don't need metrics.
```
op-will: Does it harm anything to have metrics for those? It's nice to future-proof this.

op-will:
Closes #2987

PR #3241, opened 2026-01-11
Chore(protocol): drop unused test case name field

0xcharry:
Remove the unused name field and #[allow(dead_code)] from FrameValidityTestCase in protocol/src/channel.rs tests.


einar-oplabs:
@0xcharry I believe that you already opened one PR https://github.com/op-rs/kona/pull/3232 with these changes.  Please follow the feedback there.

